PR #152, opened 2020-01-23
Clean up modelconverter

Tombana:
Small cleanup of modelconverter.

- Prints note for how to use the MLIR converter.
- Adds `compat.v1` converter to be able to use attributes (needed later for int8)
- Makes it easier to add different conversion methods.


arashb:
@lgeiger do you still think we need this PR?

Tombana:
I think this can be closed and removed.

PR #175, opened 2020-02-10
Create Python wrapper for MLIR converter

lgeiger:
For now this removes the `tf_lce_translate` binary to keep it simple, we probably want to add it back once  we setup the proper MLIR testing infrastructure.


lgeiger:
Alright CI is now finally happy with it and this is ready for review. For now this only supports TensorFlow 2.1, but I'll add handling for broader version support in future PRs. Since this only requires Python code changes.

.github/workflows/unittests.yml
```diff
@@ -126,8 +126,14 @@ jobs:
         run: |
           curl -L https://github.com/bazelbuild/bazelisk/releases/download/v1.3.0/bazelisk-linux-amd64 > bazelisk
           chmod +x bazelisk
+      - name: Force Python 3 for everything we do
+        run: |
+          ln -sf /usr/bin/python3.6 /usr/bin/python
+          ln -sf /usr/bin/python3.6 /usr/bin/python3
```
Tombana: Hopefully we can soon figure out how to do this properly. Maybe simply uninstalling python2 at the start?

larq_compute_engine/mlir/python/converter.py
```diff
@@ -0,0 +1,79 @@
...
+    get_grappler_config,
+    get_tensor_name,
+    run_graph_optimizations,
+)
+from tensorflow.python.eager import def_function
+from tensorflow.python.framework import convert_to_constants, dtypes
+from tensorflow.python.keras.saving import saving_utils
+
+
+def convert_keras_model(model):
```
Tombana: Should we add a comment to mention that this is derived from the `from_keras_model` function in [this file](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/python/lite.py) ? Then we know that updates to that function can be upstreamed into our function.

larq_compute_engine/mlir/python/converter.py
```diff
@@ -0,0 +1,79 @@
...
+            shape[0] = 1
+            tensor.set_shape(shape)
+
+    return convert_graphdef_to_tflite_flatbuffer(
+        graph_def.SerializeToString(),
+        [get_tensor_name(tensor) for tensor in input_tensors],
+        [DataType.Name(tensor.dtype.as_datatype_enum) for tensor in input_tensors],
+        [tensor.shape.as_list() for tensor in input_tensors],
+        [get_tensor_name(tensor) for tensor in output_tensors],
+    )
```
Tombana: :rocket:  This call with pybind is *sooo* much cleaner than TFs separate `toco_from_protos` binary!

.github/workflows/unittests.yml
lgeiger: Unfortunately this is also how [Tensorflow Addons](https://github.com/tensorflow/addons/blob/master/tools/ci_build/builds/release_linux.sh#L45) solves their Python version problem. I agree there must be a better way to do this, but it would involve messing with `tf_configure` or trying to find the correct environment variable to set.
I think I'd be in favour of just running the tests on macos outside the container with a custom the github provided Python version which should make CI faster as well.

lgeiger: Just for reference, I think [this is the code](https://github.com/tensorflow/tensorflow/blob/4f2afc07b26748f80a0de768fac81c7816410b44/configure.py#L203-L269) that determines the Python version used by TensorFlow (and the TensorFlow provided pybind extension).

larq_compute_engine/mlir/python/converter.py
lgeiger: I agree. Are you fine with doing this in a future PR? Then I'd like to also add support for different TensorFlow versions and add the relevant tests to CI as well.

larq_compute_engine/mlir/python/converter.py
lgeiger: I agree, pybind11 is quite nice to use.

Also checkout [this RFC](https://github.com/tensorflow/community/blob/master/rfcs/20190208-pybind11.md) for more info about pybind11 inside TensorFlow.

.github/workflows/unittests.yml
lgeiger: Or maybe we could look into switching to the default `pybind_extension` https://github.com/pybind/pybind11_bazel

Tombana: I think we also tried that for some other bazel command: switching from the TF version to the normal bazel version, but then somehow things got messed up.

The code you linked that determines the Python version suggests setting `PYTHON_LIB_PATH` and `PYTHON_BIN_PATH`, but if this is how TF Addons solves it, then I'm 100% in favor of leaving it like this.

PR #177, opened 2020-02-12
Add fused activation support for valid padding

Tombana:
This adds the `"activation"` attribute which can be set to `"RELU"`.

### DIFF
--- a/larq_compute_engine/core/bgemm_kernels_ruy.h
+++ b/larq_compute_engine/core/bgemm_kernels_ruy.h
@@ -98,6 +98,8 @@ struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, DstScalar,
         if (spec.fused_add) {
           dst_val += spec.fused_add[i];
         }
+        dst_val = std::min<DstScalar>(dst_val, spec.clamp_max);
+        dst_val = std::max<DstScalar>(dst_val, spec.clamp_min);
         *ElementPtr(dst, i, j) = dst_val;
       }
     }
--- a/larq_compute_engine/tflite/tests/bconv2d_test.cc
+++ b/larq_compute_engine/tflite/tests/bconv2d_test.cc
@@ -42,6 +42,15 @@ TfLiteRegistration* Register_CONVOLUTION_GENERIC_OPT();
 }  // namespace builtin
 }  // namespace ops
 
+std::string getActivationString(const enum ActivationFunctionType activation) {
+  if (activation == ActivationFunctionType_RELU) {
+    return "RELU";
+  } else if (activation == ActivationFunctionType_NONE) {
+    return "NONE";
+  }
+  return "UNKOWN";
+}
+
 namespace {
 
 class BaseConvolutionOpModel : public SingleOpModel {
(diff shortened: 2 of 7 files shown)


larq_compute_engine/core/bgemm_kernels_ruy.h
```diff
@@ -98,6 +98,8 @@ struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, DstScalar,
         if (spec.fused_add) {
           dst_val += spec.fused_add[i];
         }
+        dst_val = std::min<DstScalar>(dst_val, spec.clamp_max);
+        dst_val = std::max<DstScalar>(dst_val, spec.clamp_min);
```
lgeiger: This would now calculate `Conv -> Mul -> Add -> Relu`, but for us to be useful in practice we'd want to do `Conv -> Relu -> Mul -> Add` or am I missing something?

Do you think it would make sense to split `fused_add` and `fused_multiply` into the backtransformation to {-1, +1} space and the floating fused multiplier + bias?

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -317,10 +330,15 @@ TEST_P(BConv2DOpTest, SimpleTest) {
...
   const int dilation_height_factor = param.dilation_height_factor;
   const int dilation_width_factor = param.dilation_width_factor;
-  // TOOD: currently we are ignoring padding
   const Padding padding = param.padding;
+  const ActivationFunctionType activation = param.activation;
   const int num_threads = param.num_threads;
 
+  // Fused ReLu is not yet supported for zero-padding
+  if (padding == Padding_SAME && activation == ActivationFunctionType_RELU) {
+    return;
```
arashb: this should return an error message.

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -189,12 +200,13 @@ struct TestParam {
...
-                     const TensorData& filter, const TensorData& output,
-                     int stride_width = 1, int stride_height = 1,
-                     enum Padding padding = Padding_VALID,
-                     int dilation_width_factor = 1,
-                     int dilation_height_factor = 1, int num_threads = -1) {
+  BaseBConv2DOpModel(
+      register_function registration, const TensorData& input,
+      const TensorData& filter, const TensorData& output, int stride_width = 1,
+      int stride_height = 1, enum Padding padding = Padding_VALID,
+      enum ActivationFunctionType activation = ActivationFunctionType_NONE,
```
arashb: why do you need to change the entire function list fo arguments. you can just add it new argument at the end of the list.

larq_compute_engine/core/bgemm_kernels_ruy.h
Tombana: So i'd be in favor of having *two* multiply and add inputs. Maybe something like `{pre,post}_activation_{multiply,add}`.
The reason is that between Conv and Relu, we need a multiply and add *anyway* for the backtransformation. So we might as well support custom multiply and add values since the number of computational ops stays the same.
The multiply and add *after* the Relu are really extra, and the TF Lite conv also does not fuse them. Now that we have the MLIR converter, that'll be easy to add. I'll try to make a PR soon.

larq_compute_engine/tflite/tests/bconv2d_test.cc
Tombana: I agree, but then the unit tests would fail. I wanted to only skip this particular test combination but I did not know another way.

larq_compute_engine/tflite/tests/bconv2d_test.cc
Tombana: I wanted to keep it synchronized with the order of the original conv, and therefore i added it after the padding argument. I can move it to the end if you want.

larq_compute_engine/tflite/tests/bconv2d_test.cc
arashb: in that case please create an issue that we don't forget bc with this test case I think we will see no failing tests for unsupported padding.

Tombana: I added this issue with a link to that line in the file: https://github.com/larq/compute-engine/issues/182

arashb:
close it for now.

PR #178, opened 2020-02-12
Cleanup configure script

Tombana:
This removes the gpu question from the configure script.


configure.sh
```diff
@@ -59,53 +47,25 @@ while [[ "$PIP_MANYLINUX2010" == "" ]]; do
...
-        echo 'Installing tensorflow 1.14......\n'
-        pip install tensorflow==1.14
-    else
-        echo 'Installing tensorflow 2......\n'
-        pip install tensorflow
-    fi
-  fi
-
+# Check if it's installed
+if [[ $(pip show tensorflow) == *tensorflow* ]] || [[ $(pip show tf-nightly) == *tf-nightly* ]] || [[ $(pip show tensorflow-cpu) == *tensorflow-cpu* ]]; then
```
arashb: this should not build against all tf-nightly.

Tombana: Note that this tf-nightly part was already in the original configure script. By default it will install normal tensorflow, but if tf-nightly was *already* installed by the user, then it will assume that the user wants to use that instead.

Do you still think we should remove it? I'm neutral on this.

arashb: the build might fail if you build against unstable TF so I am in favour of removing it.

Tombana:
> I am currently messing with the configure script as well. Can we wait with merging until #175 and #181 are merged?

I think you already applied the same changes in #181 (removing the GPU questions etc) so then this PR can probably be closed.

PR #181, opened 2020-02-13
Build wheels for MLIR converter on CI

lgeiger:
This PR adds automatic release builds for wheels including the MLIR converter and adds support for TensorFlow 1.14 eager, 1.15 eager, 2.0 and 2.1.


.github/tools/release_linux.sh
```diff
@@ -0,0 +1,39 @@
...
+apt-get -y -qq install python${PYTHON_VERSION}-dev --no-install-recommends
+ln -sf /usr/bin/python${PYTHON_VERSION} /usr/bin/python
+ln -sf /usr/bin/python${PYTHON_VERSION} /usr/bin/python3
+
+ls /usr/include/x86_64-linux-gnu/
+ls /dt7/usr/include/x86_64-linux-gnu/
+
+# This really shouldn't be required to get 3.7 builds working
+ln -sf /usr/include/x86_64-linux-gnu/python3.7m /dt7/usr/include/x86_64-linux-gnu/python3.7m || true
+ln -sf /usr/include/x86_64-linux-gnu/python3.7m /dt8/usr/include/x86_64-linux-gnu/python3.7m || true
```
lgeiger: Not sure what's going on here, but it looks like those files are missing from the custom op container.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/Dockerfile.custom_op_ubuntu_16#L77-L83

In any case this makes the Python 3.7 wheels build.

.github/tools/release_linux.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+set -e -x
+
+# Remove the now private ppa. This can be removed after the docker image removes the
+# pre-installed python packages from this ppa.
+rm -f /etc/apt/sources.list.d/jonathonf-ubuntu-python-3_6-xenial.list
+
+ln -sf /usr/bin/python3.5 /usr/bin/python3 # Py36 has issues with add-apt
```
lgeiger: This is all magic adapted from https://github.com/tensorflow/addons/blob/master/tools/ci_build/builds/release_linux.sh

.github/workflows/release.yml
```diff
@@ -0,0 +1,100 @@
+name: lce-release
+
+# on:
+#   release:
+#     tags:
+#       - v*
+on: [push]
```
lgeiger: Before merging we should set this to only run when pushing a tag

.github/workflows/release.yml
```diff
@@ -0,0 +1,100 @@
...
+            .github/tools/release_linux.sh
+
+          sudo apt-get -y -qq install patchelf --no-install-recommends
+          python -m pip install auditwheel==2.0.0
+
+          for f in artifacts/*.whl; do
+            auditwheel show $f || true
+            # auditwheel repair fails likely due similar reasons as
+            # https://github.com/tensorflow/addons/pull/435
+            # auditwheel repair --plat manylinux2010_x86_64 $f
```
lgeiger: Not sure about this, but wheels seem to work fine without it

.github/workflows/unittests.yml
```diff
@@ -77,34 +27,21 @@ jobs:
...
-      - name: "Python Test Dependency: Build LCE PIP Package"
-        run: bazel-bin/build_pip_pkg artifacts
-      - name: "Python Test Dependency: Install LCE PIP Package"
-        run: pip3 install larq pytest pytest-xdist artifacts/*.whl
-      - name: "Run Python Tests"
-        run: |
-          # Go to a different directory to force it to use the installed package
-          # and not use the local larq_compute_engine directory
-          cd larq_compute_engine/tflite
-          python3 -m pytest . -n auto
```
lgeiger: We should refactor those tests to use the new converter at some point.

.github/workflows/unittests.yml
```diff
@@ -77,34 +27,21 @@ jobs:
...
 
   ARM:
     runs-on: ubuntu-latest
-    container:
-      image: tensorflow/tensorflow:custom-op-ubuntu16
-
```
lgeiger: This should speedup CI

.github/workflows/unittests.yml
```diff
@@ -116,24 +53,31 @@ jobs:
         run: ./bazelisk run larq_compute_engine/tflite/tests:cc_tests_aarch64_qemu --config=aarch64
 
   MLIR:
-    runs-on: ubuntu-latest
-    container:
-      image: tensorflow/tensorflow:custom-op-ubuntu16
-
+    runs-on: macos-latest
```
lgeiger: I did some speed testing and by default the MLIR build took [42:39 min](https://github.com/larq/compute-engine/runs/441958454). Without the container, without the manylinux2010 toolchain and using `--distinct_host_configuration=false`, I can get it down to [31:32 min](https://github.com/larq/compute-engine/runs/443167125?check_suite_focus=true). Building using macOS is the fastest though with [19:24 min](https://github.com/larq/compute-engine/runs/443064715?check_suite_focus=true).

WORKSPACE
```diff
@@ -1,9 +1,5 @@
 workspace(name = "larq_compute_engine")
 
-load("//third_party/tf:tf_configure.bzl", "tf_configure")
-
-tf_configure(name = "local_config_tf")
-
```
lgeiger: This was only necessary for building the custom TF ops.

configure.sh
```diff
@@ -23,147 +23,58 @@ function write_to_bazelrc() {
...
-
-
 if [[ "$PIP_MANYLINUX2010" == "1" ]]; then
   if is_linux; then
     write_to_bazelrc "build --config=manylinux2010"
     write_to_bazelrc "test --config=manylinux2010"
   fi
+fi
+
+if ! is_windows; then
```
lgeiger: I spend ~10min trying to compile it on Windows, but without look. Though these changes should at least make the configure script Windows compatible.

lgeiger:
I tested the wheel generation at https://github.com/larq/compute-engine/actions/runs/38699565

.github/workflows/release.yml
Tombana: :+1: , but we should still have several people with different OS versions check the wheel.

.github/workflows/unittests.yml
Tombana: :+1:  And probably just use the pre-built pip package instead of recompiling the entire converter.

.github/workflows/release.yml
lgeiger: Agreed. I fairly am confident about the wheels though, since auditwheel only complains about our own shared library which we definitely need in the wheel.

.github/workflows/unittests.yml
```diff
@@ -116,24 +53,31 @@ jobs:
...
+      - name: Run converter tests TF 2.0
+        run: ./bazelisk run larq_compute_engine/mlir:converter_test --distinct_host_configuration=false
+      - name: Install TF 1.15
+        run: pip install tensorflow==1.15 --no-cache
+      - name: Run converter tests TF 1.15
+        run: ./bazelisk run larq_compute_engine/mlir:converter_test --distinct_host_configuration=false
+      - name: Install TF 1.14
+        run: pip install tensorflow==1.14 --no-cache
+      - name: Run converter tests TF 1.14
+        run: ./bazelisk run larq_compute_engine/mlir:converter_test --distinct_host_configuration=false
```
lgeiger: We could parallelize this in the future, though then we'd need to build a wheel first and store it in an artifact.

.github/workflows/unittests.yml
Tombana: Note that the `configure.sh` script, in case of `manylinux2010`, adds some `crosstool_top` flag to `.bazelrc` that refers to a hardcoded path in that docker container. However, since this CI run only uses `--config=rpi3` and `--config=aarch64`, those flags are overwritten.
I'm indeed in favor of removing the container here, but maybe we should add a comment that this means we can only use bazel commands with one of those configs and not the default one!

.github/workflows/unittests.yml
Tombana: I think this is fine.
However, after installing a different tensorflow version, technically you should run `./configure.sh` again because it will update some TF paths. I think its fine in this case because all versions are installed to the same path here.

However, for TF 1.14 especially, we should run it again but with `no` to the manylinux2010 question.

lgeiger: > ./configure.sh again because it will update some TF paths. I think its fine in this case because all versions are installed to the same path here.

I removed the path configs from the ./configure.sh script since they are only needed when building custom ops that dynamically link against TensorFlow

configure.sh
```diff
@@ -212,7 +123,7 @@ build:nohdfs --define=no_hdfs_support=true
 build:nonccl --define=no_nccl_support=true
 
 build --config=v2 --config=noaws --config=nogcp --config=nohdfs --config=nonccl
-test --config=v2 --config=noaws --config=nogcp --config=nohdfs --config=nonccl
+test --config=v2
```
Tombana: What does this do?

lgeiger: This was added by mistake and will throw warnings since it is already configured on the build. This makes sure we don't build TensorFlow with those features since we don't need them.

.github/workflows/unittests.yml
lgeiger: I am not sure if I follow what you mean exactly, do you want to add this comment in a followup PR?

Tombana: I meant that either in this PR or a future PR, it might be nice to have a comment in this `unittests.yml` file, with something like:
```
# Note that the default bazel config will not work here because this does not run in the docker container. Only the `rpi3` and `aarch64` config options are valid here.
```

lgeiger: Just for reference, you can build it outside the container without the manylinux2010 toolchaing with no problem: https://github.com/larq/compute-engine/actions/runs/38647228/workflow#L67

Tombana: Ah yes thats true. The hardcoded toolchain path only gets added for manylinux2010.

PR #189, opened 2020-02-14
DOC: remove manylinux1 from docs

arashb:
since we dropped the support for TF <1.14.

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -9,12 +9,9 @@ the build process for each of these components.
 
 ### Setup Docker container ###
 We will build the LCE inside a [Docker](https://www.docker.com/) container.
-The image that you use depends on the version of TensorFlow:
-
-- To build the LCE for TensorFlow 2.x (`manylinux2010` compatible)
-  use `custom-op-ubuntu16`
-- To build the LCE for TensorFlow 1.x (not `manylinux2010` compatible)
-  use `custom-op-ubuntu14`
+To be able to build the LCE and the LCE converter's PIP package
+for TensorFlow 2.x  (`manylinux2010` compatible), we use `custom-op-ubuntu16`
+image.
 
 You can clone the LCE repository in a directoy of the
 host machine and mount that directroy as a
@@ -52,8 +49,8 @@ brew install bazelbuild/tap/bazelisk
 
 ### Configure .bazelrc ###
 Run the ```./configure.sh``` script in the root directory and answer
-"Yes" to the ```manylinux2010``` question when you want to build for
-TensorFlow 2.x or TensorFlow 1.15, and "No" for TensorFlow 1.14 and below.
+"Yes" to the ```manylinux2010``` question if you want to build the
+LCE converter's PIP package for TensorFlow 2.x.
 
 ## LCE for TensorFlow Lite ##
 LCE for TensorFlow Lite has a diverse platform support, covering
--- a/docs/build.md
+++ b/docs/build.md
@@ -9,9 +9,10 @@ the build process for each of these components.
 
 ### Setup Docker container ###
 We will build the LCE inside a [Docker](https://www.docker.com/) container.
-To be able to build the LCE and the LCE converter's PIP package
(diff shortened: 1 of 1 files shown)


arashb:
@lgeiger just to be sure, the LCE converter PIP package can not be built outside the docker, right?

docs/build.md
```diff
@@ -52,8 +49,8 @@ brew install bazelbuild/tap/bazelisk
 
 ### Configure .bazelrc ###
 Run the ```./configure.sh``` script in the root directory and answer
-"Yes" to the ```manylinux2010``` question when you want to build for
-TensorFlow 2.x or TensorFlow 1.15, and "No" for TensorFlow 1.14 and below.
+"Yes" to the ```manylinux2010``` question if you want to build the
```
Tombana: I guess then we can also remove the yes/no question from the configure script? That will avoid any confusion.

arashb: as far as I remember, you need the manylinux2010 if you want to build the PIP package in the container but activating by default will cause build issues on macOS.

arashb: on mac, does not even ask for it. so for mac is removed! 😀 then I guess we can set it to the default for linux builds, @lgeiger ?

lgeiger:
Thanks for removing the ubuntu 14 container.

> @lgeiger just to be sure, the LCE converter's PIP package can not be built outside the docker and after removing the TF ops the PIP package is the only reason we have the container now, right?

The pip package builds fine outside the container, see [this build](https://github.com/larq/compute-engine/runs/443167125?check_suite_focus=true) using [this build config](https://github.com/larq/compute-engine/actions/runs/38647228/workflow#L55-L72) without the `manylinux2010` flag. Since we don't dynamically link against TensorFlow anymore (as required for the custom ops) we don't require matching C++ toolchains. So if people want to build the pip package from source outside the container they should disable the `manylinux2010` flag and make sure to have a C++14 compatible compiler installed. We still need the flag for the prebuilt wheels in order to make sure they work on most linux systems.

docs/build.md
```diff
@@ -9,12 +9,9 @@ the build process for each of these components.
...
 We will build the LCE inside a [Docker](https://www.docker.com/) container.
-The image that you use depends on the version of TensorFlow:
-
-- To build the LCE for TensorFlow 2.x (`manylinux2010` compatible)
-  use `custom-op-ubuntu16`
-- To build the LCE for TensorFlow 1.x (not `manylinux2010` compatible)
-  use `custom-op-ubuntu14`
+To be able to build the LCE and the LCE converter's PIP package
+for TensorFlow 2.x  (`manylinux2010` compatible), we use `custom-op-ubuntu16`
+image.
```
lgeiger: What about:
```markdown
To be able to build LCE and the LCE converter's [`manylinux2010`](https://www.python.org/dev/peps/pep-0571/) compatible PIP package, we use the [`tensorflow/tensorflow:custom-op-ubuntu16`](https://hub.docker.com/r/tensorflow/tensorflow) image. 
For macOS, make sure to have a C++ compiler installed.
```

docs/build.md
lgeiger: > I guess then we can also remove the yes/no question from the configure script? That will avoid any confusion.

We still need it if people want to build the converter outside the container.

docs/build.md
```diff
@@ -52,8 +49,8 @@ brew install bazelbuild/tap/bazelisk
 
 ### Configure .bazelrc ###
 Run the ```./configure.sh``` script in the root directory and answer
-"Yes" to the ```manylinux2010``` question when you want to build for
-TensorFlow 2.x or TensorFlow 1.15, and "No" for TensorFlow 1.14 and below.
+"Yes" to the ```manylinux2010``` question if you want to build the
+LCE converter's PIP package for TensorFlow 2.x.
```
lgeiger:
```suggestion
LCE converter's PIP package inside the `tensorflow/tensorflow:custom-op-ubuntu16` container.
```

PR #198, opened 2020-02-14
(ci-skip): Add code of conduct

lgeiger:
This is the code of conduct we also use in our other open source repos.


AdamHillier:
Is there a particular reason that this isn't in the root of the project, as it is [in Larq](https://github.com/larq/larq/blob/master/CODE_OF_CONDUCT.md)?

PR #202, opened 2020-02-14
Update build docs

Tombana:
Adds example on how to build `lce_minimal` using the Makefile system.

### DIFF
--- a/docs/build_arm.md
+++ b/docs/build_arm.md
@@ -16,9 +16,9 @@ the following order:
    install Bazel, see
    [Cross-compiling LCE with Make](#cross-compiling-lce-with-make).
 
-This guide will show you how to build the [LCE example application](../examples/lce_minimal.cc).
+This guide will show you how to build the [LCE example program](../examples/lce_minimal.cc).
 See [here](./inference.md) to find out how you can create your own LCE
-inference application.
+inference binary.
 
 NOTE: Although the Raspberry Pi 3 and Raspberry Pi 4 have 64-bit CPUs, the
 officially supported OS Raspbian for the Raspberry Pi is a 32-bit OS. In order
@@ -29,10 +29,10 @@ as [Manjaro](https://manjaro.org/download/#raspberry-pi-4-xfce) should be used.
 
 First configure Bazel using the instructions [here](build.md#configure-bazelrc).
 
-To cross-compile the LCE example application for ARM architectures, the bazel
+To cross-compile the LCE example for ARM architectures, the bazel
 target needs to be built with the `--config=rpi3` (32-bit ARM) or
 `--config=aarch64` (64-bit ARM) flag. For example, to build the example
-application for 64-bit ARM systems, run the following command from the LCE root
+for 64-bit ARM systems, run the following command from the LCE root
 directory:
 ```bash
 bazel build \
--- a/docs/build_arm.md
+++ b/docs/build_arm.md
@@ -1,5 +1,5 @@
 # Building Larq Compute Engine for ARM-based systems
-This page descibes how to build a Larq Compute Engine (LCE) inference binary
+This page descibes how to build Larq Compute Engine (LCE) applications
 for 32-bit, as well as 64-bit ARM-based systems. You can either build LCE
 natively on your ARM device or cross-compile it from a non-ARM machine.
 [Bazel](https://bazel.build/) is the primary build system for LCE, and it can
@@ -7,8 +7,8 @@ be used to cross-compile a binary for ARM-based systems using a non-ARM host.
 To natively build on an ARM system itself we provide a script that uses the
 Makefile build system.
 
-This leaves us with three ways to build LCE, which we recommend in the
-following order:
+This leaves us with three ways to build LCE applications, which we recommend in
+the following order:
 1. To cross-compile LCE from a host machine, see the section
    [Cross-compiling LCE with Bazel](#cross-compiling-lce-with-bazel)
 2. To natively compile LCE, see the section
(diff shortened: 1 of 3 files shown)


docs/build_arm.md
```diff
@@ -31,20 +32,28 @@ use the optimized 64-bit kernels of LCE on a Raspberry Pi, a 64-bit OS such as
 First configure Bazel using the instructions [here](build.md). Make sure you
 have run the `./configure.sh` script as instructed.
 
-To cross-compile an LCE inference binary for ARM architectures, the bazel
+To cross-compile the LCE example application for ARM architectures, the bazel
 target needs to be built with the `--config=rpi3` (32-bit ARM) or
-`--config=aarch64` (64-bit ARM) flag. For example, to build the LCE benchmark
-tool, run the following command from the LCE root directory:
+`--config=aarch64` (64-bit ARM) flag. For example, to build the example
```
arashb: why do we need to have two different examples here? The example itself is not what matters, its about building and the building for both examples, beside just one flag, is the same.

Tombana: True. I'll just leave the `lce_minimal` example and then in the text state the name of the benchmark bazel target.

docs/build_arm.md
```diff
@@ -1,14 +1,14 @@
...
 [Bazel](https://bazel.build/) is the primary build system for LCE, and it can
 be used to cross-compile a binary for ARM-based systems using a non-ARM host.
 To natively build on an ARM system itself we provide a script that uses the
 Makefile build system.
 
-This leaves us with three ways to build LCE, which we recommend in the
-following order:
+This leaves us with three ways to build LCE applications, which we recommend in
+the following order:
 1. To cross-compile LCE from a host machine, see the section
```
arashb:
```suggestion
1. To cross-compile LCE from a host machine, see section
```

docs/build_arm.md
```diff
@@ -1,14 +1,14 @@
...
    [Cross-compiling LCE with Bazel](#cross-compiling-lce-with-bazel)
 2. To natively compile LCE, see the section
```
arashb:
```suggestion
2. To natively compile LCE, see section
```

docs/build_arm.md
```diff
@@ -17,9 +17,10 @@ following order:
    install Bazel, see the section
```
arashb:
```suggestion
   install Bazel, see section
```

docs/build_arm.md
```diff
@@ -17,9 +17,10 @@ following order:
...
    [Cross-compiling LCE with Make](#cross-compiling-lce-with-make).
 
-In this guide, we use the [LCE benchmark tool](../larq_compute_engine/tflite/benchmark)
-source to build an inference binary. See [here](./inference.md) to find out how
-you can create your own custom LCE inference application.
+This guide will show you how to build the [LCE example application](../examples/lce_minimal.cc)
+and the [LCE benchmark tool](../larq_compute_engine/tflite/benchmark).
```
arashb: I don't think we need to give two examples here. just showing  example LCE application is enough. Building LCE benchmark tool does not add any new values or insight.

docs/build_arm.md
```diff
@@ -17,9 +17,10 @@ following order:
...
+See [here](./inference.md) to find out how you can create your own LCE
+inference application.
 
 Although the Raspberry Pi 3 and Raspberry Pi 4 have 64-bit CPUs, note that the
```
arashb:
```suggestion
NOTE: the Raspberry Pi 3 and Raspberry Pi 4 have 64-bit CPUs, note that the
```

docs/build_arm.md
```diff
@@ -17,9 +17,10 @@ following order:
...
 popular distribution Raspbian for the Raspberry Pi is a 32-bit OS. In order to
```
arashb:
```suggestion 
officially supported Raspberry Pi operating system "Raspbian" is a 32-bit OS. In order to
```

docs/build_arm.md
```diff
@@ -31,20 +32,24 @@ use the optimized 64-bit kernels of LCE on a Raspberry Pi, a 64-bit OS such as
 First configure Bazel using the instructions [here](build.md). Make sure you
 have run the `./configure.sh` script as instructed.
```
arashb:
```suggestion
First configure Bazel using the instructions [here](build.md#configure-bazelrc).
```

docs/build_arm.md
```diff
@@ -1,14 +1,14 @@
 # Building Larq Compute Engine for ARM-based systems
-This page descibes how to build a Larq Compute Engine (LCE) inference binary
+This page descibes how to build Larq Compute Engine (LCE) applications
 for 32-bit, as well as 64-bit ARM-based systems. You can either build LCE
 natively on your ARM device or cross-compile it from a non-ARM machine.
 [Bazel](https://bazel.build/) is the primary build system for LCE, and it can
 be used to cross-compile a binary for ARM-based systems using a non-ARM host.
 To natively build on an ARM system itself we provide a script that uses the
 Makefile build system.
```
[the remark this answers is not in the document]
arashb: I am strongly against using the word "application". Application has different meaning (see [here](https://en.wikipedia.org/wiki/Application_software)). The right word here is "binary" or "executable". Even the TF lite benchmark guide use the word "binary".

docs/quickstart_android.md
```diff
@@ -1,6 +1,6 @@
 # Larq Compute Engine: Android Quickstart
```
arashb: The only change I approve in this file is adding the location where the lce_benchmark_model binary is stored after building with bazel. I disagree with rest of the changes.

docs/build_arm.md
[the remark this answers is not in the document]
Tombana: I'll change it back to 'binary'.

I did have that particular meaning of the word application in mind, because the developers that write a program with LCE might want to use it to build an actual Android app, for example where LCE processes the camera input. In that case, application would be correct I think.

docs/build_arm.md
Tombana: I think in this case we have to leave out the word 'section'. When there's a number, then you leave out the "the", for example "see Section 3".

PR #212, opened 2020-02-17
Add converter docs

lgeiger:


docs/converter.md
```diff
@@ -0,0 +1,30 @@
+# LCE Converter
+
+LCE allows to convert a Keras model built with `larq` to the TensorFlow Lite FlatBuffer format for inference.
```
arashb:
```suggestion
LCE converter allows to convert a `Larq's` Keras model to an LCE-compatible TensorFlow Lite FlatBuffer format for inference.
```

Tombana: I would personally stick with "Keras model built with larq" . I like the "LCE-compatible" part.

```suggestion
The LCE converter allows you to convert a Keras model built with `larq` to an LCE-compatible TensorFlow Lite FlatBuffer format for inference.
```

PR #221, opened 2020-02-17
Fix all the links

lgeiger:
This should fix all the links on docs.larq.dev

Sorry about the big diff, my editor didn't like the markdown formatting.

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -1,4 +1,4 @@
-# Build Larq Compute Engine #
+# Build Larq Compute Engine
 
 The Larq Compute Engine (LCE) repository consists of two main components:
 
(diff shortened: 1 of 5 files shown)


docs/build.md
```diff
@@ -72,14 +72,14 @@ LCE converter's PIP package inside the `tensorflow:custom-op-ubuntu16`
 container. This script generates the Bazel configuration file `.bazelrc`
 in the LCE root directory.
 
-## Build LCE Runtime ##
+## Build LCE Runtime
 
 LCE runtime has a diverse platform support, covering
-[Android](./quickstart_android.md) and [ARM-based boards](./build_arm.md)
+[Android](/compute-engine/quickstart_android) and [ARM-based boards](/compute-engine/build_arm)
```
leonoverweel: no `https://docs.larq.dev` prefix here?

lgeiger: Since the docs will most likely be accessed through the website we can omit it since it will automatically link to the right page

PR #222, opened 2020-02-17
Fix Bi-RealNet link

Tombana:


koenhelwegen:
Conflicts with #223 which fixes two other links as well

PR #240, opened 2020-02-24
add QuickNet API links

koenhelwegen:
To be merged after https://github.com/larq/docs/pull/20.


koenhelwegen:
Well spotted! Looks like we made the replacement in the `h5` files but not in the architecture implementation in zoo. We should correct that.

PR #247, opened 2020-02-25
Use vectors instead of tflite temp buffers

arashb:


larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -67,17 +67,16 @@ typedef struct {
...
   int32_t im2col_index;
-  int32_t padding_buffer_index;
-  int32_t bitpacked_weights_buffer_index;
 
-  bool padding_cache_filled = false;
-  bool bitpacked_weights = false;
+  std::vector<std::uint8_t> padding_buffer;
+  std::vector<std::uint8_t> bitpacked_weights_buffer;
+
+  bool is_weight_bitpacked = false;
```
arashb: these variables need to be right beside their corresponding buffer:
```
.. padding_buffer;
.. is_padding_correction_cached = false;

.. bitpacked_weights_buffer;
.. is_weight_bitpacked = false;
```
so someone who reads the code understands there is a related bool that needs to be updated after changing anything related to this buffer.

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -223,144 +222,60 @@ TfLiteStatus Prepare(KernelType kernel_type, const int bitwidth,
          conv_params->dilations[1] /* height */ != 1 ||
          conv_params->filter_width != 1 || conv_params->filter_height != 1);
 
-    // Figure out how many temporary buffers we need
-    int temporaries_count = 0;
```
arashb: keep the `temporaries_count` variable as before. if we decided to add a new temp buffer we don't need to add it back again.

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -223,144 +222,60 @@ TfLiteStatus Prepare(KernelType kernel_type, const int bitwidth,
...
     if (conv_params->need_im2col) {
-      conv_params->im2col_index = temporaries_count++;
-    }
-    if (conv_params->padding_type == TfLitePadding::kTfLitePaddingSame) {
-      conv_params->padding_buffer_index = temporaries_count++;
-    }
-    conv_params->bitpacked_weights_buffer_index = temporaries_count++;
+      conv_params->im2col_index = 0;
```
arashb: what was the im2co_index before? why we are hardcoding the index here?

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -223,144 +222,60 @@ TfLiteStatus Prepare(KernelType kernel_type, const int bitwidth,
...
 
-    // Allocate int array of that size
-    TfLiteIntArrayFree(node->temporaries);
-    node->temporaries = TfLiteIntArrayCreate(temporaries_count);
+      // Allocate int array of that size
```
arashb: really no need for changes in this few lines here!

arashb: I can not understand why we need these changes here!

PR #252, opened 2020-02-27
Add support for one-padding

Tombana:
This adds support for setting the attribute `pad_values = 1`.

### DIFF
--- a/larq_compute_engine/tflite/kernels/bconv2d.cc
+++ b/larq_compute_engine/tflite/kernels/bconv2d.cc
@@ -55,6 +55,7 @@ typedef struct {
   // padding
   TfLitePadding padding_type{};
   TfLitePaddingValues padding_values{};
+  int padding_value = 0;  // Must be 0 or 1
 
   // output tensor dimensions
   int64_t out_width{0};
@@ -125,12 +126,23 @@ void* Init(TfLiteContext* context, const char* buffer, size_t length) {
     conv_params->dilations[i] = dilation_vector[i].AsInt64();
 
   // reading padding
-  conv_params->padding_type =
-      m["padding"].ToString() == "VALID" ||
-              m["padding"].ToString() ==
-                  "valid"  // TODO: not sure if this check is needed
-          ? TfLitePadding::kTfLitePaddingValid
-          : TfLitePadding::kTfLitePaddingSame;
+  if (m["padding"].ToString() == "VALID" ||
+      m["padding"].ToString() == "valid") {
+    conv_params->padding_type = kTfLitePaddingValid;
+    conv_params->padding_value = 0;
+  } else if (m["padding"].ToString() == "SAME" ||
+             m["padding"].ToString() == "same") {
+    conv_params->padding_type = kTfLitePaddingSame;
+    conv_params->padding_value = 0;
+  } else if (m["padding"].ToString() == "SAME_ONE" ||
+             m["padding"].ToString() == "same_one") {
+    conv_params->padding_type = kTfLitePaddingSame;
+    conv_params->padding_value = 1;
+  } else {
+    context->ReportError(context, "Invalid padding attribute.");
+    return conv_params;
+  }
+
   return conv_params;
 }
 
@@ -379,7 +391,8 @@ void EvalOpt(TfLiteContext* context, TfLiteNode* node,
                              : nullptr;
 
   if (!params->is_padding_correction_cached &&
-      params->padding_type == TfLitePadding::kTfLitePaddingSame) {
+      params->padding_type == TfLitePadding::kTfLitePaddingSame &&
+      params->padding_value == 0) {
     // In the first run, fill the cache
     using PaddingFunctor =
         ce::core::PaddingFunctor<T, T, ce::core::FilterFormat::OHWI>;
@@ -453,7 +466,7 @@ void EvalOpt(TfLiteContext* context, TfLiteNode* node,
       GetTensorData<float>(fused_multiply), GetTensorData<float>(fused_add),
       GetTensorShape(output), GetTensorData<T>(output), GetTensorShape(im2col),
       GetTensorData<T>(im2col), params->bitpack_before_im2col,
-      params->padding_buffer.data(),
+      params->padding_buffer.data(), params->padding_value,
       CpuBackendContext::GetFromContext(context));
 }
 
--- a/larq_compute_engine/tflite/kernels/bconv2d.cc
+++ b/larq_compute_engine/tflite/kernels/bconv2d.cc
@@ -55,7 +55,7 @@ typedef struct {
   // padding
   TfLitePadding padding_type{};
   TfLitePaddingValues padding_values{};
-  int padding_value = 0;  // Must be 0 or 1
+  int pad_value = 0;  // Must be 0 or 1
 
   // output tensor dimensions
   int64_t out_width{0};
@@ -129,17 +129,19 @@ void* Init(TfLiteContext* context, const char* buffer, size_t length) {
   if (m["padding"].ToString() == "VALID" ||
       m["padding"].ToString() == "valid") {
     conv_params->padding_type = kTfLitePaddingValid;
-    conv_params->padding_value = 0;
   } else if (m["padding"].ToString() == "SAME" ||
              m["padding"].ToString() == "same") {
     conv_params->padding_type = kTfLitePaddingSame;
-    conv_params->padding_value = 0;
-  } else if (m["padding"].ToString() == "SAME_ONE" ||
-             m["padding"].ToString() == "same_one") {
-    conv_params->padding_type = kTfLitePaddingSame;
-    conv_params->padding_value = 1;
   } else {
     context->ReportError(context, "Invalid padding attribute.");
+    return conv_params;
+  }
+
+  conv_params->pad_value =
+      m["pad_values"].IsNull() ? 0 : m["pad_values"].AsInt64();
+
+  if (conv_params->pad_value != 0 && conv_params->pad_value != 1) {
+    context->ReportError(context, "Attribute pad_values must be 0 or 1.");
     return conv_params;
   }
 
@@ -392,7 +394,7 @@ void EvalOpt(TfLiteContext* context, TfLiteNode* node,
 
   if (!params->is_padding_correction_cached &&
       params->padding_type == TfLitePadding::kTfLitePaddingSame &&
-      params->padding_value == 0) {
+      params->pad_value == 0) {
     // In the first run, fill the cache
     using PaddingFunctor =
         ce::core::PaddingFunctor<T, T, ce::core::FilterFormat::OHWI>;
@@ -466,7 +468,7 @@ void EvalOpt(TfLiteContext* context, TfLiteNode* node,
       GetTensorData<float>(fused_multiply), GetTensorData<float>(fused_add),
       GetTensorShape(output), GetTensorData<T>(output), GetTensorShape(im2col),
       GetTensorData<T>(im2col), params->bitpack_before_im2col,
-      params->padding_buffer.data(), params->padding_value,
+      params->padding_buffer.data(), params->pad_value,
       CpuBackendContext::GetFromContext(context));
 }
 
--- a/larq_compute_engine/tflite/tests/bconv2d_test.cc
+++ b/larq_compute_engine/tflite/tests/bconv2d_test.cc
@@ -43,8 +43,6 @@ const char* GetPaddingName(enum Padding padding) {
       return "VALID";
     case Padding_SAME:
       return "SAME";
-    case Padding_ONE:
-      return "SAME_ONE";
     default:
       return "UNKNOWN";
   };
@@ -207,12 +205,16 @@ struct TestParam {
     param_dilation_oss << param.dilation_height_factor << "x"
                        << param.dilation_width_factor;
 
+    const int pad_values = (param.padding == Padding_ONE ? 1 : 0);
+    const Padding padding =
+        (param.padding == Padding_ONE ? Padding_SAME : param.padding);
+
     // WARNING: substitute accests only 11 arguments
-    return absl::Substitute("Op$0_I$1_K$2_P$3_S$4_D$5_T$6", param.kernel_name,
-                            param_input_oss.str(), param_filter_oss.str(),
-                            GetPaddingName(param.padding),
-                            param_strides_oss.str(), param_dilation_oss.str(),
-                            param.num_threads);
+    return absl::Substitute("Op$0_I$1_K$2_P$3_PV$4_S$5_D$6_T$7",
+                            param.kernel_name, param_input_oss.str(),
+                            param_filter_oss.str(), GetPaddingName(padding),
+                            pad_values, param_strides_oss.str(),
+                            param_dilation_oss.str(), param.num_threads);
   }
 
   int input_batch_count = 1;
--- a/larq_compute_engine/tflite/tests/bconv2d_test.cc
+++ b/larq_compute_engine/tflite/tests/bconv2d_test.cc
@@ -243,7 +243,7 @@ class BaseBConv2DOpModel : public SingleOpModel {
   BaseBConv2DOpModel(register_function registration, const TensorData& input,
                      const TensorData& filter, const TensorData& output,
                      int stride_width = 1, int stride_height = 1,
-                     enum Padding padding = Padding_VALID,
+                     enum Padding padding = Padding_VALID, int pad_values = 0,
                      int dilation_width_factor = 1,
                      int dilation_height_factor = 1, int num_threads = -1) {
     input_ = AddInput(input);
@@ -270,6 +270,7 @@ class BaseBConv2DOpModel : public SingleOpModel {
       });
       fbb.String("filter_format", "OHWI");
       fbb.String("padding", GetPaddingName(padding));
+      fbb.Int("pad_values", pad_values);
     });
     fbb.Finish();
     SetCustomOp("LqceBconv2d", fbb.GetBuffer(), registration);
@@ -371,8 +372,11 @@ TEST_P(BConv2DOpTest, SimpleTest) {
   const Padding padding = param.padding;
   const int num_threads = param.num_threads;
 
-  const Padding tflite_padding =
+  const Padding builtin_padding =
       (padding == Padding_ONE ? Padding_VALID : padding);
+  const Padding bconv_padding =
+      (padding == Padding_ONE ? Padding_SAME : padding);
+  const int pad_values = (padding == Padding_ONE ? 1 : 0);
 
   const int input_num_elem =
       input_batch_count * input_height * input_width * input_depth;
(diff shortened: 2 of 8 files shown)


Tombana:
As discussed with @lgeiger I'm fine with using `SAME_ONE` as attribute. (The distinction is not super important since users should never see that stuff, its only set by the converter).

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -34,6 +35,15 @@
 
 namespace tflite {
 
+constexpr int Padding_ONE = Padding_MAX + 1;
+
+const char* GetPaddingName(enum Padding padding) {
+  if (padding == Padding_VALID) return "VALID";
+  if (padding == Padding_SAME) return "SAME";
+  if (padding == Padding_ONE) return "ONE";
+  return "UNKNOWN";
```
lgeiger: What do you think about this?
```suggestion
  switch (padding) {
    case Padding_VALID: return "VALID";
    case Padding_SAME: return "SAME";
    case Padding_ONE: return "ONE";
    default: return "UNKNOWN";
  }
```

(50% of me is just looking for an excuse to try out the new GitHub multi-line suggestions 😉)

Tombana: I applied the change in my own commit instead of the github one, because the linter wanted it on separate lines.

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -34,6 +35,21 @@
 
 namespace tflite {
 
+constexpr int Padding_ONE = Padding_MAX + 1;
```
arashb: what is `Padding_MAX`?

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -351,13 +368,64 @@ TEST_P(BConv2DOpTest, SimpleTest) {
...
-  std::generate(std::begin(input_data), std::end(input_data), rand_generator);
+  std::generate(std::begin(bconv_input_data), std::end(bconv_input_data),
+                rand_generator);
   std::generate(std::begin(filters_data), std::end(filters_data),
                 rand_generator);
   std::generate(std::begin(channel_multipliers), std::end(channel_multipliers),
                 float_generator);
   std::generate(std::begin(bias_data), std::end(bias_data), float_generator);
 
+  int tflite_input_height = input_height;
```
arashb: this should be the exact calculation that Larq does to pad the QuantConv with ones and in Larq i believe the TF padding op is used so this also needs to use the same TF ops and the same padding layout (otherwise how Larq is used the train the weights will be different than how the inference engine is implemented) so instead of manually calculating the padding sizes which might differ from how Larq does it, we need to use the TF ops that Larq uses here. I have implemented example of using C++ API of TF ops in this test.

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -193,7 +208,7 @@ class BaseBConv2DOpModel : public SingleOpModel {
   BaseBConv2DOpModel(register_function registration, const TensorData& input,
                      const TensorData& filter, const TensorData& output,
                      int stride_width = 1, int stride_height = 1,
-                     enum Padding padding = Padding_VALID,
+                     const char* padding_str = "VALID",
```
arashb: I would rather pass an `enum` as before and when I need a string of enum use the utility function above to get its string value.

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -318,21 +333,23 @@ TEST_P(BConv2DOpTest, SimpleTest) {
...
+
   const int input_num_elem =
       input_batch_count * input_height * input_width * input_depth;
 
   const int filters_num_elem =
       filter_height * filter_width * input_depth * filter_count;
 
   using T = float;
-  std::vector<T> input_data, filters_data;
+  std::vector<T> bconv_input_data, tflite_input_data, filters_data;
```
arashb: by `tflite_input_data` you mean the `builtin_op_input_data`? why the builtin op and bconv should have different inputs?

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -125,12 +129,25 @@ void* Init(TfLiteContext* context, const char* buffer, size_t length) {
     conv_params->dilations[i] = dilation_vector[i].AsInt64();
 
   // reading padding
-  conv_params->padding_type =
-      m["padding"].ToString() == "VALID" ||
-              m["padding"].ToString() ==
-                  "valid"  // TODO: not sure if this check is needed
-          ? TfLitePadding::kTfLitePaddingValid
-          : TfLitePadding::kTfLitePaddingSame;
+  if (m["padding"].IsNull() || m["padding"].ToString() == "VALID" ||
```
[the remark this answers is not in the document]
Tombana: :+1:  I thought that the default in TF was "VALID" when no padding was specified, but [now i see](https://www.tensorflow.org/api_docs/python/tf/nn/conv2d) that there's no default value, so `Null` should indeed be an error.

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -55,6 +55,10 @@ typedef struct {
   // padding
   TfLitePadding padding_type{};
   TfLitePaddingValues padding_values{};
+  // We could copy `TfLitePadding` to include an extra value kTfLitePaddingOne.
+  // However this requires copying a lot of other code like im2col, so it is
+  // easier to simply have one extra flag here.
+  bool one_padding = false;
```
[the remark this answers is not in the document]
Tombana: I tried that first (I have the code in a separate branch) but this requires copying a lot of tflite code such as the im2col algorithm. It became quite involved, and in my opinion it was less readable than this. However I'm still open to this idea, let me know what you think.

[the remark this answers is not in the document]
arashb: why do you need to copy im2col if you subclass the struct?

larq_compute_engine/tflite/tests/bconv2d_test.cc
Tombana: The enum defines `Padding_MAX` as the maximum value of things in the enum, so that `MAX + 1` is always "unused".
We could change their enum to our own, but I think this will require changing a lot of code (I have not tested this particular change).

larq_compute_engine/tflite/tests/bconv2d_test.cc
Tombana: I agree that its better to use TF ops. However I could only find this `SingleOpModel` testing interface where you can not give it two ops in a row, thats why I did it manually.

Do you know of a way that we can build a multi-op model in this testing framework?

larq_compute_engine/tflite/tests/bconv2d_test.cc
Tombana: `builtin_op_input_data` is indeed a better name :+1: .

The reason for different inputs is that the tflite input needs to be padded with +1s. (see other comment above)

larq_compute_engine/tflite/tests/bconv2d_test.cc
arashb: in case that there is not multi-op model, you can pass the output of one op as an input to another single op model? but can you check first if there is really no multi-op model? how TF lite tests the fused ops?

larq_compute_engine/tflite/kernels/bconv2d.cc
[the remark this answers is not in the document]
Tombana: So there is `TfLitePadding` (defined in [builtin_op_data.h](https://github.com/tensorflow/tensorflow/blob/844b64f2eed4fb6ddc7e893d8baba0e35abd16dd/tensorflow/lite/c/builtin_op_data.h#L45) ) which is an `enum`.
This is used by `ComputePaddingHeightWidth` in `Prepare`.

Then inside `ConvParams` there is `PaddingType padding_type;` . This is a `class` defined in [types.h](https://github.com/tensorflow/tensorflow/blob/844b64f2eed4fb6ddc7e893d8baba0e35abd16dd/tensorflow/lite/kernels/internal/types.h#L28).

During `EvalOpt` we convert the `TfLitePadding` value to the `PaddingType` value in `ConvParams`.

So I tested the following options:

- Copy `TfLitePadding` to `LcePadding` and `PaddingType` to `LcePaddingType` . For both cases, add the extra option for one-padding.
*Pros*: we have a proper third padding type everywhere
*Cons*: because we change `PaddingType` to `LcePaddingType` we also have to copy `ConvParams` to `BConvParams`. Since im2col accepts only `ConvParams` we have to copy im2col to accept `BConvParams`.

- Change our `TfLiteBConv2DParams` to subclass the original `ConvParams`
  The original `ConvParams` has the `PaddingType` which we can not set to one-padding. Therefore, in our subclass we add `LcePadding true_padding_type;` which has support for one-padding.
*Pros*: We no longer have to copy to `ConvParams` every `Eval` step, because our class is already a `ConvParams` subclass. I benchmarked it but it does not provide a measurable speedup, but I do think its cleaner because it avoids the copy.
This does not require copying im2col or whatever so thats pretty nice.
*Cons*: its a bit confusing because then we have `params->padding_type` (from the base class) which can only be valid or same, and we have `params->true_padding_type` which can also be One.
Furthermore, when you call `ComputePaddingHeightWidth` you have to still convert `true_padding_type` back to `TfLitePaddingType` and send `ONE` to `SAME` to get the right output size.


Actually I liked this second option a lot, but it was a pretty huge diff compared to master (because things like `dilations` now moved from our `conv_params` into the base class). The current PR is a much smaller and more readable diff.

How about this: I first push a PR where i make our struct a subclass of ConvParams and then I push another PR where I add the one-padding on top of that.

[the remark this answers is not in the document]
arashb: > So there is `TfLitePadding` (defined in [builtin_op_data.h](https://github.com/tensorflow/tensorflow/blob/844b64f2eed4fb6ddc7e893d8baba0e35abd16dd/tensorflow/lite/c/builtin_op_data.h#L45) ) which is an `enum`.
> This is used by `ComputePaddingHeightWidth` in `Prepare`.
> 
> Then inside `ConvParams` there is `PaddingType padding_type;` . This is a `class` defined in [types.h](https://github.com/tensorflow/tensorflow/blob/844b64f2eed4fb6ddc7e893d8baba0e35abd16dd/tensorflow/lite/kernels/internal/types.h#L28).
> 
> During `EvalOpt` we convert the `TfLitePadding` value to the `PaddingType` value in `ConvParams`.

Both `PaddingType` and `TfLitePadding` are `enums` so you can not inhert them but you can use the trick explained here: https://www.codeproject.com/Articles/16150/Inheriting-a-C-enum-type

arashb:
I am seeing in https://github.com/larq/larq/pull/438 @lgeiger uses an additional argument to pass the padding values which is default to 0. I think for bconv we should use the same interface as well. basically in this way the padding type stays the same however, the padding default varlue for padding=SAME will be set according to our needs.

lgeiger:
> I think for bconv we should use the same interface as well. basically in this way the padding type stays the same however, the padding default varlue for padding=SAME will be set according to our needs. We can check if the default value is zero or not, it its not then we do all the extra stuff that needs to be done.

I don't have a strong opinion on that, but I think I'd prefer to stick with `padding="SAME_ONE" `. For `larq` we opted to have an additional `pad_values` argument in order to easily support experimentation with different padding values, where as for compute engine we'll only ever support plus or minus one padding and we don't need to add an additional kernel argument.

arashb:
> > I think for bconv we should use the same interface as well. basically in this way the padding type stays the same however, the padding default varlue for padding=SAME will be set according to our needs. We can check if the default value is zero or not, it its not then we do all the extra stuff that needs to be done.
> 
> I don't have a strong opinion on that, but I think I'd prefer to stick with `padding="SAME_ONE" `. For `larq` we opted to have an additional `pad_values` argument in order to easily support experimentation with different padding values, where as for compute engine we'll only ever support plus or minus one padding and we don't need to add an additional kernel argument.

my problem with `padding="SAME_ONE"` is that that its not really a new padding type but its just changing the default value of padding elements in the **SAME** padding type. Defining it as a new padding-type is confusing and it causing hacky changes in the design of the software.

larq_compute_engine/tflite/tests/bconv2d_test.cc
Tombana: New commit uses two `SingleOpModel`s. We still have to manually compute the padding amounts since the `Pad` op wants explicit paddings.

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -125,12 +129,25 @@ void* Init(TfLiteContext* context, const char* buffer, size_t length) {
...
+    conv_params->one_padding = false;
+  } else if (m["padding"].ToString() == "SAME" ||
+             m["padding"].ToString() == "same") {
+    conv_params->padding_type = kTfLitePaddingSame;
+    conv_params->one_padding = false;
+  } else if (m["padding"].ToString() == "SAME_ONE" ||
+             m["padding"].ToString() == "same_one") {
+    // For computation of output size and for im2col,
+    // the tflite type need to be set to Same.
+    conv_params->padding_type = kTfLitePaddingSame;
```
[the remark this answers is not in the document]
Tombana: Fixed with the separate `pad_values` attribute.

Tombana:
Rebased on master to be able to do the MLIR one-padding PR on top of this.

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -125,12 +126,25 @@ void* Init(TfLiteContext* context, const char* buffer, size_t length) {
...
+  } else {
+    context->ReportError(context, "Invalid padding attribute.");
+    return conv_params;
+  }
+
+  conv_params->pad_value =
+      m["pad_values"].IsNull() ? 0 : m["pad_values"].AsInt64();
+
+  if (conv_params->pad_value != 0 && conv_params->pad_value != 1) {
```
arashb: 1. the pad value can be any number (not only 0 and 1) and during the bitpacking, we automatically pack it to the right bit (0 is the problematic pad value which we need to check explicitly). 
2. currently we don't have negative numbers padding support and thats why we have this check?

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -219,7 +269,8 @@ class BaseBConv2DOpModel : public SingleOpModel {
         fbb.Int(1);
       });
       fbb.String("filter_format", "OHWI");
-      fbb.String("padding", padding == Padding_VALID ? "VALID" : "SAME");
+      fbb.String("padding", GetPaddingName(padding));
```
arashb: `GetPaddingName` still supports `SAME_ONE` but `BConv` op does not support that type and instead we have the pad_values. so we might pass the wrong argument to `BConv` op by using `GetPaddingName` utility function right?

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -318,18 +369,23 @@ TEST_P(BConv2DOpTest, SimpleTest) {
...
   const int dilation_width_factor = param.dilation_width_factor;
-  // TOOD: currently we are ignoring padding
   const Padding padding = param.padding;
   const int num_threads = param.num_threads;
 
+  const Padding builtin_padding =
+      (padding == Padding_ONE ? Padding_VALID : padding);
+  const Padding bconv_padding =
+      (padding == Padding_ONE ? Padding_SAME : padding);
+  const int pad_values = (padding == Padding_ONE ? 1 : 0);
```
arashb: I don't understand why we still need to `Padding_ONE` here!

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -358,6 +414,49 @@ TEST_P(BConv2DOpTest, SimpleTest) {
...
+        {0, 0, overflow_top, overflow_bottom, overflow_left, overflow_right, 0,
+         0},
+        1.0f, {TensorType_FLOAT32, {}});
+
+    padop.SetInput(input_data);
+    padop.Invoke();
+
+    padded_input_height = input_height + overflow_top + overflow_bottom;
+    padded_input_width = input_width + overflow_left + overflow_right;
+    EXPECT_THAT(padop.GetOutputShape(),
```
arashb: here we are somehow testing the `Padding` op? if this test fails then the `one-padding` will be shown as failure but thats actually a different problem. so I think we don't need this test here?

larq_compute_engine/tflite/tests/bconv2d_test.cc
Tombana: Correct. I still left it in `GetPaddingName` because its used in the testing-string.
It might be cleaner to change `GetPaddingName` to only have `VALID` and `SAME` and use a seperate thing `pv0` & `pv1` (pad value 0/1) in the testing-string.
What do you think?

larq_compute_engine/tflite/kernels/bconv2d.cc
Tombana: `pad_value` is the value that is written in the pixels around the image by the im2col algorithm.
(i.e. its the exact same value as in the larq layers   larq/larq#438 )

The im2col algorithm is done after bitpacking, so it can only write binary 0 (i.e. float +1, `pad_value=1`) or binary 1 (i.e. float -1, `pad_value=-1`).

When `pad_value=0` then we have zero-padding, and we write `+1` around the image and then have the zero-padding-correction-functor to fix it.  When `pad_value=1` then we have one-padding, so we *also* write `+1` around the image and then we dont have a correction.

We could also add support for `pad_value=-1` but the MLIR converter will always give us models with `+1` so i think this is cleaner.

larq_compute_engine/tflite/tests/bconv2d_test.cc
Tombana: We want to test all three cases, so this `Padding_ONE` is simply used to be able to "set up the test" so to speak. It won't be passed to any of the ops. Thats why we directly define `bconv_padding` and `builtin_padding` which is what is passed to the op.

Another option is that we add another testing-parameter called `pad_value` but then it will also be done in combination with `VALID` so then it runs way more unit tests than whats needed, and it doesnt make so much sense.
(And also I think we're already using the maximum amount of testing parameters).

larq_compute_engine/tflite/tests/bconv2d_test.cc
Tombana: This is meant to verify that if the padding op somehow changes, then we will notice it here, so we will know that its not caused by our own op.

larq_compute_engine/tflite/kernels/bconv2d.cc
arashb: what about the cases we do the im2col before bitpacking?

larq_compute_engine/tflite/tests/bconv2d_test.cc
arashb: yes, it should be removed from the padding type and add a separate substring for for tests.

larq_compute_engine/tflite/kernels/bconv2d.cc
Tombana: In that case, im2col happens in "float space" and we could choose which value we want to pad with (the `zero_byte = 0x00`). But then the bitpacking is applied and it will turn it into binary 0 or 1. So in that case we have effectively padded with -1 or +1 as well.

PR #253, opened 2020-02-28
Simplify is binary check

lgeiger:
## What do these changes do?
This PR will directly match a MLIR `ConstantOp` so we can simplify the binary check.

## How Has This Been Tested?
MLIR unit tests

### DIFF
--- a/larq_compute_engine/mlir/tests/prepare-tf.mlir
+++ b/larq_compute_engine/mlir/tests/prepare-tf.mlir
@@ -12,7 +12,8 @@ func @fuse_bsign(%arg0: tensor<8x16xf32>) -> tensor<8x16xf32> {
   // CHECK-NEXT: return %0
 }
 
-func @bconv2d(%arg0: tensor<1x112x112x2xf32>) -> tensor<1x112x112x2xf32> {
+// CHECK-LABEL: @fuse_bconv2d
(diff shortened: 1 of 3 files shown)


larq_compute_engine/mlir/tests/prepare-tf.mlir
```diff
@@ -12,7 +12,7 @@ func @bsign(%arg0: tensor<8x16xf32>) -> tensor<8x16xf32> {
 }
 
 func @bconv2d(%arg0: tensor<1x112x112x2xf32>) -> tensor<1x112x112x2xf32> {
-  %cst = constant dense<[[[[1.0, -1.0], [1.0, 1.0]], [[-1.0, 1.0], [-1.0, 1.0]]]]> : tensor<1x2x2x2xf32>
+  %cst = "tf.Const"() { value = dense<[[[[1.0, -1.0], [1.0, 1.0]], [[-1.0, 1.0], [-1.0, 1.0]]]]> : tensor<1x2x2x2xf32> } : () -> tensor<1x2x2x2xf32>
```
lgeiger: This change is not required to make the tests pass, but I think it is a good idea to test that a `tf.Const` correctly gets matched by the transform as well since the transform no matches a MLIR `ConstantOp` instead of the TFLite and TF versions.

PR #255, opened 2020-02-28
Cleanup MLIR tests

lgeiger:
This should make the MLIR tests more readable and introduces a consistent naming and checking scheme.

This tries to follow the best practices mentioned at https://mlir.llvm.org/getting_started/TestingGuide/

### DIFF
--- a/larq_compute_engine/mlir/tests/prepare-tf.mlir
+++ b/larq_compute_engine/mlir/tests/prepare-tf.mlir
@@ -1,35 +1,41 @@
 // RUN: lce-tf-opt %s -tfl-prepare-lce | FileCheck %s --dump-input-on-failure
 
-func @bsign(%arg0: tensor<8x16xf32>) -> tensor<8x16xf32> {
+// CHECK-LABEL: @fuse_bsign
+func @fuse_bsign(%arg0: tensor<8x16xf32>) -> tensor<8x16xf32> {
   %0 = "tf.Sign"(%arg0) : (tensor<8x16xf32>) -> tensor<8x16xf32>
   %cst = constant dense<0.1> : tensor<f32>
   %2 = "tf.AddV2"(%0, %cst) : (tensor<8x16xf32>, tensor<f32>) -> tensor<8x16xf32>
   %3 = "tf.Sign"(%2) : (tensor<8x16xf32>) -> tensor<8x16xf32>
   return %3 : tensor<8x16xf32>
-// CHECK-LABEL: bsign
-// CHECK:  %0 = "tf.LqceBsign"(%arg0) : (tensor<8x16xf32>) -> tensor<8x16xf32>
-// CHECK:  return %0
+
+  // CHECK-NEXT: %0 = "tf.LqceBsign"(%arg0)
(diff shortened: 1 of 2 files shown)


larq_compute_engine/mlir/tests/prepare-tf.mlir
```diff
@@ -1,35 +1,41 @@
...
-// CHECK:  %[[CST1:.*]] = constant dense<-2.000000e+00> : tensor<2xf32>
-// CHECK:  %[[CST2:.*]] = constant dense<4.000000e+00> : tensor<2xf32>
-// CHECK:  %[[TRP:.*]] = "tf.Transpose"
-// CHECK:  %[[CONV:.*]] = "tf.LqceBconv2d64"(%arg0, %[[TRP]], %[[CST1]], %[[CST2]]) {data_format = "NHWC", dilations = [1, 1, 1, 1], filter_format = "OHWI", padding = "SAME", strides = [1, 1, 1, 1]} : (tensor<1x112x112x2xf32>, tensor<2x1x2x2xf32>, tensor<2xf32>, tensor<2xf32>) -> tensor<1x112x112x2xf32>
-// CHECK:  return %[[CONV]]
+
+  // CHECK: %cst = constant
+  // CHECK: %[[fused_multiply:.*]] = constant dense<-2.000000e+00> : tensor<2xf32>
+  // CHECK: %[[fused_add:.*]] = constant dense<4.000000e+00> : tensor<2xf32>
+  // CHECK-NEXT: %[[transpose:.*]] = "tf.Transpose"(%cst
```
AdamHillier: What is the difference between `CHECK` and `CHECK-NEXT`?

lgeiger: [`CHECK-NEXT`](https://llvm.org/docs/CommandGuide/FileCheck.html#the-check-next-directive) verifies that matches happen on exactly consecutive lines with no other lines in between them. So it is a bit stricter than just using `CHECK` here.

Tombana: From the docs that you linked to:

>  A “CHECK-NEXT:” cannot be the first directive in a file.

It looks like you should first do one `CHECK` and then use `CHECK-NEXT` for the following ones. The tests in `optimize.mlir` seem to start with `CHECK-NEXT`.

lgeiger: I think this is fine since `CHECK-LABEL` is the first directive on any of the tests. E.g. see [this tflite test](https://github.com/tensorflow/tensorflow/blob/1d8b9fcff2951dc9b8eb0f66d3c1b45e7226e48d/tensorflow/compiler/mlir/lite/tests/legalize-tf.mlir#L66-L73) which uses the same approach.

Tombana: Is there a closing bracket missing from this transpose line? Does the FileCheck system only do partial-matches or something and therefore it passes?

lgeiger: This is intentional since we only want to partially match the first argument.

PR #256, opened 2020-02-28
Update BConv TableGen definition to match TFLite Op

lgeiger:
This makes sure that the TableGen op definition matches the TFLite op implementation.

### DIFF
--- a/larq_compute_engine/mlir/ir/lce_ops.td
+++ b/larq_compute_engine/mlir/ir/lce_ops.td
@@ -44,7 +44,7 @@ TODO
     TF_AnyStrAttrOf<["SAME", "VALID"]>:$padding,
     DefaultValuedAttr<TF_ConvnetDataFormatAttr, "NHWC">:$data_format,
     DefaultValuedAttr<I64ArrayAttr, "{1, 1, 1, 1}">:$dilations,
-    DefaultValuedAttr<TF_AnyStrAttrOf<["HWIO", "OHWI", "OHWI_PACKED"]>, "OHWI">:$filter_format
+    DefaultValuedAttr<TF_AnyStrAttrOf<["OHWI", "OHWI_PACKED"]>, "OHWI">:$filter_format
   );
 
   let results = (outs
--- a/larq_compute_engine/mlir/ir/lce_ops.td
+++ b/larq_compute_engine/mlir/ir/lce_ops.td
@@ -44,7 +44,7 @@ TODO
     TF_AnyStrAttrOf<["SAME", "VALID"]>:$padding,
     DefaultValuedAttr<TF_ConvnetDataFormatAttr, "NHWC">:$data_format,
     DefaultValuedAttr<I64ArrayAttr, "{1, 1, 1, 1}">:$dilations,
-    DefaultValuedAttr<TF_AnyStrAttrOf<["HWIO", "OHWI"]>, "OHWI">:$filter_format
+    DefaultValuedAttr<TF_AnyStrAttrOf<["HWIO", "OHWI", "OHWI_PACKED"]>, "OHWI">:$filter_format
   );
 
   let results = (outs
--- a/larq_compute_engine/mlir/ir/lce_ops.td
+++ b/larq_compute_engine/mlir/ir/lce_ops.td
@@ -35,25 +35,21 @@ TODO
   }];
 
   let arguments = (ins
-    TensorOf<[BF16, F16, F32, F64, I32]>:$input,
-    TensorOf<[BF16, F16, F32, F64, I32]>:$filter,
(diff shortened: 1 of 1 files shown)


larq_compute_engine/mlir/ir/lce_ops.td
```diff
@@ -35,25 +35,21 @@ TODO
...
     TensorOf<[F32]>:$fused_multiply,
     TensorOf<[F32]>:$fused_add,
 
     I64ArrayAttr:$strides,
-    TF_AnyStrAttrOf<["SAME", "VALID", "EXPLICIT"]>:$padding,
+    TF_AnyStrAttrOf<["SAME", "VALID"]>:$padding,
     DefaultValuedAttr<TF_ConvnetDataFormatAttr, "NHWC">:$data_format,
     DefaultValuedAttr<I64ArrayAttr, "{1, 1, 1, 1}">:$dilations,
-    DefaultValuedAttr<TF_AnyStrAttrOf<["HWIO", "OHWI", "OHWI_PACKED"]>, "HWIO">:$filter_format
+    DefaultValuedAttr<TF_AnyStrAttrOf<["HWIO", "OHWI"]>, "OHWI">:$filter_format
```
arashb: `OHWI_PACKED` was reserved for the time we start to bitpack the weights during the conversion. so I think we should still keep it.

lgeiger: Since it is not yet supported and the tablegen definition doesn't change the forward compatibility of the flatbuffer, I don't see a reason to keep it, or am I missing something?

arashb: I am starting to work on this next week. having bitpacked weights is necessary for other projects that we will pursue soon. so why should we remove it now and then add again next week?

lgeiger:
```suggestion
    DefaultValuedAttr<TF_AnyStrAttrOf<["HWIO", "OHWI", "OHWI_PACKED"]>, "OHWI">:$filter_format
```

lgeiger: I don't have a strong opinion on this. I added it back in 350d437

Tombana:
`"HWIO"` is not supported by our op so *that* is the one that should be removed.

larq_compute_engine/mlir/ir/lce_ops.td
```diff
@@ -35,25 +35,21 @@ TODO
...
     TensorOf<[F32]>:$fused_multiply,
     TensorOf<[F32]>:$fused_add,
 
     I64ArrayAttr:$strides,
-    TF_AnyStrAttrOf<["SAME", "VALID", "EXPLICIT"]>:$padding,
+    TF_AnyStrAttrOf<["SAME", "VALID"]>:$padding,
     DefaultValuedAttr<TF_ConvnetDataFormatAttr, "NHWC">:$data_format,
     DefaultValuedAttr<I64ArrayAttr, "{1, 1, 1, 1}">:$dilations,
-    DefaultValuedAttr<TF_AnyStrAttrOf<["HWIO", "OHWI", "OHWI_PACKED"]>, "HWIO">:$filter_format
+    DefaultValuedAttr<TF_AnyStrAttrOf<["HWIO", "OHWI", "OHWI_PACKED"]>, "OHWI">:$filter_format
```
Tombana:
```suggestion
    DefaultValuedAttr<TF_AnyStrAttrOf<["OHWI", "OHWI_PACKED"]>, "OHWI">:$filter_format
```

arashb: but we still support it in the RefImpl. right? we are planning to remove the RefImpl. (it is not used anyway) but its not remove *yet* so maybe just remove it with the same PR which removes the RefImpl.?

Tombana: Currently in `Prepare` we already assert that its OHWI, [see this line](https://github.com/larq/compute-engine/blob/1071647e0b045e6b9fe5666a55a367cca6aff99e/larq_compute_engine/tflite/kernels/bconv2d.cc#L175) .
And the RefImp also assumes OHWI now (although it probably still has some code for HWIO support somewhere).

arashb: but RefImpl do not support OHWI format? so if someone just go and change to the ref. impl. for lets say only testing purposes the code will not work! so the current code is already inconsistent with itself and we should not have that assert at the first place. I am gonna send a PR to cleanup this mess.

Tombana: The Ref Impl also has never supported dilations for example. I think we should just remove it asap.

PR #259, opened 2020-03-03
Add core tests back to CI

arashb:
## What do these changes do?

The core tests for some reason were removed from CI. In this PR I do the followings:
- adding back the Core Tests to CI
- removing the im2col tests where the impl. was originally removed in #208 

## How Has This Been Tested?

N/A
## Benchmark Results

N/A
## Related issue number

N/A


Tombana:
I think the only test that is still relevant is "packbits_tests". 

The other tests ("bconv2d_tests", "bgemm_tests", "fused_bgemm_tests") are for outdated code that is not used anymore.

arashb:
> I think the only test that is still relevant is "packbits_tests".
> 
> The other tests ("bconv2d_tests", "bgemm_tests", "fused_bgemm_tests") are for outdated code that is not used anymore.

the `bgemm_functor` is used in the `bgemm_impl_ref.h' which is the ref impl. for RUY and it will not be removed (only the ref. impl. that uses the TF op impl. will be removed). The `bconv2d` and  `fused_bgemm_functor` will be removed in a future PR.

PR #260, opened 2020-03-03
Split fused multiply into backtransform and post multiply

Tombana:
## What do these changes do?

This PR is needed for a next PR where we support fused activation functions. We will then have the following computations applied at the end of the bgemm kernel:
- backtransformation `x = n - 2 * x` (can be done in `int32`)
- fused activation `x = relu(x)`  (can be done in `int32` or `float`)
- post-multiply and post-add `x = a + b * x`  (must be done in `float`. Comes from potential BatchNorm layers or other `Mul` and `Add` layers).

Currently there is no fused activation, so the backtransform and post-multiply&add are currently merged into one multiply&add.  This PR will split those two, so that its ready for fused activations.

The PR also renames `fused_{add,multiply}` to `post_{add,multiply}`, to make it extra clear that this transformation comes *after* the activation function (which is usually the final thing in an op).

## How Has This Been Tested?

The C++ unit tests have been updated to no longer include the backtransform into the post-multiply&add.

## Benchmark Results

Since the backtransform and post-multiply&add are now separate, it slightly slows things down.

On the Amsterdam Pixel 1 phone, the inference time of quicknet goes up from `22.2 ms` to `22.5 ms` (1.4% slowdown).
However, once we actually fuse the batchnorm and activation, the time becomes `21.3 ms` (4% speedup). Since this speedup will apply to all networks in larq zoo, it is worth it.

As @AdamHillier mentioned, we could always provide a separate bgemm kernel for the case where there is no activation function, and there we can keep the backtransform fused.

### DIFF
--- a/larq_compute_engine/core/bgemm_impl_ruy.h
+++ b/larq_compute_engine/core/bgemm_impl_ruy.h
@@ -54,8 +54,8 @@ struct BGemmImplUsingRuy {
 
     TSpec spec;
     spec.backtransform_add = params.backtransform_add;
-    spec.post_multiply = params.post_multiply;
-    spec.post_add = params.post_add;
+    spec.post_activation_multiplier = params.post_activation_multiplier;
+    spec.post_activation_bias = params.post_activation_bias;
 
     // The allocator is used to allocate memory for pre-packed matrices
     ruy::Allocator allocator;
--- a/larq_compute_engine/core/bgemm_impl_ruy.h
+++ b/larq_compute_engine/core/bgemm_impl_ruy.h
@@ -53,8 +53,8 @@ struct BGemmImplUsingRuy {
     dst.data = dst_data;
 
     TSpec spec;
-    spec.fused_multiply = params.fused_multiply;
-    spec.fused_add = params.fused_add;
+    spec.post_multiply = params.post_multiply;
+    spec.post_add = params.post_add;
 
     // The allocator is used to allocate memory for pre-packed matrices
     ruy::Allocator allocator;
--- a/larq_compute_engine/core/bgemm_impl_ruy.h
+++ b/larq_compute_engine/core/bgemm_impl_ruy.h
@@ -53,6 +53,7 @@ struct BGemmImplUsingRuy {
     dst.data = dst_data;
 
     TSpec spec;
+    spec.backtransform_add = params.backtransform_add;
     spec.post_multiply = params.post_multiply;
     spec.post_add = params.post_add;
 
--- a/larq_compute_engine/core/bgemm_kernels_arm64.h
+++ b/larq_compute_engine/core/bgemm_kernels_arm64.h
@@ -89,7 +89,8 @@ using namespace ruy;
 #define RUY_OFFSET_DEPTH 76
 #define RUY_OFFSET_CLAMP_MIN 80
 #define RUY_OFFSET_CLAMP_MAX 84
-#define RUY_OFFSET_FLAGS 88
+#define RUY_OFFSET_BACKTRANSFORM_ADD 88
+#define RUY_OFFSET_FLAGS 92
 
 template <typename Params>
 void CheckOffsetsInKernelParams32BP(const Params&) {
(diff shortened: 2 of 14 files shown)


AdamHillier:
Will we still do the backtransform add/mul with the post-transform add/mul together as one op if there's no relu inbetween?

Tombana:
> Will we still do the backtransform add/mul with the post-transform add/mul together as one op if there's no relu inbetween?

We could do that but I think at least for now we will not. Just like the tflite kernels will always add a bias term, even if there is no bias input tensor.

larq_compute_engine/core/bgemm_kernels_ruy.h
```diff
@@ -91,12 +91,15 @@ struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, DstScalar,
           TBitpacked rhs_val = Element(rhs, k, j);
           accum += xorpopcount<TBitpacked, AccumScalar>(lhs_val, rhs_val);
         }
+        // Backtransform can still be done in int32
+        accum = spec.backtransform_add - 2 * accum;
```
lgeiger: Should we do something like this here, or am I missing an edge case?
```suggestion
        accum = spec.backtransform_add - (accum << 1);
```

AdamHillier: Wouldn't the compiler do this automatically?

Tombana:
I believe the C++ compiler will do that automatically. Additionally, this
is reference code not supposed to be fully optimized but rather should be
very readable.

We should check if this trick can be done in the assembly kernel though :+1:

larq_compute_engine/core/bgemm_impl_ruy.h
```diff
@@ -53,8 +53,9 @@ struct BGemmImplUsingRuy {
     dst.data = dst_data;
 
     TSpec spec;
-    spec.fused_multiply = params.fused_multiply;
-    spec.fused_add = params.fused_add;
+    spec.backtransform_add = params.backtransform_add;
```
arashb: IMO, it would be better to define `pre_actication_BN` stuff and and just fuse the backtransformation into that. In this way, you have implemented pre-actication BN and backtransformation at the same time and there is no need to define explcit  backtranformation attributes  all over the places.

larq_compute_engine/core/bgemm_kernels_arm64.h
```diff
@@ -310,6 +320,16 @@ void BinaryKernelNeonOutOfOrder32BP4x4(
       "ld1 {v6.4s}, [%[rhs_ptr]], #16\n"
       "ld1 {v7.4s}, [%[rhs_ptr]], #16\n"
 
+      // Perform the backtransformation (in int32)
```
arashb: how do we cover the cases that backtransformation needs to be done in float?

larq_compute_engine/core/bgemm_impl_ruy.h
Tombana: I like the `pre_XXX` and `post_XXX` idea, I've been thinking about it too. The `float` vs `int32` is tricky though.
The reason I would rather do the backtransform in `int32` is that it will still work for the `int8` assembly kernels. For `int8`, all the `int32` stuff can stay the same but all the `float` stuff will be *approximated* by these weird exponent+shifts.  So if we change the backtransform to a float, then it would become an approximation in the int8 kernel, and I'd rather keep it as an 'exact' calculation whenever possible.

larq_compute_engine/core/bgemm_kernels_arm64.h
Tombana: See above post. With the `int8` case in mind, I'd rather always do the backtransform as `int32` so its always exact.

larq_compute_engine/core/bgemm_impl_ruy.h
arashb: so based on our phone discussion, since we are not planning to support BT+BN(float)+RELU+BN(FLOAT) the float impl. is not relevant then.

PR #263, opened 2020-03-04
Reduce normal register loads

AdamHillier:
## What do these changes do?
This change swaps some of the (normal, non-SIMD) registers used for intermediate results in the ARM64 binary conv kernels, in such a way that reloading several parameters from memory is no longer necessary.

Two additional registers are used by the kernel (which were previously unused): `x14` and `x15`.

Three `ldr` memory read instructions have been eliminated.

## How Has This Been Tested?
The ARM64 tests pass locally.

## Benchmark Results
Benchmark results (μs) on a Raspberry Pi 4 (1 thread) on QuickNet Large: 78828.5 -> 78580.3.

Results are the average of 500 runs but the results are so close that there probably isn't a meaningful difference here.

### DIFF
--- a/larq_compute_engine/core/bgemm_kernels_arm64.h
+++ b/larq_compute_engine/core/bgemm_kernels_arm64.h
@@ -409,8 +409,8 @@ void BinaryKernelNeonOutOfOrder32BP4x4(
       "50:\n"
       "mov w5, #0\n"
       "51:\n"
-      "ldr w7, [x3, x5, lsl #2]\n"
-      "str w7, [x4, x5, lsl #2]\n"
+      "ldr w13, [x3, x5, lsl #2]\n"
+      "str w13, [x4, x5, lsl #2]\n"
       "add w5, w5, #1\n"
       "cmp w5, w1\n"
       "blt 51b\n"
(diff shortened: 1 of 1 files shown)


AdamHillier:
@larq/compute-engine any idea why the CI isn't running?

lgeiger:
> Any idea why the CI isn't running?

Oh, I think we've configured CI to only run on push, since this is a PR from a fork, it doesn't work.
See https://github.com/larq/zookeeper/pull/62/

lgeiger:
The `MLIR_Python` failure is unrelated. Not sure what's going on with CI there.

AdamHillier:
Now there's different CI failures that are unrelated to the code. The Git checkout is failing for some bizarre reason.

lgeiger:
> The Git checkout is failing for some bizarre reason.

Looks like this might be a problem with the git submodule checkout, though I don't see any failure on master.

lgeiger:
> Looks like this might be a problem with the git submodule checkout, though I don't see any failure on master.

I just rebased and it seems the checkout fine now. After https://github.com/actions/checkout/pull/173 is merged we should upgrade to the new checkout actions which hopefully will be less flaky.

PR #267, opened 2020-03-05
Fused activation

Tombana:
## What do these changes do?
This implements support for fused ReLu activation functions, both in the tflite kernels as well as the MLIR converter. (This will also be required for int8 support later)

## How Has This Been Tested?

Unit tests have been added to the `cc_tests` suite, and MLIR tests have also been added.

## Benchmark Results

Quicknet, Pixel 1 phone

| backtransform | ReLu & BN fusion | padding | time in ms |
| ---- | ---- | ---- | ---- |
| fused with float mul&add | no | zero-padding | 22.2 |
| split from float mul&add | no | zero-padding | 22.4 |
| split from float mul&add | no | one-padding | 22.3 |
| split from float mul&add | yes | one-padding | 21.3 |

So going from zero-padding to one-padding took away `0.1 ms`. The fused relu&BN took away `1.0 ms`.
Overal compared to before splitting the backtransform, it is about a 4% speedup.


larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -135,6 +142,20 @@ void* Init(TfLiteContext* context, const char* buffer, size_t length) {
...
     context->ReportError(context, "Invalid padding attribute.");
+  }
+
+  // Read fused activation
+  if (m["activation"].IsNull() || m["activation"].ToString() == "" ||
+      m["activation"].ToString() == "NONE") {
+    conv_params->activation = kTfLiteActNone;
+  } else if (m["activation"].ToString() == "RELU") {
+    conv_params->activation = kTfLiteActRelu;
+  } else if (m["activation"].ToString() == "RELU1") {
```
lgeiger: Are you sure that `RELU1` is already supported in TF 2.1? I think support in the converter was only added in TF 2.2.

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -146,6 +167,18 @@ void* Init(TfLiteContext* context, const char* buffer, size_t length) {
     return conv_params;
   }
 
+  if (conv_params->padding_type == TfLitePadding::kTfLitePaddingSame &&
+      conv_params->pad_value == 0 &&
```
lgeiger: Should we check the following instead?
```suggestion
      conv_params->pad_value != 1 &&
```

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -146,6 +167,18 @@ void* Init(TfLiteContext* context, const char* buffer, size_t length) {
...
+      conv_params->activation != kTfLiteActNone) {
+    context->ReportError(context,
+                         "Fused activations are not supported in combination "
+                         "with zero-padding.");
```
lgeiger:
```suggestion
                         "Fused activations are only supported with valid or"
                         "one-padding.");
```

larq_compute_engine/mlir/tests/optimize.mlir
```diff
@@ -56,3 +56,68 @@ func @fuse_mul_into_bconv2d(%arg0: tensor<256x32x32x3xf32>, %arg1: tensor<16x3x3
...
+  // CHECK: %0 = "tf.LqceBconv2d64"(%arg0, %arg1, %cst, %cst_0) {activation = "RELU", data_format = "NHWC", dilations = [1, 1, 1, 1], filter_format = "OHWI", pad_values = 1 : i32, padding = "SAME", strides = [1, 1, 1, 1]}
+  // CHECK-NEXT: return %0
+}
+
+
+// CHECK-LABEL: @do_not_fuse_relu_into_bconv2d_padding_same
+func @do_not_fuse_relu_into_bconv2d_padding_same(%arg0: tensor<256x32x32x3xf32>, %arg1: tensor<16x3x3x3xf32>) -> tensor<256x32x32x16xf32> {
+  %post_activation_multiplier = constant dense<1.0> : tensor<16xf32>
+  %post_activation_bias = constant dense<0.0> : tensor<16xf32>
+  %0 = "tf.LqceBconv2d64"(%arg0, %arg1, %post_activation_multiplier, %post_activation_bias) {activation = "NONE", data_format = "NHWC", dilations = [1, 1, 1, 1], filter_format = "OHWI", padding = "SAME", strides = [1, 1, 1, 1]} : (tensor<256x32x32x3xf32>, tensor<16x3x3x3xf32>, tensor<16xf32>, tensor<16xf32>) -> tensor<256x32x32x16xf32>
```
lgeiger: Should we explicity set `pad_values=0`?
```suggestion
  %0 = "tf.LqceBconv2d64"(%arg0, %arg1, %post_activation_multiplier, %post_activation_bias) {activation = "NONE", data_format = "NHWC", dilations = [1, 1, 1, 1], filter_format = "OHWI", pad_values = 0 : i32, padding = "SAME", strides = [1, 1, 1, 1]} : (tensor<256x32x32x3xf32>, tensor<16x3x3x3xf32>, tensor<16xf32>, tensor<16xf32>) -> tensor<256x32x32x16xf32>
```

larq_compute_engine/mlir/tests/optimize.mlir
```diff
@@ -56,3 +56,68 @@ func @fuse_mul_into_bconv2d(%arg0: tensor<256x32x32x3xf32>, %arg1: tensor<16x3x3
...
+  %1 = "tfl.relu"(%0) : (tensor<256x32x32x16xf32>) -> tensor<256x32x32x16xf32>
+  return %1 : tensor<256x32x32x16xf32>
+
+  // CHECK: %0 = "tf.LqceBconv2d64"(%arg0, %arg1, %cst, %cst_0) {activation = "NONE", data_format = "NHWC", dilations = [1, 1, 1, 1], filter_format = "OHWI", padding = "SAME", strides = [1, 1, 1, 1]}
```
lgeiger: See above
```suggestion
  // CHECK: %0 = "tf.LqceBconv2d64"(%arg0, %arg1, %cst, %cst_0) {activation = "NONE", data_format = "NHWC", dilations = [1, 1, 1, 1], filter_format = "OHWI", pad_values = 0 : i32, padding = "SAME", strides = [1, 1, 1, 1]}
```

larq_compute_engine/tflite/kernels/bconv2d.cc
Tombana: Yes that is cleaner. Right now its still fine because in `Init` it is asserted that the pad_value can only be 0 or 1.

larq_compute_engine/tflite/kernels/bconv2d.cc
Tombana: Since all relu's are implemented by clamping, they are trivially supported by the kernel code. They were indeed not yet supported by the converter.

larq_compute_engine/tflite/kernels/bconv2d.cc
Tombana: :+1:  Reminder to self: this should also be changed in the unit test "death test" that expects this error message.

larq_compute_engine/tflite/kernels/bconv2d.cc
lgeiger: Makes sense :+1:

Tombana:
> Very cool. Do you know how much stems from the the relu fusing and how much from the removal of the padding correction?

I just ran some benchmarks to check this, added to the original post.

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -416,20 +429,24 @@ TEST_P(BConv2DOpTest, SimpleTest) {
   std::generate(std::begin(input_data), std::end(input_data), rand_generator);
   std::generate(std::begin(filters_data), std::end(filters_data),
                 rand_generator);
-  std::generate(std::begin(channel_multipliers), std::end(channel_multipliers),
-                float_generator);
-  std::generate(std::begin(bias_data), std::end(bias_data), float_generator);
+  std::generate(std::begin(post_activation_multiplier_data),
+                std::end(post_activation_multiplier_data), float_generator);
```
arashb: why `post_activation_multiplier` can only be these numbers in float_list{0.125, 0.25, 0.75, 0.875, 1.0}? why not create them randomly instead of randomly choosing one of these?

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -416,20 +429,24 @@ TEST_P(BConv2DOpTest, SimpleTest) {
...
+  std::generate(std::begin(post_activation_bias_data),
+                std::end(post_activation_bias_data), float_generator);
```
arashb: the same question about the `post_activatio_multiplier` applies here too! I guess the values can be any value not only the one listed in float_list?

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -535,6 +559,20 @@ INSTANTIATE_TEST_SUITE_P(
         ::testing::Values(1, 2),  // number of threads
         ::testing::ValuesIn(BConv2DOpTest::GetKernelsTuples(*kKernelMap))),
     TestParam::TestNameSuffix);
+
+TEST(BConv2DTests, BConvErrorTest) {
```
arashb: I think this test is unnecessary. it is just going to cover whether one if statement inside the code works or not. I would rather replace this one with a new combination of tests which include the activation (see my comment above)

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -520,7 +543,8 @@ INSTANTIATE_TEST_SUITE_P(
     BConv2DTests, BConv2DOpTest,
     // WARNING: ::testing::Combine accepts max 10 arguments!!!
     ::testing::Combine(
-        ::testing::Values(1),  // batches
+        ::testing::Values(ActivationFunctionType_NONE,
```
arashb: here you are removing the test of of `batches` in order to add the test for activation!!! I do think we should keep the test for it and add a new test which coverers a different combination of testing arguments. maybe you can use test fixture to avoid code duplication.

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -503,15 +512,29 @@ TEST_P(BConv2DOpTest, SimpleTest) {
...
       dilation_width_factor, dilation_height_factor, num_threads);
 
   m_builtin.SetInput(padded_input_data);
   m_builtin.SetFilter(filters_data);
   m_builtin.SetBias(bias_data);
   m_builtin.Invoke();
-
   auto expected_array = m_builtin.GetOutput();
+
+  // Apply the post multiply and add to the tflite model
```
arashb: can't you replace this with a TF Op?

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -80,6 +86,7 @@ typedef struct {
   std::vector<std::uint8_t> bitpacked_weights_buffer;
   bool is_weight_bitpacked = false;
 
+  bool initialized = false;
```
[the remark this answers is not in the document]
Tombana: The op, i.e. whether or not the `Init` function succeeded. What do you think is a good name?
- `init_succeeded`
- `op_initialized`
- `conv_params_initialized`
- `op_ready`
- ?

larq_compute_engine/tflite/tests/bconv2d_test.cc
Tombana: Thanks for spotting this. This is mainly a remnant of the fact that we first fused it with the backtransform. Since the order of multiplications is different than in tflite, some numbers produced float-precision errors, but these nice "round" numbers (at least in base 2) do not.

I will change it to random floats in the [-1,1] interval.

larq_compute_engine/tflite/tests/bconv2d_test.cc
Tombana: I think its important that the op throws an error if you give it wrong parameters. If it doesn't throw an error and the computation is wrong it would be much harder to debug. Especially since this combination "SAME" + fused_relu is something that appears a lot in larq models.
Therefore I think its important to test if this error works as we expect it to. Even though its only one if statement, it check if the error message is being printed and the op doesn't just crash.

> I would rather replace this one with a new combination of tests which include the activation (see my comment above)

What do you mean by this? Test for this error message on all combinations of filter size, strides, dilations, etc?

larq_compute_engine/tflite/tests/bconv2d_test.cc
Tombana: :+1: 

My reasoning was this: the `Combine` thing only accepts 10 arguments and `batches` was always 1, its not like we were "testing for batches". If we copy this stuff to a new test, we would have more code duplication (all the test parameters for example), so that's why I did it like this.

I think you argument is valid, but if we change it, then we should also test for different batch sizes.

Instead of creating multiple tests, what do you think about fusing some other parameters into a single one. For example filter count could be merged with filter height/width, so then instead of
```cpp
::testing::Values(std::array<int, 2>{1, 1},
                  std::array<int, 2>{3, 3},
                  std::array<int, 2>{2, 3}),  // filter height/width
::testing::Values(1, 4, 64),                  // filter count
```
we would have something like this for example:
```cpp
::testing::Values(std::array<int, 3>{1, 1, 1},
                  std::array<int, 2>{4, 3, 3},
                  std::array<int, 3>{64, 3, 3},
                  std::array<int, 3>{4, 2, 3},
                  std::array<int, 3>{64, 2, 3}
),  // filter count,height,width
```

This way we can still use a single test. I'm also fine with creating two separate tests and having some duplication with test parameters, just wanted to get your opinion.

larq_compute_engine/tflite/tests/bconv2d_test.cc
Tombana: Yes but that would be *a lot* of extra code, I think it would make things a lot more unreadable. It would mean all these extra classes at the top of the file, with `MulOp` and `AddOp` and so on. Then we would have all the code to setup the op, set input tensors, read output tensors etc etc. In my opinion this is a lot cleaner.

arashb: ok, agreed 👍

larq_compute_engine/tflite/tests/bconv2d_test.cc
arashb: agreed but in this way only hard-coded combinations will be tested and not all possible combinations! can you check if there is a way to generate these combinations?

larq_compute_engine/tflite/kernels/bconv2d.cc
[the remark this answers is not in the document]
arashb: i prefer `conv_params_initialized` however can you just double check how it`s normally done in other tflite ops? its very strange to me that thats the only way though an error during the `init` process. maybe instead of returning an error they just though an `std::exception`?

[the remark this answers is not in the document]
Tombana: I also find it strange. I think its because for builtin ops, the flatbuffer already contains the correct struct, will all values filled in. Their `Init` functions are always empty. They don't have to parse anything or check for correctness, only custom ops have to do that.

I'll go with `conv_params_initialized`.

PR #268, opened 2020-03-05
Replace mul + add with shl + sub

lgeiger:
## What do these changes do?
This replaces the multiplication and addition in the back transformation with a left shift and a sub. Which in theory might be faster and will free up one register. This can be done since the accumulator will always be positive.

## How Has This Been Tested?
CI tests and benchmarking

## Benchmark Results
I benchmarked this on a Pixel 1 with QuickNet and the speed difference is within the run variation which I think is not surprising.

### DIFF
--- a/larq_compute_engine/core/bgemm_kernels_arm64.h
+++ b/larq_compute_engine/core/bgemm_kernels_arm64.h
@@ -284,9 +284,6 @@ void BinaryKernelNeonOutOfOrder32BP4x4(
       // Load some parameters needed for the end work on current block.
       "ldrb w4, [%[params], #" RUY_STR(RUY_OFFSET_FLAGS) "]\n"
 
-      // Load backtransform multiply (duplicate 4 times into v12)
-      "mov w1, -2\n"
-      "dup v12.4s, w1 \n"
       // Load backtransform add (duplicate 4 times into v13)
       "ldr w1, [%[params], #" RUY_STR(RUY_OFFSET_BACKTRANSFORM_ADD) "]\n"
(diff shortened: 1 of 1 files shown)


arashb:
> ## Benchmark Results
> I benchmarked this on a Pixel 1 with QuickNet and the speed difference is within the run variation which I think is not surprising.

by this you mean you didn't observe any speedup?

lgeiger:
> by this you mean you didn't observe any speedup?

I saw some speedups but they were well within the standard deviation of the different runs so I can't claim a true performance improvement here. This might be more noticable on the different on the Raspberry PI or when benchmarking the kernels itself.

arashb:
> > by this you mean you didn't observe any speedup?
> 
> I saw some speedups but they were well within the standard deviation of the different runs so I can't claim a true performance improvement here. This might be more noticable on the Raspberry PI or when benchmarking the kernels itself.

can you also report the benchmark results in RPi? I don't see a value in changing the code which has the same functionality, do not improve the performance and IMO less readable.

lgeiger:
>  I don't see a value in changing the current code to something which has the same functionality, do not improve the performance and IMO is less readable.

I fully agree :+1: 
I benchmarked this a bit more carefully on my old personal Pixel phone with more iterations:

**Master**
```
count=500 first=26260 curr=26319 min=25725 max=27017 avg=26176.2 std=220
```

**PR**
```
count=500 first=25777 curr=25548 min=25293 max=26716 avg=25751.9 std=242
```

From this testing the speed difference of 0.42 ms is above one sigma so I think this can count as a true improvement.

PR #269, opened 2020-03-05
Add end2end test

Tombana:
## What do these changes do?
This adds an end-to-end CI test.
It can be run with `bazel test larq_compute_engine/mlir:end2end_test`.

It will take a simple larq model, convert it with MLIR, and run it in tflite, and compare the output to the output in TF.

## How Has This Been Tested?
It will test if all the fusions and optimizations in the MLIR converter are correctly done (more extensively than the MLIR tests themselves).
The tflite test will run `Invoke` *twice* to make sure there's no bug where the warmup round is different from a normal round (i.e. the temporary-buffer bug).


larq_compute_engine/mlir/python/end2end_test.py
```diff
@@ -0,0 +1,79 @@
...
+    out = tf.keras.layers.BatchNormalization(
+        gamma_initializer=tf.keras.initializers.RandomNormal(1.0),
+        beta_initializer="uniform",
+    )(out)
+    # A layer with one-padding
+    out = lq.layers.QuantConv2D(
+        filters=16,
+        kernel_size=3,
+        padding="same",
+        # pad_values=1.0,
```
lgeiger: We should enable this once we've released a new version of larq

larq_compute_engine/mlir/python/end2end_test.py
```diff
@@ -0,0 +1,79 @@
...
+from tensorflow.python.eager import context
+
+from larq_compute_engine.mlir.python.converter import convert_keras_model
+
+
+@pytest.fixture
+def eager_mode():
+    """pytest fixture for running test in eager mode"""
+    with context.eager_mode():
+        yield
```
lgeiger: I don't think we need this here since the test will always be run with TensorFlow 2.x which has eager execution enabled by default.
```suggestion
```

larq_compute_engine/mlir/python/end2end_test.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def testnet():
```
lgeiger: If you make this a fixture it will be passed into the test directly:
```suggestion
@pytest.fixture
def model():
```

larq_compute_engine/mlir/python/end2end_test.py
```diff
@@ -0,0 +1,79 @@
...
+    out = tf.keras.layers.BatchNormalization(
+        gamma_initializer=tf.keras.initializers.RandomNormal(1.0),
+        beta_initializer="uniform",
+    )(out)
+    out = tf.keras.layers.GlobalAvgPool2D()(out)
+    return tf.keras.Model(inputs=img_input, outputs=out)
+
+
+def test_simple_model(eager_mode, tmp_path):
+    model = testnet()
```
lgeiger: See above
```suggestion
def test_simple_model(model, tmp_path):
```

larq_compute_engine/mlir/python/end2end_test.py
```diff
@@ -0,0 +1,79 @@
...
+    converted = convert_keras_model(model)
+    modelfile = tmp_path / "end2end_model.tflite"
+    with modelfile.open("wb") as f:
+        f.write(converted)
+
+    x = np.zeros(shape=(1, 4, 4, 1))
```
lgeiger: Should we feed in random data here, or would this be tricky with the C++ side?

larq_compute_engine/mlir/python/end2end_test.py
```diff
@@ -0,0 +1,79 @@
...
+    y = model(x)
+
+    # The model emits 16 floats. Write them to a text file
+    y = np.squeeze(y)
+
+    expected_output_file = tmp_path / "expected_output.txt"
+    with expected_output_file.open("w") as f:
+        for value in y:
+            f.write(f"{value}\n")
```
lgeiger: How about using [`np.savetxt`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.savetxt.html) instead?

larq_compute_engine/mlir/python/end2end_test.py
Tombana: Yes we can do that. Write the random data to a txt file and read that in C++ again. I'll try to do that.

larq_compute_engine/mlir/python/end2end_test.py
```diff
@@ -0,0 +1,72 @@
+import sys
+import pytest
+import larq as lq
+import subprocess
+import tensorflow as tf
+import numpy as np
+from tensorflow.python.eager import context
```
lgeiger:
```suggestion
```

larq_compute_engine/mlir/python/end2end_test.py
```diff
@@ -0,0 +1,70 @@
...
+
+    x = np.zeros(shape=(1, 4, 4, 1))
+    y = model(x)
+
+    # The model emits 16 floats. Write them to a text file
+    y = np.squeeze(y)
+
+    expected_output_file = tmp_path / "expected_output.txt"
+    with expected_output_file.open("w") as f:
+        np.savetxt(f, y, delimiter=" ")
```
lgeiger: This should be enough:
```suggestion
    np.savetxt(expected_output_file, y, delimiter=" ")
```

Tombana: I'll do that together with the random-input commit.

larq_compute_engine/mlir/python/end2end_verify.cc
```diff
@@ -0,0 +1,132 @@
...
+  TFLITE_MINIMAL_CHECK(output_tensor->dims != nullptr);
+  TFLITE_MINIMAL_CHECK(output_tensor->dims->size == 2);
+  TFLITE_MINIMAL_CHECK(output_tensor->dims->data[0] == 1);
+  int num_output_elements = output_tensor->dims->data[1];
+  TFLITE_MINIMAL_CHECK(num_output_elements == (int)output.size());
+
+  // Do inference twice, to check if the warmpup run and internal temporary
+  // buffers work as expected
+  for (int run = 0; run < 2; ++run) {
+    // Fill input buffers
```
arashb: this can be moved outside the for loop. you don't need to fill the input buffer with the same values twice.

Tombana: That depends if the input buffer is used to store intermediate activations, i.e., tflite may share the memory for different tensors.

larq_compute_engine/mlir/python/end2end_test.py
```diff
@@ -0,0 +1,76 @@
...
+import larq as lq
+import subprocess
+import tensorflow as tf
+import numpy as np
+
+from larq_compute_engine.mlir.python.converter import convert_keras_model
+
+
+@pytest.fixture
+def model():
```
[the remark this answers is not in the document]
Tombana: I agree that we can test some larq zoo models.
However we should *also* test this simple model because it explicitly tests for some important things like relu fusion, zero-padding (test if relu's do *not* get fused here), one-padding etc.

Tombana:
> I would rather use [ctypes](https://docs.python.org/2/library/ctypes.html) or [pybind](https://github.com/pybind/pybind11) instead of `subprocess` and storing the test values as a file and passing it to the binary.

:+1:  I'll try to setup pybind.

(I explicitly chose not to build the LCE python wrapper because we don't officially support it (we also dont have a bazel target for it) and this was simpler.)

larq_compute_engine/mlir/python/end2end_verify.cc
arashb: I highly doubt that the tflite invalidates the input buffer (e.g the case the input buffer comes from another part of the code such as image pre-processing). but anyway this is not critical and we keep it the same way as it is.

arashb:
> > I would rather use [ctypes](https://docs.python.org/2/library/ctypes.html) or [pybind](https://github.com/pybind/pybind11) instead of `subprocess` and storing the test values as a file and passing it to the binary.
> 
> 👍 I'll try to setup pybind.
> 
> (I explicitly chose not to build the LCE python wrapper because we don't officially support it (we also dont have a bazel target for it) and this was simpler.)

just to make sure, we understand each other: I don't mean that you have to use tflite_runtime python package or write pybind modules for entire LCE. what needs to be done is very simple process I explain below:
1. replacing the `main` function with a normal C{++} function which appropriate arguments
2. changing the `cc_binary` to `cc_library` so you can load it as a library with pybind
3. define pybind module to call the C function in step 1. 

these 3 steps does not mean we officially support the LCE python wrapper, ... and bazel target will for be trivial.

larq_compute_engine/mlir/python/end2end_test.py
```diff
@@ -0,0 +1,60 @@
...
+
+    out = tf.keras.layers.GlobalAvgPool2D()(out)
+    return tf.keras.Model(inputs=img_input, outputs=out)
+
+
+def test_simple_model(model):
+    converted = convert_keras_model(model)
+
+    inputs = np.random.random((1, 4, 4, 1))
+
```
arashb:
```suggestion
```

larq_compute_engine/mlir/python/end2end_test.py
```diff
@@ -0,0 +1,60 @@
...
+    outputs = model(inputs).numpy()
```
arashb:
```suggestion
    inputs = np.random.random((1, 4, 4, 1))
    outputs = model(inputs).numpy()
```

larq_compute_engine/mlir/python/end2end_test.py
```diff
@@ -0,0 +1,60 @@
+import sys
+import pytest
+import larq as lq
+import tensorflow as tf
+import numpy as np
+
+from larq_compute_engine.mlir.python.converter import convert_keras_model
+
+from larq_compute_engine.mlir._end2end_verify import verify_model
```
arashb:
```suggestion
from larq_compute_engine.mlir.python.converter import convert_keras_model
from larq_compute_engine.mlir._end2end_verify import verify_model
```

larq_compute_engine/mlir/python/end2end_verify.cc
```diff
@@ -0,0 +1,118 @@
...
+    for (int i = 0; i < num_input_elements; ++i) {
+      input_tensor_data[i] = input[i];
+    }
+
+    // Run inference
+    TFLITE_MINIMAL_CHECK(interpreter->Invoke() == kTfLiteOk);
+
+    // Read output buffers
```
arashb: it would be a lot cleaner if you just return the output buffer to the callee instead of return `true` or `false` bc then you use the python assert/test to check the differences. so instead of having a function `verify_model` you have a function to invoke the inference and get the output back in python callee and you compare using whatever testsuite you are using.

Tombana: Yes. However we want to do two runs on the same interpreter, so then we'd have to concatenate the outputs or something. I'm not sure if thats cleaner, what do you think?

Also I like the output thats generated by the C++ code:
image

arashb: I don't understand. you are already comparing only the output of the second inference. why do you need to check the both the first and second? For the output format, I would rather stick to the official output format of the testsuite (we can keep this code commented out for debugging purposes though)

Tombana: Currently we compare both outputs.
I'll try to return a list of outputs.

Tombana: It now returns the outputs to python :+1: 
I still want to check both runs, because we want to catch bugs where somehow a warmup run is different from a normal run (maybe a buffer is not ready yet but we're already using it in the first run).

larq_compute_engine/mlir/python/end2end_verify.cc
```diff
@@ -0,0 +1,99 @@
...
+
+#define MINIMAL_CHECK(x)                                    \
+  if (!(x)) {                                               \
+    std::cerr << "Error at line " << __LINE__ << std::endl; \
+    return result;                                          \
+  }
+
+std::vector<std::vector<float>> runModel(const pybind11::bytes& _flatbuffer,
+                                         const std::vector<float>& input,
+                                         const std::vector<float>& output) {
```
arashb: do you still need the `output` function argument?

lgeiger:
> However, I think we need to move the end2end test outside of 'MLIR_Python' section to its own section bc its not only MLIR related anymore and includes a lot of other things. In addition, this will bring back the CI bottleneck runtime to around 25min again since it can run in parallel with MLIR_Python.

I agree, that we should rename the `MLIR_Python` test, however I don't think we would benefit from running the end2end tests in parallel to the converter tests, since they largely need to build the same requirements so the speedup will be very minor.

Since we now have the end2end test and the MLIR LIT tests, I think the only reason to keep the converter test is that they ensure our Python code works with multiple versions of TensorFlow. What do you think about mocking the `LCE converter` for the Python only tests, so we don't need to build any C++ to run the Python code and then have Python only tests to test the converter against different versions of TensorFlow? However I think this should be done in a future PR to get this merged soon.

arashb:
> > However, I think we need to move the end2end test outside of 'MLIR_Python' section to its own section bc its not only MLIR related anymore and includes a lot of other things. In addition, this will bring back the CI bottleneck runtime to around 25min again since it can run in parallel with MLIR_Python.
> 
> I agree, that we should rename the `MLIR_Python` test, however I don't think we would benefit from running the end2end tests in parallel to the converter tests, since they largely need to build the same requirements so the speedup will be very minor.
> 
> Since we now have the end2end test and the MLIR LIT tests, I think the only reason to keep the converter test is that they ensure our Python code works with multiple versions of TensorFlow. What do you think about mocking the `LCE converter` for the Python only tests, so we don't need to build any C++ to run the Python code and then have Python only tests to test the converter against different versions of TensorFlow? However I think this should be done in a future PR to get this merged soon.

my point is end2end test has nothing to do with MLIR tests. its an integration test for the entire process and right now its buried down inside the MLIR tests which IMO should be moved outside in a separate directory.

PR #271, opened 2020-03-06
Run ARM tests with debug builds. Closes #270.

### LINKED ISSUE #270: Run tests with debug builds
We use the `RUY_DCHECK_EQ` macro in a few places such as here: https://github.com/larq/compute-engine/blob/4877c9b30789f728707586044473b7983ef642e4/larq_compute_engine/core/bgemm_kernels_common.h#L105-L108

This macro is only defined on debug builds (otherwise it's a no-op), but it looks like our tests aren't running debug builds, because I got suspicious that these checks weren't running, and it turns out I was right.

When I add `RUY_DCHECK_EQ(true, false);` to the block above, all the tests still pass. If I change the macro to `RUY_CHECK_EQ` the tests fail as expected.

I'm not sure what the issue is here, presumably something to do with our Bazel config.

AdamHillier:

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -50,13 +50,13 @@ jobs:
         run: ./configure.sh <<< $'yes\n' #yes for manylinux2010
         shell: bash
       - name: "TF Arm32: Cross-compile and run unit tests in qemu"
-        run: ./bazelisk run larq_compute_engine/core/tests:cc_tests_arm32_qemu --config=rpi3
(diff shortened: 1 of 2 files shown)


AdamHillier:
I have verified that this indeed fixes #270 locally, as the line `RUY_DCHECK_EQ(true, false);` causes the tests to fail now, whereas it didn't before.

arashb:
> I have verified that this indeed fixes #270 locally, as the line `RUY_DCHECK_EQ(true, false);` causes the tests to fail now, whereas it didn't before.

nice! can you add this to other tests as well?

PR #274, opened 2020-03-08
remove armv7 android ABI from LCE Lite AAR

arashb:
## What do these changes do?

- remove creating armv7 android ABI from script so no one accidentally use the unoptimized ops ( we still build the x86 ABI since it is used by default for the android emulator in Android Studio) 
- add a new variable for version. @Tombana, @lgeiger how can we automatically sync this version number with release version? 

## How Has This Been Tested?

N/A
## Benchmark Results

N/A
## Related issue number

N/A

### DIFF
--- a/larq_compute_engine/tflite/java/build_lce_aar.sh
+++ b/larq_compute_engine/tflite/java/build_lce_aar.sh
@@ -12,7 +12,7 @@ TMPDIR=`mktemp -d`
 trap "rm -rf $TMPDIR" EXIT
 
 AAR_NAME=lce-lite
-VERSION=0.1.2
+VERSION=$(git describe --tags)
 
 BUILDER=bazel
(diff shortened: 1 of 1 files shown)


lgeiger:
> how can we automatically sync this version number with release version?

We don't have an automated way to update it, but I think for now that's not too bad as long as we  document it well in `CONTRIBUTING.md`.

arashb:
> I've seen the use of `git` commands within such scripts in order to get the correct version number. I'm not sure how to use them properly, I think `git describe --tags` is a good start.
> 
> ```shell
> VERSION=$(git describe --tags)
> ```

great 👍 i tested this and if you run it on the the master branch will return something like this `v0.1.2-14-g4877c9b` which also includes identification for the latest commits. I will add in the `CONTRIBUTING.md` that this needs to be run right after the tag is created.

PR #275, opened 2020-03-09
Cache tensorflow pip dependencies

lgeiger:
## What do these changes do?
This enables caching of PIP dev dependencies following the guide at https://github.com/actions/cache/blob/master/examples.md#python---pip
This should reduce the time spent installing Python dependencies.

## How Has This Been Tested?
CI


lgeiger:
This actually doesn't speed up the pip installation.

We could cache the actual site-packages like in https://github.com/actions/cache/issues/175, but this seems brittle due to our multiple installs of TensorFlow.

PR #276, opened 2020-03-09
Run test using "bazel test" instead of "bazel run"

lgeiger:
This allows tests to be interleaved with building code which has the potential to speedup CI: https://docs.bazel.build/versions/2.0.0/user-manual.html#test

This also unifies how we run tests on CI to make it more consistent.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -50,13 +50,13 @@ jobs:
         run: ./configure.sh <<< $'yes\n' #yes for manylinux2010
         shell: bash
       - name: "TF Arm32: Cross-compile and run unit tests in qemu"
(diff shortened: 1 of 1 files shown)


Tombana:
I originally chose for run because it shows more output. I've had occasions
where the CI logs didn't show what I needed because of 'test'. The parallel
building is nicer though, so if it really speeds up CI then I'm in favor of
test :+1:

arashb:
> This allows tests to be interleaved with building code which has the potential to speedup CI: https://docs.bazel.build/versions/2.0.0/user-manual.html#test
> 
> This also unifies how we run tests on CI to make it more consistent.

the main reason we use `run` instead of `test` is that the qemu tests actually are executed through a bash script not with the a testsuite directly. @Tombana also mentioned thats also why you would not see the outputs properly even by setting the `--test_output=all` flag.

lgeiger:
Makes sense. Closing

PR #277, opened 2020-03-10
Merge some bconv test params (as the # of params is limited).

AdamHillier:
I'm running into the same issue discussed in #267 where the limit of 10 test parameters is problematic, so I've merged the input height/width with the input depth for a single input shape parameter, and likewise merged the filter height/width with the filter count.

I believe the exact same number of tests will be run, with the same bconv parameters.

### DIFF
--- a/larq_compute_engine/tflite/tests/bconv2d_test.cc
+++ b/larq_compute_engine/tflite/tests/bconv2d_test.cc
@@ -158,10 +158,8 @@ namespace testing {
 typedef TfLiteRegistration* (*register_function)(void);
 
 typedef std::tuple<int,                 // batch count
-                   std::array<int, 2>,  // input shape [HW]
-                   int,                 // input depth
-                   std::array<int, 2>,  // filter shape [HW]
-                   int,                 // filter count
+                   std::array<int, 3>,  // input shape [HWI]
+                   std::array<int, 3>,  // filter shape [HWO]
                    std::array<int, 2>,  // strides [HW]
                    std::array<int, 2>,  // dilations [HW]
                    Padding,             // paddding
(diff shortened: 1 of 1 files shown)


larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -519,12 +517,21 @@ INSTANTIATE_TEST_SUITE_P(
     // WARNING: ::testing::Combine accepts max 10 arguments!!!
     ::testing::Combine(
         ::testing::Values(1),  // batches
-        ::testing::Values(std::array<int, 2>{7, 7},
-                          std::array<int, 2>{8, 5}),  // input height/width
-        ::testing::Values(1, 64, 130),                // input depth
-        ::testing::Values(std::array<int, 2>{1, 1}, std::array<int, 2>{3, 3},
-                          std::array<int, 2>{2, 3}),  // filter height/width
-        ::testing::Values(1, 4, 64),                  // filter count
+        ::testing::Values(std::array<int, 3>{7, 7, 1},
```
arashb: see [here](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#how-to-write-value-parameterized-tests) to create generators for your input arguments. you can pass the generators to the `::testing::Combine` to create tuples with them.

Tombana: Do you mean we can nest the `Combine` call?

arashb: > Do you mean we can nest the `Combine` call?

I mean you can pass a `generator` instead of an `std::array` to `Combine`

AdamHillier: @arashb I don't know how to do this, I think it might be easiest if you make a commit on top of this branch/open a PR into this branch with the changes you have in mind?

PR #280, opened 2020-03-10
Switch converter tests to be Python only

lgeiger:
## What do these changes do?
This switches the converter tests to be Python only since the conversion process is now handled by the End2End tests from #269 and the MLIR LIT unittests. The only point of the converter test is now to ensure that our Python code is compatible with multiple TensorFlow version.
This allows us to easily parallelize the testing which should further speed up CI.

## How Has This Been Tested?
CI

### DIFF
--- a/larq_compute_engine/mlir/BUILD
+++ b/larq_compute_engine/mlir/BUILD
@@ -168,14 +168,6 @@ py_library(
     ],
 )
 
-py_test(
-    name = "converter_test",
-    srcs = ["python/converter_test.py"],
-    deps = [
-        ":converter",
-    ],
-)
-
 exports_files([
     "python/converter.py",
     "__init__.py",
--- a/larq_compute_engine/mlir/python/converter_test.py
+++ b/larq_compute_engine/mlir/python/converter_test.py
@@ -1,25 +1,26 @@
 import sys
-import pytest
+import unittest
+from unittest import mock
+
 import larq_zoo as lqz
-from tensorflow.python.eager import context
+
+sys.modules["larq_compute_engine.mlir._graphdef_tfl_flatbuffer"] = mock.MagicMock()
 
 from larq_compute_engine.mlir.python.converter import convert_keras_model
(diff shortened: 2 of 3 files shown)


larq_compute_engine/mlir/BUILD
```diff
@@ -168,14 +168,6 @@ py_library(
...
 )
 
-py_test(
-    name = "converter_test",
-    srcs = ["python/converter_test.py"],
-    deps = [
-        ":converter",
-    ],
-)
-
```
lgeiger: Unfortunately I couldn't get the mocking working with bazel since the Python paths were messed up.

larq_compute_engine/mlir/python/converter_test.py
```diff
@@ -1,25 +1,26 @@
 import sys
-import pytest
+import unittest
```
lgeiger: `pytest` will automatically import `larq_compute_engine/__init__.py` which would break our mocking setup here. This switches the test to pure Python `unittest`.

larq_compute_engine/mlir/python/converter_test.py
```diff
@@ -1,25 +1,26 @@
...
-from larq_compute_engine.mlir.python.converter import convert_keras_model
+sys.modules["larq_compute_engine.mlir._graphdef_tfl_flatbuffer"] = mock.MagicMock()
 
+from larq_compute_engine.mlir.python.converter import convert_keras_model
+from larq_compute_engine.mlir._graphdef_tfl_flatbuffer import (
+    convert_graphdef_to_tflite_flatbuffer as mocked_converter,
+)
 
-@pytest.fixture
-def eager_mode():
```
Tombana: How come you don't need eager mode anymore?

lgeiger: Good catch, added in afb97556640c0ea111b494c6c312acb68ecedeba

PR #281, opened 2020-03-10
Split and expose converter functions

jneeven:
## What do these changes do?
Splits `convert_keras_model` in code that creates the input and output tensors and the graph, and an additional `convert_tensorflow_graph` function. This allows a user to call the LCE converter not only on a keras model, but also on a graph without a surrounding model. 

Example use:
```python
    func = lce.concrete_function_from_keras_model(model)
    graph = convert_variables_to_constants_v2(func, lower_control_flow=False).graph

    # TODO: do some manual graph edits here, adding custom ops etc.
    graph = edit_graph(graph)

    lce.convert_tensorflow_graph(graph, input_tensors, output_tensors)
```

## How Has This Been Tested?
I have used this modified `lce.convert_keras_model` on BinaryResNetE18 from `larq-zoo`, and it still compiles and runs on the Pixel correctly.

*Note: there are some import changes, but this is just `isort` doing its thing.*


larq_compute_engine/mlir/python/converter.py
```diff
@@ -37,46 +39,18 @@ def concrete_function_from_keras_model(model):
...
     # Run a constant folding using grappler since we currently don't implement
     # folding for LCE custom ops
     graph_def = run_graph_optimizations(
         graph_def,
         input_tensors,
         output_tensors,
         config=get_grappler_config(["constfold"]),
-        graph=frozen_func.graph,
+        graph=graph,
     )
```
jneeven: Do we need these optimizations? Removing them would allow even more flexibility, as we could pass in the graph def instead.

larq_compute_engine/__init__.py
```diff
@@ -1,3 +1,11 @@
...
+    concrete_function_from_keras_model,
+    convert_keras_model,
+    convert_tensorflow_graph,
+)
 
-__all__ = ["convert_keras_model"]
+__all__ = [
+    "convert_keras_model",
+    "concrete_function_from_keras_model",
+    "convert_tensorflow_graph",
```
lgeiger: I would be in favour of not exposing this as a public API for now. For example the [`TFLiteConverter`](https://www.tensorflow.org/api_docs/python/tf/lite/TFLiteConverter) in TensorFlow 2.0 only supports `from_concrete_functions`, `from_keras_model` and `from_saved_model`. Which I think it would be good to adhere to that.

I am fine having this as an implementation detail, but I don't think it should be a public API. You could still import it using `larq_compute_engine.mlir.python.converter. convert_tensorflow_graph` if you really need to do some advanced stuff.

jneeven: Okay, in that case I don't think it makes sense to merge this at all. I tried doing some custom graph stuff but ended up needing to change `convert_tensorflow_graph` anyway, so I think if users want to convert a custom graph rather than a model, they will probably write their own function to do so.

jneeven:
Closing this for now as it is unclear whether this would actually be useful. See my last comment.

Issue #86, opened 2019-11-04
Speedup CI builds using caching


lgeiger:
GitHub actions now supports [caching](https://help.github.com/en/github/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows) using [`actions/cache`](https://github.com/actions/cache). I think this could significantly speedup our CI build in particular the TFLite part.

Bazel supports different ways of caching build artifacts either via [remote caching](https://docs.bazel.build/versions/master/remote-caching.html) probably the [GCS backend](https://docs.bazel.build/versions/master/remote-caching.html#google-cloud-storage) is the easiest to setup or simple [disk caching](https://docs.bazel.build/versions/master/remote-caching.html#disk-cache) using the `--disk_cache=/path/to/build/cache` flag which might also be useful for local development.

PR #287, opened 2020-03-16
Enable Bazel GCS remote cache

lgeiger:
## What do these changes do?

This enables Bazel remote caching in order to speed up CI for MLIR and End2End test.

It remains to be seen what the network egress cost of using this cache will be and if the network latency between GCS and GitHub actions is low enough to make a significant improvement. Since Ingress is free on GCS I didn't prevent writing to the cache from branches other than master. If cache integrity becomes a problem, we should investigate having a readonly cache for builds not from master (see https://docs.bazel.build/versions/master/remote-caching.html#read-from-and-write-to-the-remote-cache).

## CI Benchmarks

build job | master | first cache warmup | cached build
--- | --- | --- | ---
MLIR | 11:04 | DNF | 4:36
End2End | 21:52 | 30:12 |  6:16

The first uncached run takes quite a bit longer since it needs to upload ~3GB of cached objects, but subsequent build are super fast 🚀 .
We should monitor our network egress cost, since I don't know how much traffic the caching takes up. The cache is configured to delete objects after 14 days and is located in US East.

## Related issue number
#86

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -63,6 +63,10 @@ jobs:
     if: "!contains(github.event.head_commit.message, 'ci-skip')"
     steps:
       - uses: actions/checkout@v2
+      - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
+        with:
+          service_account_key: ${{ secrets.gcs_bazel_cache }}
+          export_default_credentials: true
       - name: Install Bazelisk
         run: |
           curl -L https://github.com/bazelbuild/bazelisk/releases/download/v1.3.0/bazelisk-darwin-amd64 > bazelisk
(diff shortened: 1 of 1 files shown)


lgeiger:
> Is it possible to see what build artifacts will be cached and for which build configurations?

The build artifacts on GCS are not human readable. Bazel cache correctly handles different build configs so they won't share artifacts if the config changes. Since we only use this cache on CI this is not a problem since the build env won't change there.

AdamHillier:
I know I'm the only one who currently uses a fork of compute engine, but presumably this means I'll no longer be able to run CI in the fork because the gcloud stuff will fail without the right credentials?

lgeiger:
> I know I'm the only one who currently uses a fork of compute engine, but presumably this means I'll no longer be able to run CI in the fork because the gcloud stuff will fail without the right credentials?

At least locally it will fall back to built without the cache if it can't find the bucket so I think it will be fine.

AdamHillier:
> At least locally it will fall back to built without the cache if it can't find the bucket so I think it will be fine.

Running `bazelisk run larq_compute_engine/tests:end2end_test --distinct_host_configuration=false --remote_http_cache=https://storage.googleapis.com/plumerai-bazel-cache/lce-macos --google_default_credentials` locally throws an error at me but this is a very minor issue that only affects me so let's merge this and I can fix it later for forks if necessary.

PR #288, opened 2020-03-16
Use `fastbuild` instead of `dbg` for tests.

AdamHillier:
In #271 I set the compilation mode to `dbg` so that the compiler flag `-DNDEBUG` is not set, and debug assertions don't get stripped.

According to the [Bazel documentation](https://docs.bazel.build/versions/master/user-manual.html#flag--compilation_mode), the compilation mode `fastbuild` also does not set `-DNDEBUG`, but (as the name suggests) should be faster.

I see a 15-25% speed-up when building the tests (specifically the TFLite CC tests, on MacOS) with `fastbuild` rather than `dbg`. I have also verified that adding the statement `RUY_DCHECK_EQ(true, false)` still causes the ARM and native tests to fail (which was the original issue addressed by #271).

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -124,7 +124,7 @@ build:nohdfs --define=no_hdfs_support=true
 build:nonccl --define=no_nccl_support=true
 
 build --config=v2 --config=noaws --config=nogcp --config=nohdfs --config=nonccl
-test --config=v2 --compilation_mode dbg
+test --config=v2 --compilation_mode=fastbuild
 
 # Android configs. Bazel needs to have --cpu and --fat_apk_cpu both set to the
 # target CPU to build transient dependencies correctly. See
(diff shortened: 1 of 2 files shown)


configure.sh
```diff
@@ -124,7 +124,7 @@ build:nohdfs --define=no_hdfs_support=true
 build:nonccl --define=no_nccl_support=true
 
 build --config=v2 --config=noaws --config=nogcp --config=nohdfs --config=nonccl
-test --config=v2 --compilation_mode dbg
+test --config=v2 --compilation_mode=fastbuild
```
lgeiger: I tried to enable this in the past for the MLIR converter tests, but it failed or segfaulted if I remember correctly. However we change quite a bit in the testing setup since then, so if CI is happy this looks great.

AdamHillier: Yeah it seems to work fine, although doesn't look like it makes the MLIR builds much (if at all) quicker. Definitely helps the other test builds though.

PR #289, opened 2020-03-16
Consolidate TFLite and core CC tests to improve test parallelization

lgeiger:
## What do these changes do?
This adds new targets that allow to run core and TFLite tests in on command which is handy for running tests locally and should improve test parallelization to slightly speed up CI

## How Has This Been Tested?
CI

## Benchmark Results
build job | master | PR
--- | --- | ---
TFLite | 5:22 | 4:32
ARM32 | 7:34 | 7:35
Aarch64 | 6:09 | 6:00


lgeiger:
For the TFLite build this helps, but for ARM it doesn't improve speed.
@larq/compute-engine I don't have a strong opinion on this, but I'd be happy to revert the ARM build changes since it doesn't translate to a speedup.

lgeiger:
> @lgeiger Maybe this could help for MLIR as well, by combining the python test, the end2end test and the mlir FileCheck tests?

For the Python tests this won't help since we don't compile anything there.

We could definitely try it out for the end2end and MLIR tests though which would save network traffic from the cache and might reduce some overhead.

Issue #279, opened 2020-03-10
Automate generating cartesian products in the bconv unit tests


Tombana:
The test parameters at the bottom of [bconv2d_test.cc](https://github.com/larq/compute-engine/blob/master/larq_compute_engine/tflite/tests/bconv2d_test.cc) should automatically be generated by something that generates cartesian products (like a nested `::testing::Combine`).

lgeiger:
BConv tests have been simplified a lot in #467 so I think we can close this.

PR #292, opened 2020-03-18
Remove empty unittest parameter combination

Tombana:
## What do these changes do?

The tflite tests have
```cpp
if (padding == Padding_SAME && activation == ActivationFunctionType_RELU) { return; }
```
This PR simply removes that particular combination from the test parameter list so that we don't have these "empty" tests that always succeed.

This is achieved by combining the padding and activation parameters into one tuple, so that `SAME` + `RELU` can be removed from the list.


Tombana:
> @Tombana I think the bconv tests are becoming really hard to track due to manually setting the different combinations of testing arguments. We need to find a way to fix this. I am sure by looking at TF c++ tests we can find some inspirations.

I agree that we need something cleaner (issue #279). However, for this particular PR the point is that padding and activation can *not* be generated by a cartesian product because we explicitly want to remove certain combinations. So even if we cleanup the testing parameters, then padding and activation should stay as a 'combined' parameter.

arashb:
> > @Tombana I think the bconv tests are becoming really hard to track due to manually setting the different combinations of testing arguments. We need to find a way to fix this. I am sure by looking at TF c++ tests we can find some inspirations.
> 
> I agree that we need something cleaner (issue #279). However, for this particular PR the point is that padding and activation can _not_ be generated by a cartesian product because we explicitly want to remove certain combinations. So even if we cleanup the testing parameters, then padding and activation should stay as a 'combined' parameter.

I see your point but then the solution wouldn't be manually removing that combination from the test matrix but rather adding the appropriate test criteria to the test cases. In this particular, case it means checking int the test for the combination of the padding, activation and expecting the error, ... from the software. (if you think `EXPECT_DEATH` is too slow you can look for other alternatives instead of changing the test matrix)

Tombana:
Closed because we have `GTEST_SKIP()` now (will soon be merged).

PR #293, opened 2020-03-19
update the benchmark results

arashb:
## What do these changes do?

update the benchmark table and add links to new QN versions.


Tombana:
I think this should be done on PR #240

arashb:
> I think this should be done on PR #240

I will update the benchmark results here. later on we can rebase on top of whichever PR merge into master earlier.

PR #298, opened 2020-03-21
Consolidate MLIR and End2End tests

lgeiger:
## What do these changes do?
This consolidates the MLIR and End2End tests. This should reduce the amount of network traffic and general overhead on CI since we now have proper caching, so splitting the tests doesn't improve speed anymore.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -71,30 +71,9 @@ jobs:
       - name: Configure Bazel
         run: ./configure.sh
       - name: Install pip dependencies
-        run: pip install numpy six --no-cache-dir
+        run: pip install tensorflow==2.1.0 larq~=0.9.1 larq_zoo==1.0.b3 pytest tensorflow_datasets==1.3.2 --no-cache-dir
       - name: Run FileCheck tests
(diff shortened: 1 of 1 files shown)


arashb:
> ## What do these changes do?
> This consolidates the MLIR and End2End tests. This should reduce the amount of network traffic and general overhead on CI since we now have proper caching, so splitting the tests doesn't improve speed anymore.

the reason we split was not necessary bc of the CI test but rather bc MLIR build is different entity than and end-to-end integration test and its logical to have them separate independent of CI speed.

lgeiger:
> the reason we split was not necessary bc of the CI test but rather bc MLIR build is different entity than and end-to-end integration test and its logical to have them separate independent of CI speed. 

The tests are still split into different entities, the only difference is that they now run in sequence on the same VM instead of running in parallel on different VMs.
The main reason to do this is not to increase build speed, but it should cut our network cost in half since both tests have to build a lot of the same code and we pay for every GB we read from the cache. I don't have an overview of much our our caching costs yet  (I don't think it will be a lot), but since the disk size of the cache is ~3GB I think it is good to keep the number of avoidable cache reads to a minimum.

PR #299, opened 2020-03-23
Fix docs link in contributing guide

lgeiger:


lgeiger:
The test will pass when rebased onto #301

Issue #290, opened 2020-03-17
Benchmark when "bitpack after im2col" is faster


Tombana:
In `tflite/kernels/bconv2d.cc` in `decide_bitpack_before_im2col` the kernel decides if it should do bitpacking before or after im2col.

Currently this is simply `if (channels_in >= bitwidth / 4)`, based on old benchmarks. A simple benchmark shows that this is no longer the correct decision boundary, it should be closer to `bitwidth/2` judging from this plot:
![plot_latency](https://user-images.githubusercontent.com/248512/76899362-a244f500-6897-11ea-8503-b5270e959698.png)

> Plot explanation: This shows inference times for our bconv2d op on an input size `(224,224)` on the Pixel 1. In the case "bitpack before im2col" then basically it doesn't matter if you have 1 or 64 input channels because they all get bitpacked into one `uint64`. Therefore that line is flat. For 32-bit there is the clear jump of going from one `uint32` to two `uint32`s but otherwise that is flat as well.

Furthermore, the proper decision probably depends on the "depth" dimension of the matrix multiplication, which has size `filter_height * filter_width * channels_in`.

We should
- benchmark if "depth" is a better indicator than `channels_in`
- benchmark if other parts of input/filter shape are relevant in this decision.

If needed (plot suggests it wont be) the kernel can also decide to use 32-bit bitpacking vs 64-bit bitpacking dynamically based on input/filter shapes.

lgeiger:
Since #457 we will always bitpack before im2col, so this issue can be closed.

PR #302, opened 2020-03-24
Support a dynamic bitpacking bitwidth.

AdamHillier:
## What do these changes do?
This PR removes the two op registrations (one 32 bit, one 64 bit) and replaces them with a single generic op registration, with a `bitpacking_bitwidth` parameter in the `conv_params` that is set in the op `Init` (based on if we're compiling a 32-bit kernel or a 64-bit kernel). Functionally, this does not change the current behaviour, but it lays the groundwork to select a bitwidth more intelligently (e.g. based on benchmarking as in #290).

These changes are required for the upcoming feature of reading/writing bitpacked activations, because that is being implemented with byte (8-bit) bitpacking.

## How Has This Been Tested?
The `bconv2d_test.cc` test (run on x86 with the C++ kernel) currently fails because it relies on being able to directly register a 32 bit op or a 64 bit op, and this PR removes separate registrations.

To me, this testing behaviour doesn't actually make much sense, as we compile a 64-bit op and test it with 32-bit bitpacking, which is not a scenario that is supported. I think it would make more sense to test 32-bit bitpacking by compiling and 32-bit kernels directly.

Thoughts @larq/compute-engine?


AdamHillier:
(Apologies, I intended to make this a draft PR.)

Tombana:
How about 3 registrations: 32-bit, 64-bit and 'dynamic' . This can be achieved by having the template parameter `int bitwidth` in `Prepare`, which for the 'dynamic' registration will be 0. Then in Prepare, you can do `if (bitwidth != 0){ conv_params->bitwidth = bitwidth; } else { /* something smart */ }`.
This allows us to keep easily testing the 32 and 64 bit implementations, as well as supporting reading bitpacked input in the future.

AdamHillier:
> How about 3 registrations: 32-bit, 64-bit and 'dynamic' . This can be achieved by having the template parameter `int bitwidth` in `Prepare`, which for the 'dynamic' registration will be 0. Then in Prepare, you can do `if (bitwidth != 0){ conv_params->bitwidth = bitwidth; } else { /* something smart */ }`.

I haven't added the dynamic op yet because there is no dynamic choice happening yet, but it will be trivial to do so in the future.

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -510,38 +518,40 @@ void EvalOpt(TfLiteContext* context, TfLiteNode* node,
...
-      bconv2d::Init, bconv2d::Free,
-      bconv2d::Prepare<bconv2d::KernelType::kRuyOptimized, 64>,
-      bconv2d::Eval<bconv2d::KernelType::kRuyOptimized, std::uint64_t>};
+      bconv2d::Init<64>, bconv2d::Free,
+      bconv2d::Prepare<bconv2d::KernelType::kRuyOptimized>,
+      bconv2d::Eval<bconv2d::KernelType::kRuyOptimized>};
   return &r;
 }
 
-// Use this registration wrapper to decide which impl. to use.
```
arashb: this comment needs to be back 😀

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -358,9 +366,9 @@ TfLiteStatus Prepare(KernelType kernel_type, const int bitwidth,
   return kTfLiteOk;
 }
 
-template <KernelType kernel_type, int bitwidth>
+template <KernelType kernel_type, const int default_bitpacking_bitwidth>
```
arashb: `const` is not required here. the template arguments are passed at compile-time.

Tombana: I've never seen `const` in a template argument. Does it actually do anything?
```suggestion
template <KernelType kernel_type, int default_bitpacking_bitwidth>
```

AdamHillier: It seems to prevent assigning to the value in the body of the function - I tend to add const to everything reflexively and just remove it when I need to, because I'm too used to JavaScript and Scala and Rust :joy: I have removed it though now.

PR #303, opened 2020-03-24
Fix typos and standardise std:: namespace usage.

AdamHillier:
This PR doesn't change any behaviour. It fixes a couple of typos in comments, and standardises the use of `std::` before integer types.

I don't feel strongly at all about using `std::uint8_t` versus just `uint8_t` (or `std::size_t` vs `size_t`, et cetera), but in the LCE codebase we're using both at the moment, and I think it's much cleaner to standardise the usage one way or the other.

I've gone with the `std::` prefixed versions because my very quick analysis indicates that that usage is about 50% more prevalent than the non-prefixed versions in the existing code. I'm equally happy to go the other way and use the non-prefixed versions everywhere.

### DIFF
--- a/larq_compute_engine/core/benchmark/aarch64/packbits_bench.cc
+++ b/larq_compute_engine/core/benchmark/aarch64/packbits_bench.cc
@@ -12,25 +12,25 @@
 
 using namespace compute_engine::core;
 
-constexpr size_t min_size = 128;
-constexpr size_t max_size = 1 << 16;  // 64K
+constexpr std::size_t min_size = 128;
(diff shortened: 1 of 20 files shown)


AdamHillier:
> LGTM. Thanks for doing this.
> 
> Side note: in the PR you mentioned that it also fixes some typos in comments, but I only see the `std::` and `ruy::` changes. Did I not look hard enough or did you forget to commit those changes?

Ah oops yes I forgot to commit them, thanks for noticing that :) 

Just a couple of `can not` -> `cannot` and one missing `TF_LITE_ASSERT` (that is present in the other branches of the if statement).

PR #305, opened 2020-03-25
Support reading and writing bitpacked activations in C++ kernels.

AdamHillier:
## What do these changes do?

This feature is disabled by default and so does not change any behaviour for new or existing converted models; enabling this feature will require changes to the converter to set the correct op flags.

Currently, binary convolutions expect full precision input, bitpack that input, perform the computation, and then write full-precision output. This feature allows a binary convolutions to instead perform bitpacking inside the kernel and write (8-bit or 32-bit) bitpacked output, and for the subsequent binary convolution to skip the usual initial bitpacking step and read the already-bitpacked input directly. Doing this significantly reduces the number of reads and writes to memory and the the overall memory footprint of the model inference.

Support is added only for the C++ kernels (x86 and Arm32), including the reference kernel. Support for the optimised assembly kernels (Arm64) requires additional future work.

## How have these changes been tested?

The existing `bconv2d_test.cc` tests have been extended to support testing bitpacked input/output tensors with the Ruy c++ kernel (8-bit bitpacking) and the reference kernel (32-bit bitpacking).

### DIFF
--- a/larq_compute_engine/core/bgemm_kernels_ruy.h
+++ b/larq_compute_engine/core/bgemm_kernels_ruy.h
@@ -88,7 +88,7 @@ struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, DstScalar,
   }
 };
 
-// A template specialisation for writing 8-bit bitpacked output.
+// A template specialisation for writing 32-bit bitpacked output.
 template <typename LhsScalar, typename RhsScalar, typename Spec>
 struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, std::int32_t,
                    Spec> {
@@ -112,9 +112,9 @@ struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, std::int32_t,
     // We are writing 32-bit bitpacked output (where we bitpack along the
     // channel axis) and so we need to operate on blocks of 32 channels at a
     // time. As the destination is column major, this means blocks of 32 rows at
-    // a time. The blocks Ruy uses are always a power of two and are almost
-    // always >> 8. However, when running with multiple threads and a very large
-    // input size, Ruy may use blocks of 4 rows.
+    // a time. The blocks Ruy uses are always a power of two and are usually
+    // greater than 32. However, when running with multiple threads and a very
+    // large input size, Ruy may use blocks of fewer rows.
     //     In this scenario, we round the start and end row down and up to the
     // nearest multiple of 32 respectively. This is a thread-safe way to ensure
     // that the result is correct, at the cost of some rare repeated
--- a/larq_compute_engine/core/bgemm_kernels_ruy.h
+++ b/larq_compute_engine/core/bgemm_kernels_ruy.h
@@ -90,7 +90,7 @@ struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, DstScalar,
 
 // A template specialisation for writing 8-bit bitpacked output.
 template <typename LhsScalar, typename RhsScalar, typename Spec>
-struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, std::uint8_t,
+struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, std::int8_t,
                    Spec> {
   using AccumScalar = typename Spec::AccumScalar;
   using LhsLayout = typename Spec::StandardCppKernelLhsLayout;
@@ -99,7 +99,7 @@ struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, std::uint8_t,
   void Run(const ruy::PackedMatrix<LhsScalar>& lhs,
            const ruy::PackedMatrix<RhsScalar>& rhs, const Spec& spec,
            int start_row, int start_col, int end_row, int end_col,
-           ruy::Matrix<std::uint8_t>* dst) const {
+           ruy::Matrix<std::int8_t>* dst) const {
     static_assert(std::is_same<LhsScalar, RhsScalar>::value,
                   "Inputs to binary kernel should have the same type.");
     static_assert(
--- a/larq_compute_engine/core/bgemm_kernels_ruy.h
+++ b/larq_compute_engine/core/bgemm_kernels_ruy.h
@@ -90,7 +90,7 @@ struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, DstScalar,
 
 // A template specialisation for writing 8-bit bitpacked output.
 template <typename LhsScalar, typename RhsScalar, typename Spec>
-struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, std::int8_t,
+struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, std::int32_t,
                    Spec> {
   using AccumScalar = typename Spec::AccumScalar;
   using LhsLayout = typename Spec::StandardCppKernelLhsLayout;
@@ -99,7 +99,7 @@ struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, std::int8_t,
   void Run(const ruy::PackedMatrix<LhsScalar>& lhs,
            const ruy::PackedMatrix<RhsScalar>& rhs, const Spec& spec,
            int start_row, int start_col, int end_row, int end_col,
-           ruy::Matrix<std::int8_t>* dst) const {
+           ruy::Matrix<std::int32_t>* dst) const {
     static_assert(std::is_same<LhsScalar, RhsScalar>::value,
                   "Inputs to binary kernel should have the same type.");
     static_assert(
@@ -109,18 +109,21 @@ struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, std::int8_t,
 
     using TBitpacked = LhsScalar;
 
-    // We are writing 8-bit bitpacked output (where we bitpack along the channel
-    // axis) and so we need to operate on blocks of 8 channels at a time. As the
-    // destination is column major, this means blocks of 8 rows at a time. The
-    // blocks Ruy uses are always a power of two and are almost always >> 8.
-    // However, when running with multiple threads and a very large input size,
-    // Ruy may use blocks of 4 rows.
+    // We are writing 32-bit bitpacked output (where we bitpack along the
+    // channel axis) and so we need to operate on blocks of 32 channels at a
+    // time. As the destination is column major, this means blocks of 32 rows at
+    // a time. The blocks Ruy uses are always a power of two and are almost
+    // always >> 8. However, when running with multiple threads and a very large
+    // input size, Ruy may use blocks of 4 rows.
     //     In this scenario, we round the start and end row down and up to the
-    // nearest multiple of 8 respectively. This is a thread-safe way to ensure
+    // nearest multiple of 32 respectively. This is a thread-safe way to ensure
     // that the result is correct, at the cost of some rare repeated
     // computation, which is acceptable for this non-optimised kernel.
-    start_row = 8 * (start_row / 8);
-    end_row = 8 * ((end_row + 7) / 8);
+    //
+    // Note that the rows in all these calculations are the *unpacked* rows.
+
+    start_row = 32 * (start_row / 32);
+    end_row = 32 * ((end_row + 31) / 32);
 
     int clamped_end_row = std::min(end_row, dst->layout.rows);
     int clamped_end_col = std::min(end_col, dst->layout.cols);
@@ -140,12 +143,12 @@ struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, std::int8_t,
         "Binary Kernel (Standard Cpp) Bitpacked Output.");
 
     const int depth = lhs.layout.rows;
-    const int dst_stride_bitpacked = (dst->layout.stride + 7) / 8;
+    const int dst_stride_bitpacked = (dst->layout.stride + 31) / 32;
 
     // The destination is column major and we need to bitpack along the row
     // (channels) axis so we need to loop over column index then row index.
     for (int j = start_col; j < clamped_end_col; j++) {
-      std::uint8_t bitpacked_column = 0;
+      std::uint32_t bitpacked_column = 0;
       for (int i = start_row; i < clamped_end_row; i++) {
         using AccumScalar = typename Spec::AccumScalar;
         AccumScalar accum = 0;
@@ -169,10 +172,10 @@ struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, std::int8_t,
           dst_val += spec.post_activation_bias[i];
         }
         if (dst_val < 0) {
-          bitpacked_column += 1 << ((i - start_row) % 8);
+          bitpacked_column |= 1ULL << ((i - start_row) % 32);
         }
-        if (((i - start_row + 1) % 8 == 0) || (i + 1 == clamped_end_row)) {
-          *(dst->data.get() + i / 8 + j * dst_stride_bitpacked) =
+        if (((i - start_row + 1) % 32 == 0) || (i + 1 == clamped_end_row)) {
+          *(dst->data.get() + i / 32 + j * dst_stride_bitpacked) =
               bitpacked_column;
           bitpacked_column = 0;
         }
--- a/larq_compute_engine/core/bgemm_kernels_ruy.h
+++ b/larq_compute_engine/core/bgemm_kernels_ruy.h
@@ -88,6 +88,99 @@ struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, DstScalar,
   }
 };
 
+// A template specialisation for writing 8-bit bitpacked output.
+template <typename LhsScalar, typename RhsScalar, typename Spec>
+struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, std::uint8_t,
+                   Spec> {
+  using AccumScalar = typename Spec::AccumScalar;
+  using LhsLayout = typename Spec::StandardCppKernelLhsLayout;
+  using RhsLayout = typename Spec::StandardCppKernelRhsLayout;
+  explicit BgemmKernel(ruy::Tuning) {}
+  void Run(const ruy::PackedMatrix<LhsScalar>& lhs,
+           const ruy::PackedMatrix<RhsScalar>& rhs, const Spec& spec,
+           int start_row, int start_col, int end_row, int end_col,
+           ruy::Matrix<std::uint8_t>* dst) const {
+    static_assert(std::is_same<LhsScalar, RhsScalar>::value,
+                  "Inputs to binary kernel should have the same type.");
+    static_assert(
+        std::is_unsigned<LhsScalar>::value &&
+            std::is_integral<LhsScalar>::value,
+        "Input to binary kernel should be of type unsigned integral.");
+
+    using TBitpacked = LhsScalar;
+
+    // We are writing 8-bit bitpacked output (where we bitpack along the channel
+    // axis) and so we need to operate on blocks of 8 channels at a time. As the
+    // destination is column major, this means blocks of 8 rows at a time. The
+    // blocks Ruy uses are always a power of two and are almost always >> 8.
+    // However, when running with multiple threads and a very large input size,
+    // Ruy may use blocks of 4 rows.
+    //     In this scenario, we round the start and end row down and up to the
+    // nearest multiple of 8 respectively. This is a thread-safe way to ensure
+    // that the result is correct, at the cost of some rare repeated
+    // computation, which is acceptable for this non-optimised kernel.
+    start_row = 8 * (start_row / 8);
+    end_row = 8 * ((end_row + 7) / 8);
+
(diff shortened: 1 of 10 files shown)


larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -37,30 +37,30 @@ const int kTensorNotAllocated = -1;
...
   // padding
   TfLitePadding padding_type{};
   TfLitePaddingValues padding_values{};
   int pad_value = 0;  // Must be 0 or 1
 
   // output tensor dimensions
-  std::int64_t out_width{0};
-  std::int64_t out_height{0};
+  std::int32_t out_width{0};
+  std::int32_t out_height{0};
```
AdamHillier: I now can't remember why I had to change these to int32 rather than int64 (it was to fix some sort of compilation error), but I can't see any issue with doing so? I don't think we need a double-word to store input/filter/output dimensions.

larq_compute_engine/tflite/tests/BUILD
```diff
@@ -5,7 +5,7 @@ package(
 
 cc_test(
     name = "bconv2d_test",
-    size = "small",
+    size = "medium",
```
AdamHillier: These tests now take quite a while to run (~150 seconds on my 8 core machine) so I had to change this to avoid them timing out.

larq_compute_engine/tflite/kernels/bconv2d.cc
Tombana: I think the `int64` was a remnant of some flatbuffer parsing logic where it seemed like a flatbuffer `int8`/`int32`/`int64` could alwasy be read by a `.GetAsInt64()` but not the other way around, so we initially went with int64 to be sure that it worked. Now that we know how to change the flatbuffer types (using the mlir IR) we can go back to int32 (or even int16 since image sizes will probably never get that big).
I think at some point in the future we want to follow the tflite implementation, where we also change the `int64 strides[4]` and `int64 dilations[4]` to `int32 stride_height, stride_width, dilation_height, dilation_width`, i.e. remove the useless batch and channel strides. When we do that we can also change all the int64 to int32. For now we can leave it.

AdamHillier: Makes sense, and apparently it works fine with `uint64` so I'm not sure what the previous problem I had was. Have reverted in 7ff1f0b.

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -403,7 +460,7 @@ inline void GetConvParamsType(const TfLiteBConv2DParams& conv_params,
   op_params.quantized_activation_max = conv_params.output_activation_max;
 }
 
-template <class T, class TBitpacked>
+template <typename T, typename TBitpacked, typename DstScalar>
 void EvalOpt(TfLiteContext* context, TfLiteNode* node,
              TfLiteBConv2DParams* params) {
```
Tombana, line 465: Right now it might be unclear what `T` refers to.
How about `<typename TIn, typename TBitpacked, typename TOut>` ?  Or `<typename SrcScalar, typename TBitpacked, typename DstScalar>` .

larq_compute_engine/tflite/kernels/bconv2d_impl.h
```diff
@@ -65,21 +65,24 @@ inline void im2col(const ConvParams& params, const RuntimeShape& input_shape,
 // The inputs post_mutiply and post_activation_bias are currently float
 // in order to accomodate for batchnorm scales
 // Later this might be changed to the int8 system of multipliers+shifts
-template <class T, class TBitpacked>
+template <typename T, typename TBitpacked, typename AccumScalar,
+          typename DstScalar>
 inline void BConv2D(const ConvParams& params, const RuntimeShape& input_shape,
```
Tombana, line 70: Same as above, I'd rather rename `T` to `SrcScalar` or `TIn` or so, now that input and output type are no longer the same.

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -525,12 +587,45 @@ TfLiteStatus Eval(TfLiteContext* context, TfLiteNode* node) {
 
   if (kernel_type == KernelType::kRuyOptimized) {
     switch (conv_params->bitpacking_bitwidth) {
+// On Arm64, we don't yet support 8-bit bitpacked input or writing bitpacked
+// output.
+#if RUY_PLATFORM(ARM_64)
```
Tombana, line 592: I think on Arm64 we should let it fall-back to the C++ kernel. Then you can remove this entire `#if RUY_PLATFORM(ARM_64)` block.

I would suggest adding a line near here:
https://github.com/larq/compute-engine/blob/8e8da88a3ce997da4705f3713f282c9bde1154f0/larq_compute_engine/core/bgemm_impl_ruy.h#L74
with something like

```cpp
// For writing bitpacked output, fallback to the standard C++ kernel.
if (std::is_same<DstScalar, std::uint8_t>::value) {
  bgemm_runtime_path = ruy::Path::kStandardCpp;
}
```

AdamHillier: Nice, that's a good idea

larq_compute_engine/tflite/kernels/bconv2d_impl.h
```diff
@@ -113,29 +116,45 @@ inline void BConv2D(const ConvParams& params, const RuntimeShape& input_shape,
...
+      // dimension [batch, input height, input width, channels].
+      RuntimeShape packed_input_shape;
+
+      packbits_tensor(input_shape, input_data, packed_input_shape,
+                      input_data_bp);
+
+      im2col<TBitpacked>(params, packed_input_shape, input_data_bp.data(),
+                         packed_filter_shape, output_shape, im2col_shape,
+                         packed_im2col_data, result_shape, &rhs_data);
+    }
```
Tombana, line 151: I don't feel too strongly about this, but what do you think of this:

```suggestion
    RuntimeShape packed_input_shape;
    TBitpacked* input_data_bp;
    if (read_bitpacked_input) {
      packed_input_shape = input_shape;
      input_data_bp =
          reinterpret_cast<const TBitpacked*>(input_data);
    } else {
      // Buffer for bitpacked input data.
      static std::vector<TBitpacked> input_data_bp_buffer;
      // The input tensor has this shape which we bitpack along the channels
      // dimension [batch, input height, input width, channels].
      packbits_tensor(input_shape, input_data, packed_input_shape,
                      input_data_bp_buffer);
      input_data_bp = input_data_bp_buffer.data();
    }
    im2col<TBitpacked>(params, packed_input_shape, input_data_bp,
                       packed_filter_shape, output_shape, im2col_shape,
                       packed_im2col_data, result_shape, &rhs_data);
```

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -15,6 +15,7 @@
    ==============================================================================*/
 
 #include <gtest/gtest.h>
+#include <stdint.h>
```
Tombana, line 18: For C headers `xxx.h` there's usually a C++ version `cxxx` (without the `.h`), see [this page](https://en.cppreference.com/w/cpp/header#C_compatibility_headers), which puts standard library stuff in the `std` namespace.
```suggestion
#include <cstdint>
```

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -586,21 +729,48 @@ INSTANTIATE_TEST_SUITE_P(
...
                           ActivationFunctionType_RELU),  // activation function
-        ::testing::Values(1, 2),                         // number of threads
+#if RUY_PLATFORM(ARM_64)
+        // There's not yet Arm64 kernel support for bitpacked activations.
+        ::testing::Values(false),  // read bitpacked input
+        ::testing::Values(false),  // write bitpcked output
+#else
+        ::testing::Values(false, true),  // read bitpacked input
+        ::testing::Values(false, true),  // write bitpacked output
+#endif
```
Tombana, line 739: Just a reminder, when the arm-64 case falls back to the standard cpp kernel as per another comment, we can remove this `#if` block here.

larq_compute_engine/tflite/kernels/bconv2d_impl.h
AdamHillier: This didn't quite work verbatim, but I've done a variant of your suggestion in 51b14d1.

larq_compute_engine/tflite/kernels/bconv2d_impl.h
AdamHillier: Done in 21aac16.

larq_compute_engine/tflite/kernels/bconv2d.cc
AdamHillier: Went with `SrcScalar`, in 21aac16.

larq_compute_engine/tflite/kernels/bconv2d_impl.h
AdamHillier: I thought this suggestion looked a lot nicer, but it broke things so I've reverted this change :( Can you see anything obviously wrong with 51b14d1?

larq_compute_engine/tflite/tests/bconv2d_test.cc
AdamHillier: Done in a8d1c1d.

larq_compute_engine/tflite/kernels/bconv2d.cc
AdamHillier: I do think this is a good idea but a first pass at it spews a lot of compiler errors (adding the three lines to `bgemm_impl_ruy.h` works fine, but removing the `#if RUY_PLATFORM(ARM_64)` breaks).

AdamHillier:
Thank you very much for reviewing @Tombana. This PR is now extended to work with the reference kernel as well (which took far, far, far longer than I expected it would).

A few remarks:

- In the reference kernel 32-bit bitpacking is used; in the C++ Ruy kernel 8-bit bitpacking is used. This now means that reference and optimised ops cannot be used simultaneously in the same interpreter, because they won't interact correctly when using bitpacked activations. This shouldn't ever actually be a problem because mixing the two wouldn't make sense, but it's not a restriction that we actually enforce anywhere and technically did work previously.

- One problem I encountered with 32-bit bitpacking in the reference kernel was being forced to use the `kTFLiteInt32` type for bitpacked tensors, as TFL doesn't have an unsigned 32 bit integer type. Unfortunately, this conflicted with using `std::uint8_t` and `std::uint32_t` as destination scalars throughout my implementation to template the bitpacked versions of the kernels. TFLite didn't like me populating a `kTFLiteInt32` tensor with `std::uint32_t` data, or extracting `std::uint32_t` data from the tensor. Rather than searching for a cumbersome workaround, I elected to change all `std::uint8_t` and `std::uint32_t` usage as destination scalars to the signed equivalents `std::uint8` and `std::uint32`.

    This means there is now a bit of a mis-match between `TBitpacked` and `DstScalar`, because in the former we still used the unsigned types (although I don't think there's a technical restriction given that the bitpacking functions use `std::make_unsigned`, and so would also work with the signed scalars).

    How do you feel about switching to the signed types everywhere?

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -383,6 +408,134 @@ MATCHER_P(FloatNearPointwise, tol, "Out of range") {
...
+  EXPECT_EQ(lce_output_data, builtin_output_data_bp);
+}
+
+// A specialisation for float output (i.e., not writing bitpacked output).
+inline void test_lce_op_output(const std::vector<float>& lce_output_data,
+                               const std::vector<int>& builtin_output_shape,
+                               const std::vector<float>& builtin_output_data) {
+  EXPECT_THAT(lce_output_data, ::testing::Pointwise(FloatNearPointwise(1e-4),
+                                                    builtin_output_data));
+}
```
AdamHillier, line 457: I was using `if constexpr` (C++17 feature) to obviate the need for these templated util functions (used in `test_lce_op` just below). This worked fine on x86 but unfortunately not on ARM, even when changing the relevant compiler flags to `-std=c++17`, so we're stuck with this ugliness.

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -569,8 +714,82 @@ TEST_P(BConv2DOpTest, SimpleTest) {
...
+            m_builtin.GetOutputShape(), builtin_output);
+      }
+    } else {  // Not reading bitpacked input.
+      test_lce_op<T, T>(
+          registration, input_shape, packed_filter_shape, input_depth,
+          stride_width, stride_height, bconv_padding, pad_values,
+          dilation_width_factor, dilation_height_factor, activation,
+          num_threads, input_data, packed_filters_data,
+          post_activation_multiplier_data, post_activation_bias_data,
+          m_builtin.GetOutputShape(), builtin_output);
```
AdamHillier, line 790: This sequence of seven calls to `test_lce_op`, varying only in type parameters, is not ideal but I don't see a much better option.

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -167,15 +171,42 @@ void* Init(TfLiteContext* context, const char* buffer, std::size_t length) {
...
+    // We need to gatekeep with a null check in case the model was converted
+    // before these flags were added to the converter.
+    conv_params->read_bitpacked_input = m["read_bitpacked_input"].AsBool();
+    conv_params->write_bitpacked_output = m["write_bitpacked_output"].AsBool();
+  } else {
+    conv_params->read_bitpacked_input = false;
+    conv_params->write_bitpacked_output = false;
+  }
+
+  // If we are reading bitpacked input then the both input tensor and the
```
AdamHillier, line 193:
```suggestion
  // If we are reading bitpacked input then both the input tensor and the
```

Tombana:
> * In the reference kernel 32-bit bitpacking is used; in the C++ Ruy kernel 8-bit bitpacking is used. This now means that reference and optimised ops cannot be used simultaneously in the same interpreter, because they won't interact correctly when using bitpacked activations. This shouldn't ever actually be a problem because mixing the two wouldn't make sense, but it's not a restriction that we actually enforce anywhere and technically did work previously.

It should indeed not be a problem. In the long-term however, I think we have to try to get the 32- (or even 64-)bit RUY version working (it already does, but simply not multithreaded), because it will be faster because of memory-alignment. I think we can do that in a separate PR, later.

> * One problem I encountered with 32-bit bitpacking in the reference kernel was being forced to use the `kTFLiteInt32` type for bitpacked tensors, as TFL doesn't have an unsigned 32 bit integer type. Unfortunately, this conflicted with using `std::uint8_t` and `std::uint32_t` as destination scalars throughout my implementation to template the bitpacked versions of the kernels. TFLite didn't like me populating a `kTFLiteInt32` tensor with `std::uint32_t` data, or extracting `std::uint32_t` data from the tensor. Rather than searching for a cumbersome workaround, I elected to change all `std::uint8_t` and `std::uint32_t` usage as destination scalars to the signed equivalents `std::uint8` and `std::uint32`.
>   This means there is now a bit of a mis-match between `TBitpacked` and `DstScalar`, because in the former we still used the unsigned types (although I don't think there's a technical restriction given that the bitpacking functions use `std::make_unsigned`, and so would also work with the signed scalars).
>
>   How do you feel about switching to the signed bitpacking types everywhere?

I've ran into the same problems, its very annoying that you can't put `uint32` data into a tflite tensor without reinterpret-casting it to int32.
Given that one could forget about `std::make_unsigned` (which can cause hard-to-debug problems), I vote for an entirely new datatype, now that we are getting int8 quantized types as well. The RFC is here: #325
This does not solve the issue of the tflite tensor type enum, but at least our code will be much clearer and readable.

> * It will probably be easiest to review the new changes commit-by-commit.

Thanks!

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -383,6 +408,134 @@ MATCHER_P(FloatNearPointwise, tol, "Out of range") {
...
+  std::size_t output_rows_bp, output_cols_bp, output_bitpadding;
+  core::packbits_matrix<core::BitpackOrder::Canonical>(
+      builtin_output_data.data(),
+      builtin_output_shape.at(0) * builtin_output_shape.at(1) *
+          builtin_output_shape.at(2),
+      builtin_output_shape.at(3), builtin_output_data_bp, output_rows_bp,
+      output_cols_bp, output_bitpadding, core::Axis::RowWise);
+
+  // We need the outputs here to be bit-exact, so don't allow for floating
+  // point imprecision.
```
Tombana, line 447: I'm wondering if this would fail if the float in the builtin op was `-0.00001f` and the float in our op was `0.00001f`. I think its fine for now, but if this happens in the future we would have to add an extra loop here that manually checks the bits and allows for a mismatch in that case.

AdamHillier: Ah that is a good point. I guess the likelihood of that happening is quite slim but iterating though and checking individually would not be hard.

larq_compute_engine/tflite/kernels/bconv2d_impl.h
Tombana: I think in the non-bitpacked block, you don't set `const TBitpacked* input_data_bp` so its not passed to im2col. I'll also give it a try.

larq_compute_engine/core/bgemm_kernels_ruy.h
```diff
@@ -88,6 +88,102 @@ struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, DstScalar,
   }
 };
 
+// A template specialisation for writing 8-bit bitpacked output.
```
AdamHillier, line 91:
```suggestion
// A template specialisation for writing 32-bit bitpacked output.
```

larq_compute_engine/core/bgemm_kernels_ruy.h
```diff
@@ -88,6 +88,102 @@ struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, DstScalar,
...
+        "Input to binary kernel should be of type unsigned integral.");
+
+    using TBitpacked = LhsScalar;
+
+    // We are writing 32-bit bitpacked output (where we bitpack along the
+    // channel axis) and so we need to operate on blocks of 32 channels at a
+    // time. As the destination is column major, this means blocks of 32 rows at
+    // a time. The blocks Ruy uses are always a power of two and are almost
+    // always >> 8. However, when running with multiple threads and a very large
+    // input size, Ruy may use blocks of 4 rows.
```
AdamHillier, line 117:
```suggestion
    // a time. The blocks Ruy uses are always a power of two and are usually
    // greater than 32. However, when running with multiple threads and a very
    // large input size, Ruy may use blocks of fewer rows.
```

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -353,6 +408,9 @@ TfLiteStatus Prepare(KernelType kernel_type,
     // Determine the type
     if (conv_params->bitpack_before_im2col) {
       switch (conv_params->bitpacking_bitwidth) {
+        case 8:
+          im2col->type = kTfLiteInt8;
+          break;
```
AdamHillier, line 413:
```suggestion
```

Is the bitwidth ever 8 anymore?

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -584,16 +656,34 @@ TfLiteStatus Eval(TfLiteContext* context, TfLiteNode* node) {
 
   if (kernel_type == KernelType::kRuyOptimized) {
     switch (conv_params->bitpacking_bitwidth) {
+      case 8:
+        if (conv_params->write_bitpacked_output)
+          EvalOpt<float, std::uint8_t, std::int32_t>(context, node,
+                                                     conv_params);
+        else
+          EvalOpt<float, std::uint8_t, float>(context, node, conv_params);
+        return kTfLiteOk;
```
AdamHillier, line 665:
```suggestion
```

Potentially here also?

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -383,8 +395,67 @@ MATCHER_P(FloatNearPointwise, tol, "Out of range") {
...
+  EXPECT_THAT(lce_output_data, ::testing::Pointwise(FloatNearPointwise(1e-4),
+                                                    builtin_output_data));
+}
+
+template <typename TInput, typename TOutput>
+void runTest(const TestParam& param) {
+  static_assert(std::is_same<TInput, float>::value ||
+                    std::is_same<TInput, std::int8_t>::value ||
+                    std::is_same<TInput, std::int32_t>::value,
+                "The LCE op input type must be float or uint8 or uint32.");
```
AdamHillier, line 454:
```suggestion
                "The LCE op input type must be float or int8 or int32.");
```

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -383,8 +395,67 @@ MATCHER_P(FloatNearPointwise, tol, "Out of range") {
...
+  static_assert(std::is_same<TOutput, float>::value ||
+                    std::is_same<TOutput, std::int8_t>::value ||
+                    std::is_same<TOutput, std::int32_t>::value,
+                "The LCE op output type must be float or uint8 or uint32.");
```
AdamHillier, line 458:
```suggestion
                "The LCE op output type must be float or int8 or int32.");
```

AdamHillier:
GitHub won't let me approve my own PR, even if a substantial part of it was written by someone else, but thank you very much @Tombana for all of this :tada:

PR #306, opened 2020-03-25
(ci-skip) Fix comment typos.

AdamHillier:


AdamHillier:
@lgeiger I thought the `(ci-skip)` in the commit message would stop CI running?

lgeiger:
> I thought the (ci-skip) in the commit message would stop CI running?

Looks like this doesn't work anymore since we moved to run CI on PRs instead of on push :(

larq_compute_engine/core/bgemm_kernels_arm64.h
```diff
@@ -552,12 +552,12 @@ void CheckOffsetsInKernelParams64BP(const Params&) {
...
 // v4 -- v7 from RHS:
 //
-//                                      int32 RHS 2x4 block
+//                                    int64 RHS 2x4 block
 //                          /--------------------------------------\
 //                          |v4.d[0]         ...          v7.d[0]  |
 //                          |v4.d[1]         ...          v7.d[1]  |
 //                          \--------------------------------------/
-//    int32 LHS 4x2 block
+//     int64 LHS 4x2 block
```
AdamHillier: @arashb/@tombana I could be wrong about this but I think these are int64 blocks (here and line 555)?

AdamHillier:
Also it doesn't matter much but [line 135](https://github.com/larq/compute-engine/blob/f561da84157433b2cf23611a38e46827493190d9/larq_compute_engine/core/bgemm_kernels_arm64.h#L135) conflicts with the [diagram of the accumulator block](https://github.com/larq/compute-engine/blob/f561da84157433b2cf23611a38e46827493190d9/larq_compute_engine/core/bgemm_kernels_arm64.h#L147-L150).

larq_compute_engine/core/bgemm_kernels_arm64.h
arashb: you are right, it should be int64.

arashb:
> Also it doesn't matter much but [line 135](https://github.com/larq/compute-engine/blob/f561da84157433b2cf23611a38e46827493190d9/larq_compute_engine/core/bgemm_kernels_arm64.h#L135) conflicts with the [diagram of the accumulator block](https://github.com/larq/compute-engine/blob/f561da84157433b2cf23611a38e46827493190d9/larq_compute_engine/core/bgemm_kernels_arm64.h#L147-L150).
> 
> The latter doesn't look right to me regardless because it shows `v16 ... v22` which is 7 registers but I think it should be either 4 or 8?

The comment is wrong but the diagram is correct. the destination registers are the following 4 registers: v16, v18, v20, v22.

PR #308, opened 2020-03-26
Consolidate four vector register loads into one instruction.

AdamHillier:
## What do these changes do?
This PR uses the ['multiple structures' version of the LD1 instruction](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0802a/LD1_advsimd_mult_vector.html) to combine four vector register load instructions into a single instruction, in several places in the optimised Arm64 kernels.

Although (as I expected) there is no benchmarking speed-up, I still think it's valuable because it makes the code a bit shorter and for some reason makes the Arm64 CI run slightly faster (I think Qemu prefers these instructions).

## How Has This Been Tested?
CI.

## Benchmark Results
QuickNet Large on a Raspberry Pi 4, single threaded, averaged over 250 runs (milliseconds +- standard deviation) shows no significant change: (67.59 +- 0.06) -> (67.50 +- 0.27).

### DIFF
--- a/larq_compute_engine/core/bgemm_kernels_arm64.h
+++ b/larq_compute_engine/core/bgemm_kernels_arm64.h
@@ -188,15 +188,8 @@ void BinaryKernelNeonOutOfOrder32BP4x4(
       "ldr w12, [%[params], #" RUY_STR(RUY_OFFSET_DEPTH) "]\n"
 
       // Load the first 64 bytes of LHS and RHS data.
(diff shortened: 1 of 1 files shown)


arashb:
what is the benefit of using multiple structure ld1 according to ARM documentation? only shorter code?

AdamHillier:
> what is the benefit of using multiple structure ld1 according to ARM documentation? only shorter code?

It doesn't explicitly say (or at least I can't find it) but I think so, yes. This seems to make the benchmarking binary slightly smaller as well.

PR #309, opened 2020-03-26
(skip-ci) Add missing copyright notice.

AdamHillier:
I noticed that in TensorFlow this file has a copyright notice at the top, so I've added it to our modified version too.


AdamHillier:
This `(skip-ci)` this isn't working but this is just a comment so probably fine just to override CI and merge if this is approved.

PR #313, opened 2020-03-26
Add ImageNet evaluation tool from TFLite

AdamHillier:
## What do these changes do?

This PR adds a slightly modified version of the TFLite ImageNet evaluation tool: https://github.com/tensorflow/tensorflow/tree/master/tensorflow/lite/tools/accuracy/ilsvrc.

I have:

- Stripped down the code and build files to only what we need for evaluating image classification accuracy (there's a bunch of extra stuff in the TF Lite tool for profiling, et cetera, but we can add that at a later date if we need it).

- In doing so I have removed the existing TF Lite tests, because I don't think they were testing anthing useful for us (mostly things like error messages for the binary).

- Modified the file `tflite/tools/evaluation/stages/tflite_inference_stage.cc` to register our custom ops correctly.

- Modified the file `tflite/tools/evaluation/stages/image_preprocessing_stage.cc` to hard-code bias and scaling factors for image normalisation that best approximate what we do in Larq Zoo (it's not identical because the tool only supports per-image normalization, not per-channel).

- Generated the necessary label `.txt` files to use with the binary (corresponding to the ImageNet validation set). This was a massive pain and took ages because the conversion script they gave was terrible so I'm checking in those two files if that's okay?

- Edited their README and added it to `tflite/tools/accuracy/ilsvrc/README.md`.

## How Has This Been Tested?
The tool reports 62.12% accuracy for QuickNetLarge on the first 2500 images of the ImageNet validation set, which seems to be the right ballpark. It's all existing code with very light modifications, and is a utility rather than core functionality, so I don't think we need to be too concerned about it.

We could potentially add a test that builds this to CI (if only for when we update TF versions)?

## Benchmark Results
N/A


AdamHillier:
@larq/compute-engine how do you feel about checking in the `.txt` files, and do you think we should add building this to our CI?

Tombana:
> @larq/compute-engine how do you feel about checking in the `.txt` files, and do you think we should add building this to our CI?

Since creating the txt file is such a pain, and its only 500KB, I'm fine with that :+1:  but I'm also fine with having the txt file in a private repo.
I agree that it does not need to be in CI since its only a utility. @arashb , @lgeiger what do you think?

Tombana:
> you are copying the entire tool from tensorflow. there is a mechanism in bazel to link against part of a third party without copy.

I agree. Which of the files did you modify? At least one for the registration and another one for the means/scales. I think most other files do not need to be copied. See for example
https://github.com/larq/compute-engine/blob/6ad069f93cf37d35f3af30fa2eaddb348a400d55/larq_compute_engine/mlir/BUILD#L114

AdamHillier:
> you are copying the entire tool from tensorflow. there is a mechanism in bazel to link against part of a third party without copy.

I am linking against the code from TF, and I haven't copied the entire tool. You should look at the build files.

There were two files I *had* to edit: `image_inference_stage` and `image_preprocessing_stage`. This was needed in order to do the correct preprocessing and to register the LCE ops. Then `image_classification_stage`, `imagenet_model_evaluator`, and `imagenet_accuracy_eval` all have direct dependencies on those modified files, so had to also be copied over (with modified `#include` statements).

I have copied over no other C++ code, and inject the remaining dependencies from TF.

Tombana:
> I am linking against the code from TF, and I haven't copied the entire tool. You should look at the build files.
> 
> There were two files I _had_ to edit: `image_inference_stage` and `image_preprocessing_stage`. This was needed in order to do the correct preprocessing and to register the LCE ops. Then `image_classification_stage`, `imagenet_model_evaluator`, and `imagenet_accuracy_eval` all have direct dependencies on those modified files, so had to also be copied over (with modified `#include` statements).
> 
> I have copied over no other C++ code, and inject the remaining dependencies from TF.

My bad, I didn't look close enough.

I'm not 100% sure about this, but it might still be possible to do it without the modified `#include` and therefore saving a few files. If they include only a header `file.h` with a declaration `func();` then if you edit `func()` in `file.cpp` then you can keep using the "old" `file.h` but simply link against the new `file.cpp`. Then you *do* have to copy the `BUILD` targets of those things because you *do* have to change the dependencies, but at least not the source code.

lgeiger:
> Generated the necessary label .txt files to use with the binary (corresponding to the ImageNet validation set). This was a massive pain and took ages because the conversion script they gave was terrible so I'm checking in those two files if that's okay?

I'd be fine with that, I've seen many other repos also have similar lists of ImageNet labels checked in 👍 

The files look very similar to [`imagenet2012_validation_labels.txt`](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/image/imagenet2012_validation_labels.txt) and [`imagenet2012_labels.txt`](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/image/imagenet2012_labels.txt), are they the same? If so we could also download them from there on the fly if we don't want to check them in.

AdamHillier:
> The files look very similar to [`imagenet2012_validation_labels.txt`](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/image/imagenet2012_validation_labels.txt) and [`imagenet2012_labels.txt`](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/image/imagenet2012_labels.txt), are they the same? If so we could also download them from there on the fly if we don't want to check them in.

Damn they may indeed be the same. In why case it seems to be pretty silly for them to have given a terrible conversion script to generate these files that requires you to have the ImageNet2012 devkit installed...

Definitely happy to download them from here instead.

AdamHillier:
> I'm not 100% sure about this, but it might still be possible to do it without the modified `#include` and therefore saving a few files. If they include only a header `file.h` with a declaration `func();` then if you edit `func()` in `file.cpp` then you can keep using the "old" `file.h` but simply link against the new `file.cpp`. Then you _do_ have to copy the `BUILD` targets of those things because you _do_ have to change the dependencies, but at least not the source code.

To add to my previous comment, I also had to make an actual edit to `imagenet_accuracy_eval` to correct an off-by-one error (for some reason it expected 1001 classes).

Tombana:
> To add to my previous comment, I also had to make an actual edit to `imagenet_accuracy_eval` to correct an off-by-one error (for some reason it expected 1001 classes).

:+1: Interesting, maybe its a "background" class?

AdamHillier:
> I'm not 100% sure about this, but it might still be possible to do it without the modified `#include` and therefore saving a few files. If they include only a header `file.h` with a declaration `func();` then if you edit `func()` in `file.cpp` then you can keep using the "old" `file.h` but simply link against the new `file.cpp`. Then you _do_ have to copy the `BUILD` targets of those things because you _do_ have to change the dependencies, but at least not the source code.

Thank you @Tombana - I was able to remove a few of the header files, and one of the C++ files. Everything that remains has non-trivial modifications.

To be completely honest, I'm not so sure about copying the `.cc` file but importing the header file from TF, as they seem pretty tightly-coupled. I feels quite brittle to me, though granted it does save space.

Tombana:
> To be completely honest, I'm not so sure about copying the `.cc` file but importing the header file from TF, as they seem pretty tightly-coupled. I feels quite brittle to me, though granted it does save space.

I agree with your point about the brittle nature of this. I'd say its not only for space but also to make it clear that those are the only changed files when one is looking through the folders. (In fact the modifications are so small that it'd almost be nicer to not copy them at all but use the bazel "patch" feature to simply patch a few lines of the TF files, but I don't think we have to do that here.)
Depending on how other feel about it, I'm also fine with leaving the files as you did it first. It looks like its only 5 extra files, and the entire tool is a lot larger than this so we're still only copying a small fraction.

AdamHillier:
> Depending on how other feel about it, I'm also fine with leaving the files as you did it first. It looks like its only 5 extra files, and the entire tool is a lot larger than this so we're still only copying a small fraction.

Your point about highlighting exactly what's been modified makes a lot of sense, let's go with this.

AdamHillier:
Ah my mistake the off-by-one thing is in `imagenet_model_evaluator.cc`: line 250. I had confused myself, there actually is no change in `imagenet_accuracy_eval.cc` so I've removed it - thanks for spotting that.

I've also added a comment to `imagenet_model_evaluator.cc` highlighting the changed line.

arashb:
why we need to have this tool in LCE repository? this tool has nothing to do with LCE codebase it just uses the LCE ops but we have instructions how to integrate LCE with other binaries.

AdamHillier:
> why we need to have this tool in LCE repository? this tool has nothing to do with LCE codebase it just uses the LCE ops but we have instructions how to integrate LCE with other binaries.

You could say the same thing about the benchmarking binary.

arashb:
> > why we need to have this tool in LCE repository? this tool has nothing to do with LCE codebase it just uses the LCE ops but we have instructions how to integrate LCE with other binaries.
> 
> You could say the same thing about the benchmarking binary.

benchmarking binary is 20 lines not 1000 lines.

Tombana:
I just noticed that these files were quite heavily refactored between TF 2.1 and TF 2.2. So maybe its better to rebase on top of the `tf-2.2` branch and patch the 2.2 version of the tool, so we don't need to edit those files once we merge the 2.2 PR. (Although that means we can not merge this until the 2.2 PR is merged).
In fact, in that branch the `image_preprocessing_stage.cc` file contains a `Normalize()` function in which it would be trivial to hardcode the scales per-channel, see [these lines](https://github.com/tensorflow/tensorflow/blob/acf4951a2f5fdc181ed14c163381c0cf135d9ee6/tensorflow/lite/tools/evaluation/stages/image_preprocessing_stage.cc#L231-L235) .

arashb:
> > why we need to have this tool in LCE repository? this tool has nothing to do with LCE codebase it just uses the LCE ops but we have instructions how to integrate LCE with other binaries.
> 
> You could say the same thing about the benchmarking binary.

even for benchmarking binary I only included the main file there were a doze of other files which I didn't include.

Tombana:
> benchmarking binary is 20 lines not 1000 lines.

I think the number of lines is irrelevant.

Other than that, it's a fair point to discuss whether we want both the benchmark tool and the accuracy tool in a separate repo, in which bazel pulls in LCE as dependency (or maybe a git submodule). I think both are fine, as long as benchmark & accuracy are in the same repo.

arashb:
> > benchmarking binary is 20 lines not 1000 lines.
> 
> I think the number of lines is irrelevant.
> 
> Other than that, it's a fair point to discuss whether we want both the benchmark tool and the accuracy tool in a separate repo, in which bazel pulls in LCE as dependency (or maybe a git submodule). I think both are fine, as long as benchmark & accuracy are in the same repo.

no i strongly disagree the number of lines are very relevant. we can just copy the entire tensorflow bc the number of lines are irrelevant.

Tombana:
> Ah my mistake the off-by-one thing is in `imagenet_model_evaluator.cc`: line 250.
> I've also added a comment to `imagenet_model_evaluator.cc` highlighting the changed line.

Thanks! From that line it seems we could simply add an extra dummy line to the txt file instead? (I'm not sure, it might mess up other things, but worth to try)

AdamHillier:
> no i strongly disagree the number of lines are very relevant. we can just copy the entire tensorflow bc the number of lines are irrelevant.

https://en.wikipedia.org/wiki/Straw_man

> > Ah my mistake the off-by-one thing is in `imagenet_model_evaluator.cc`: line 250.
> > I've also added a comment to `imagenet_model_evaluator.cc` highlighting the changed line.
> 
> Thanks! From that line it seems we could simply add an extra dummy line to the txt file instead? (I'm not sure, it might mess up other things, but worth to try)

We definitely could if it were still checked in, but there's now just a README instruction to download it from GitHub. I guess we could also add an instruction "and add an extra line to it" but that seems pretty ugly.

Tombana:
> We definitely could if it were still checked in, but there's now just a README instruction to download it from GitHub. I guess we could also add an instruction "and add an extra line to it" but that seems pretty ugly.

Agreed, in that case this is fine :+1: .

Tombana:
> no i strongly disagree the number of lines are very relevant. we can just copy the entire tensorflow bc the number of lines are irrelevant.

I don't understand your point. For this tool, only a minimal amount of files have been copied & modified. If the tool would require edits to every tensorflow file, then indeed we would have to copy all of tensorflow, we couldn't really do it another way then?

In which repository the tool belongs, does not depend on the number of lines in the files, but on their functionality IMO.

arashb:
> > no i strongly disagree the number of lines are very relevant. we can just copy the entire tensorflow bc the number of lines are irrelevant.
> 
> I don't understand your point. For this tool, only a minimal amount of files have been copied & modified. If the tool would require edits to every tensorflow file, then indeed we would have to copy all of tensorflow, we couldn't really do it another way then?
> 
> In which repository the tool belongs, does not depend on the number of lines in the files, but on their functionality IMO.

thats exactly the point. you are forcing me to accept copying more than 1000 lines of code from another repository to LCE repo for a tool that its functionality has nothing to do LCE and just uses LCE ops.

AdamHillier:
> thats exactly the point. you are forcing me to accept copying more than 1000 lines of code from another repository to LCE repo for a tool that its functionality has nothing to do LCE and just uses LCE ops.

Nobody is forcing you to do anything, I think we're just expressing a different opinion. If you don't want to add the tool that's your prerogative.

I think it is a good thing to have: it's a practical tool with a lot of value to researchers and potential users - as much value as the benchmarking tool.

Tombana:
I think the following are the pros/cons and for me:

Pros of having the accuracy & benchmark tool in this repo:

- Easier for development: if you change something (e.g. 8-bit quantization) then it will be very easy to compile the benchmark & accuracy tools on your changed branch, so that you can test your changes. If it would be in a separate repo with LCE as dependency, you have to first update that dependency to your own branch etc. I think that would make development harder.

- Similar to how TF has these tools inside the TF repo

Pros of having them in a separate repo:

- Can be considered cleaner, since they are only helper tools and they are not part of the core of LCE.

- 1055 extra lines of build files / code in this repo

Personally I think the easier development heavily outweighs the other arguments.

AdamHillier:
Just to add, the benchmarking stuff is 170 lines, not 20.

AdamHillier:
> I just noticed that these files were quite heavily refactored between TF 2.1 and TF 2.2. So maybe its better to rebase on top of the `tf-2.2` branch and patch the 2.2 version of the tool, so we don't need to edit those files once we merge the 2.2 PR. (Although that means we can not merge this until the 2.2 PR is merged).
> In fact, in that branch the `image_preprocessing_stage.cc` file contains a `Normalize()` function in which it would be trivial to hardcode the scales per-channel, see [these lines](https://github.com/tensorflow/tensorflow/blob/acf4951a2f5fdc181ed14c163381c0cf135d9ee6/tensorflow/lite/tools/evaluation/stages/image_preprocessing_stage.cc#L231-L235) .

Let's pause this then until 2.2 is released.

lgeiger:
Let's take a step back here.

I think the evaluation tool is super useful and it will be necessary to evaluate how our int8 quantization performs.

If we want to have this inside the compute-engine repo or somewhere else, is a valid discussion we should have, both approaches have different tradeoffs.

To be very clear, nobody is forcing anything here. The reason why we are using PRs is so we can have those discussions and I think disagreement is necessary and incredibly useful to ultimately arrive at a better solution.
However, please keep the discussion constructive so we can work on addressing those issues step by step and not get ourselves wrapped up here.

arashb:
> Just to add, the benchmarking stuff is 170 lines, not 20.

just for the record this is [benchmark tool](https://github.com/larq/compute-engine/blob/master/larq_compute_engine/tflite/benchmark/lce_benchmark_main.cc) its just a 46 line main file that calls the real benchmark tool from TF lite.

AdamHillier:
Can be re-opened if needed at a later date.

PR #321, opened 2020-04-01
Switch CI testing from mac to linux

lgeiger:
## What do these changes do?
This unifies our CI testing to run on Ubuntu. We originally opted to run the MLIR tests on Mac to speedup the builds. Since we now have caching we don't need to do this anymore.

## How Has This Been Tested?
CI

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -90,7 +90,7 @@ jobs:
         with:
           python-version: 3.7
       - name: Install dependencies
-        run: pip install tensorflow==${{matrix.tf-version}} larq~=0.9.1 larq_zoo==1.0.b3 tensorflow_datasets==1.3.2 --no-cache
+        run: pip install tensorflow==${{matrix.tf-version}} larq~=0.9.1 larq_zoo==1.0.b3 tensorflow_datasets==1.3.2 packaging --no-cache
       - name: Run Converter test
         run: PYTHONPATH=./ python larq_compute_engine/mlir/python/converter_test.py
 
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -90,9 +90,9 @@ jobs:
         with:
           python-version: 3.7
       - name: Install dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/unittests.yml
```diff
@@ -89,9 +90,9 @@ jobs:
         with:
           python-version: 3.7
       - name: Install dependencies
-        run: pip install tensorflow==${{matrix.tf-version}} larq~=0.9.1 larq_zoo==1.0.b3 tensorflow_datasets==1.3.2 -e . --no-cache
+        run: pip install tensorflow==${{matrix.tf-version}} larq~=0.9.1 larq_zoo==1.0.b3 tensorflow_datasets==1.3.2 packaging --no-cache
```
Tombana, line 93: What is this? How is it still installed if there's no `-e .` ?

.github/workflows/unittests.yml
```diff
@@ -89,9 +90,9 @@ jobs:
...
       - name: Run Converter test
-        run: python larq_compute_engine/mlir/python/converter_test.py
+        run: PYTHONPATH=./ python larq_compute_engine/mlir/python/converter_test.py
```
Tombana, line 95: Is the `PYTHONPATH` thing set by Github Actions?

.github/workflows/unittests.yml
lgeiger: `packaging` is a pip dependency we use in lce to do version comparison.

On linux to properly build the wheel we needed to add [this workaround](https://github.com/larq/compute-engine/blob/master/setup.py#L19) copied from TensorFlow Addons. Unfortunately this prevents us from installing the package without the C++ dependencies. Therefore I add the working directory to the `PYTHONPATH` which removes the need to install LCE via `pip` for testing.

.github/workflows/unittests.yml
lgeiger: This is so we don't need to install compute engine via pip. See above.

PR #324, opened 2020-04-02
Upgrade actions/checkout@v2

lgeiger:
## What do these changes do?
`actions/checkout@v2.1.0` added support for submodules, so we can upgrade to speedup checkout of the submodules.

## How Has This Been Tested?
CI

## Benchmark Results
-


lgeiger:
The new checkout actions requires git 2.18 or newer to support submodules, but the version in the tf container is ancient.

PR #327, opened 2020-04-07
Add 8-bit quantization support to kernels

Tombana:
## What do these changes do?

This PR implements 8-bit quantization support.

We now have three input types and three output types: `{int8, int32, float}` where `int8` means 8-bit quantized and `int32` means 32-bit bitpacked.

To distinguish all the different type combinations, the `OutputTransform` helper class has been added, which will automatically perform the correct output transformation based on the types.

## How Has This Been Tested?
Unit tests have been added. Combinations with reading bitpacked input and output are also tested.
As there are now 7 combinations of input/output types, several helper structs had to be added to deal with all the types.

An `LceTensor<T>` class has also been added to the test to deal with quantization and cleanup the test in general.

The extremely large amount of testing combinations now leads to very long CI times for the qemu runs, so for the qemu runs the tests have been splitted into `SmallTest` and `BigTest`. This is also very convenient for quick local development iterations.

## Benchmark Results
These are the non-optimized implementations so benchmarks are not relevant.


larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -277,10 +390,10 @@ class BaseBConv2DOpModel : public SingleOpModel {
     filter_ = AddInput(filter);
     output_ = AddOutput(output);
 
-    int channels_out = GetShape(filter_)[0];
-    post_activation_multiplier_ =
-        AddInput({TensorType_FLOAT32, {channels_out}});
-    post_activation_bias_ = AddInput({TensorType_FLOAT32, {channels_out}});
+    // int channels_out = GetShape(filter_)[0];
+    // AddInput({TensorType_FLOAT32, {channels_out}});
```
AdamHillier, line 394: Unintentional commented code?

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -637,118 +924,131 @@ void runTest(const TestParam& param) {
...
-  // confuses the pre-processor (EXPECT_DEATH is a macro).
-  typedef BConv2DOpModel<float, float> FP_BConv2DOpModel;
-  typedef BConv2DOpModel<float, std::int8_t> Bitpacked_BConv2DOpModel;
+TEST(BConv2DTests, ReluErrorTest) {
+  LceTensor<float> input_tensor({1, 16, 16, 64});
+  LceTensor<std::int32_t> packed_filter_tensor({128, 3, 3, 64});
+  LceTensor<float> post_tensor({128});
+  LceTensor<float> output_tensor;
+  LceTensor<std::int32_t> packed_output_tensor;
+
```
AdamHillier, line 1040:
```suggestion

  // We have to use typedefs or else the template invocation in the type
  // confuses the pre-processor (EXPECT_DEATH is a macro).
```

(This adds back a comment that I think is useful to have)

Tombana: Thanks for spotting that, this is an extremely useful comment!

larq_compute_engine/core/bgemm_kernels_common.h
```diff
@@ -1,61 +1,23 @@
...
 // Original is in `lite/experimental/ruy/spec.h`
-// Modifications:
-// - bias changes to multiply + add
-// - clamp_min, clamp_max types changed from DstScalar to AccumScalar
-// - (later) 8-bit quantization related stuff
-
+// We simply subclass our `OutputTransform` struct
 template <typename tAccumScalar, typename tDstScalar>
-struct BinaryBasicSpec {
+struct BinaryBasicSpec : OutputTransform<tAccumScalar, tDstScalar> {
```
arashb, line 15: if we have OutputTransform as a member of BinaryBasicSpec instead of subclassing, we can simplify the code quite more.

larq_compute_engine/core/bgemm_kernels_ruy.h
```diff
@@ -69,20 +71,8 @@ struct BgemmKernel<ruy::Path::kStandardCpp, LhsScalar, RhsScalar, DstScalar,
...
-        DstScalar dst_val = static_cast<DstScalar>(accum);
-        if (spec.post_activation_multiplier) {
-          dst_val *= spec.post_activation_multiplier[i];
-        }
-        if (spec.post_activation_bias) {
-          dst_val += spec.post_activation_bias[i];
-        }
-        *ElementPtr(dst, i, j) = dst_val;
+        // Post-process the accumulated value and store the result
+        *ElementPtr(dst, i, j) = output_transform.Run(accum, i);
```
arashb, line 75: spec.output_transform.Run

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -618,17 +884,35 @@ void runTest(const TestParam& param) {
...
+        for (int out_x = 0; out_x < output_height; ++out_x) {
+          for (int out_c = 0; out_c < filter_count; ++out_c) {
+            *out_ptr *= post_activation_multiplier_data[out_c];
+            *out_ptr += post_activation_bias_data[out_c];
+            ++out_ptr;
+          }
         }
       }
     }
+  } else {
```
arashb, line 900: we can move the comments on top of if and remove the empty else block

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -757,15 +1056,113 @@ TEST(BConv2DTests, BConvErrorTest) {
...
-            {TensorType_FLOAT32, {128, 3, 3, 64}}, {TensorType_UINT8, {}}, 64,
+            compute_engine::tflite::Register_BCONV_2D64_OPT, input_tensor,
+            packed_filter_tensor, output_tensor, post_tensor, post_tensor, 64,
             1, 1, Padding_SAME, 0, ActivationFunctionType_NONE, 1, 1, false,
             true, 1);
       },
       "Writing bitpacked output is only supported with valid or one-padding.");
 }
 
+TEST(BConv2DTests, Int8ErrorTest) {
```
arashb, line 1067: we need to create a separate files for 8-bit bconv tests and a separate file for test helper functions and also a separate file for common things used for different tests (e.g, Bconv2DOpModel)

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -386,6 +496,88 @@ class BConv2DOpTest : public ::testing::TestWithParam<TestParamTuple> {
...
+    // per_channel_quantization is used.
+    float s = 1.0f / float(reciprocal_scale);
+
+    per_channel_quantization = true;
+    std::size_t count = shape[channel_index];
+    per_channel_quantization_scales = std::vector<float>(count, s);
+    per_channel_quantization_offsets = std::vector<std::int64_t>(count, 0);
+  }
+
+  T Quantize(const std::int32_t x) const {
```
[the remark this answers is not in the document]
Tombana: For our tests we only allow scales of the form `1/n` so that we can represent `-1` and `+1` without error. Therefore we have some specialized `Quantize` functions that are not available in TF code.

larq_compute_engine/core/packbits.h
```diff
@@ -421,7 +457,7 @@ template <typename TBitpacked, typename TUnpacked>
 void unpack_bitfield(const TBitpacked in, TUnpacked*& out,
                      std::size_t num_elements) {
   for (size_t i = 0; i < num_elements; ++i) {
-    *out++ = (in & (1ULL << i)) ? TUnpacked(-1.0f) : TUnpacked(1.0f);
+    *out++ = (in & (1ULL << i)) ? TUnpacked(-1) : TUnpacked(1);
```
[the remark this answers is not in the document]
Tombana: The resulting code will indeed be the same. However I think a `TUnpacked(-1)` is slightly cleaner.

larq_compute_engine/core/bconv2d_output_transform.h
```diff
@@ -0,0 +1,124 @@
...
+  }
+};
+
+template <typename AccumScalar, typename ActivationScalar>
+struct OutputTransform {};
+
+// Parameters that are needed only for float kernels
+template <typename AccumScalar>
+struct OutputTransform<AccumScalar, float> : OutputTransformBase<AccumScalar> {
+  const float* post_activation_multiplier = nullptr;
```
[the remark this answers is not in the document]
Tombana: In order to add a constructor here, it would require access to several things in `kernels/bconv2d.cc` such as `vector<>`s in the bconv2dparams struct. Therefore I think its cleaner to leave the construction code back in `bconv2d.cc` . What do you think?

larq_compute_engine/core/packbits.h
```diff
@@ -301,16 +321,28 @@ inline void pack_optimized(const float* in, std::uint64_t* out) {
...
-inline void pack_bitfield(const TIn* in, TOut* out) {
-  if (bitpack_order == BitpackOrder::Canonical)
-    pack_canonical(in, out);
-  else
-    pack_optimized(in, out);
+inline void pack_bitfield(const TIn* in, TOut* out,
+                          const std::int32_t zero_point) {
+  // Note: The expressions in these if-statements are known at compile-time so
+  // they are all optimied away
+  constexpr bool UseZeroPoint = !std::is_floating_point<TIn>::value;
```
arashb, line 328: `UseZeroPoint` -> `is_quantized`?

larq_compute_engine/core/packbits.h
```diff
@@ -301,16 +321,28 @@ inline void pack_optimized(const float* in, std::uint64_t* out) {
 
 // Helper function
 template <BitpackOrder bitpack_order, class TIn, class TOut>
-inline void pack_bitfield(const TIn* in, TOut* out) {
-  if (bitpack_order == BitpackOrder::Canonical)
-    pack_canonical(in, out);
-  else
-    pack_optimized(in, out);
+inline void pack_bitfield(const TIn* in, TOut* out,
+                          const std::int32_t zero_point) {
```
arashb, line 325:
```suggestion
                          const std::int32_t zero_point = 0) {
```

arashb, line 325: by defining default value for `zero_point` the user does not need to pass this argument if its not using this function for non-quantized inputs.

larq_compute_engine/core/packbits.h
```diff
@@ -301,16 +321,28 @@ inline void pack_optimized(const float* in, std::uint64_t* out) {
...
+    else
+      pack_optimized(in, out);
+  }
 }
 
 template <BitpackOrder bitpack_order, class TIn, class TOut>
 inline void packbits_array(const TIn* input_array, const std::size_t n,
-                           TOut* bitpacked_array) {
+                           TOut* bitpacked_array,
+                           const std::int32_t zero_point) {
```
arashb, line 345:
```suggestion
                           const std::int32_t zero_point = 0) {
```

larq_compute_engine/core/tests/packbits_arm32_tests.cc
```diff
@@ -23,8 +23,8 @@ TEST(BitpackingTests, BitpackingARM32) {
   }
 
   std::vector<std::uint64_t> expected(n_packed);
-  ce::core::packbits_array<ce::core::BitpackOrder::Optimized>(input.data(), n,
-                                                              expected.data());
+  ce::core::packbits_array<ce::core::BitpackOrder::Optimized>(
+      input.data(), n, expected.data(), 0);
```
arashb, line 27:
```suggestion
      input.data(), n, expected.data());
```

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -470,32 +537,180 @@ inline void GetConvParamsType(const TfLiteBConv2DParams& conv_params,
...
-void EvalOpt(TfLiteContext* context, TfLiteNode* node,
-             TfLiteBConv2DParams* params) {
-  const auto* input = GetInput(context, node, 0);
-  const auto* filter = GetInput(context, node, 1);
+using compute_engine::core::OutputTransform;
+using compute_engine::core::OutputTransformBase;
+
+// Fill the parts of the OutputTransform struct that are common to each
+// destination type
+void GetBaseParams(TfLiteContext* context, TfLiteNode* node,
```
arashb, line 545: I think all of the helper functions added here, need to move to a seprate file (`quantization_utils.h`?). the bconv2d file is already huge.

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -633,51 +872,77 @@ void EvalRef(TfLiteContext* context, TfLiteNode* node,
...
+          context, node, params);
+    case 64:
+      return Eval_stage4<kernel_type, SrcScalar, DstScalar, std::uint64_t>(
+          context, node, params);
+  };
+  return kTfLiteError;
+}
+
+template <KernelType kernel_type, typename SrcScalar>
+TfLiteStatus Eval_stage2(TfLiteContext* context, TfLiteNode* node,
```
arashb, line 917: `Eval_stage2` what happens during stage2?

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -633,51 +872,77 @@ void EvalRef(TfLiteContext* context, TfLiteNode* node,
...
-      return kTfLiteOk;
-    }
+    EvalRef<SrcScalar, TBitpacked, DstScalar>(context, node, params);
+    return kTfLiteOk;
   }
+  return kTfLiteError;
+}
+
+template <KernelType kernel_type, typename SrcScalar, typename DstScalar>
+TfLiteStatus Eval_stage3(TfLiteContext* context, TfLiteNode* node,
```
arashb, line 903: what happens during stage3?

larq_compute_engine/core/packbits.h
Tombana: For this function, I deliberately left it like this, because its an internal function (we only use it inside `packbits.h`) so this way we can't accidentally forget to pass the `zero_point` argument. For a function like `packbits_matrix`, which is used everywhere, I did set the `zero_point` default value, so that it would be cleaner and easier to use.
I don't feel too strongly about this, so I'm also fine with setting `zero_point=0` by default here.

larq_compute_engine/tflite/kernels/bconv2d_impl.h
```diff
@@ -136,31 +147,32 @@ inline void BConv2D(const ConvParams& params, const RuntimeShape& input_shape,
       static std::vector<TBitpacked> input_data_bp_buffer;
       // The input tensor has this shape which we bitpack along the channels
       // dimension [batch, input height, input width, channels].
-      ce::core::packbits_tensor(input_shape, input_data, packed_input_shape,
-                                input_data_bp_buffer);
+      ce::core::packbits_tensor(input_shape, input_data, params.input_offset,
```
arashb, line 150: what is `input_offset`?

larq_compute_engine/tflite/kernels/bconv2d.cc
Tombana: I agree that it makes sense to move this. However these functions need to access our Bconv2dparams struct. Shall we move that struct to a separate header file as well?

larq_compute_engine/tflite/kernels/bconv2d_impl.h
Tombana: Its `zero_point`, but for some structs, tflite renamed it to `offset` so we can't change that name unfortunately.

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -496,51 +443,110 @@ void runTest(const TestParam& param) {
...
+  }
 
   // Bitpack filters
   std::size_t filter_rows_bp, filter_cols_bp, filter_bitpadding;
   using namespace compute_engine::core;
   packbits_matrix<BitpackOrder::Canonical>(
       filters_data.data(), filter_count * filter_height * filter_width,
       input_depth, packed_filters_data, filter_rows_bp, filter_cols_bp,
-      filter_bitpadding, Axis::RowWise);
+      filter_bitpadding, Axis::RowWise, 0);
```
arashb, line 548:
```suggestion
      filter_bitpadding, Axis::RowWise);
```

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -496,51 +443,110 @@ void runTest(const TestParam& param) {
   const auto is_reference_registration =
       (registration == compute_engine::tflite::Register_BCONV_2D32_REF);
 
-  if ((padding == Padding_SAME && pad_values == 0) &&
-      is_reference_registration) {
+  if (is_reference_registration &&
```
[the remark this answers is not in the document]
Tombana: I think I first added more conditions to that if-statement, and then later removed them again, and that way they ended up exchanged.

larq_compute_engine/tflite/kernels/bconv2d.cc
arashb: yes, I think so

larq_compute_engine/core/packbits.h
arashb: it seems to me this function accepts general types (both quantinzed and non-quantized inputs), therefore I think for the general functions it makes sense to set the default value but specialized functions I agree with you.

arashb: oh wait. Nevermind. I think here your approach is better.

larq_compute_engine/tflite/kernels/bconv2d.cc
Tombana: I moved these functions and the struct to their own separate files in 654ef34

arashb:
> beside the following file renaming, LGTM 👍
> 
> * `bconv2d_output_transform_setup.h` -> `bconv2d_output_transform_utils.h` (we consistently use `_utils.h` through the repo)
> * `bconv2d_int8.cc` -> `bconv2d_int8_test.cc` test files should end with `_test`
> * `tflite/tests/bconv2d_op.h` -> `tflite/tests/bconv2d_op_model.h`(In this file we create OpModel - a model with one op - and not an Op)

@Tombana since the PR is not only adding the 8-bit kernels but also include lots of code refactoring, would you please also do the benchmarks on android to make sure this PR does not cause any performance degradation?

Tombana:
> @Tombana since the PR is not only adding the 8-bit kernels but also include lots of code refactoring, would you please also do the benchmarks on android to make sure this PR does not cause any performance degradation (its very unlikely but just to be on the safe side)?

I benchmarked Quicknet on the Pixel 1. On the int8 branch `20.372 ms`, on master it was `20.368 ms` (0.004 ms faster), so I think thats random noise and there's no real performance degradation.

PR #329, opened 2020-04-09
Consistently use bazel test and --test_output=all on CI

lgeiger:
## What do these changes do?
This PR changes CI to consistently use `bazel test --test_output=all`. This will hopefully make sure that asserts are enabled during all of our tests.

## How Has This Been Tested?
CI


.github/workflows/unittests.yml
```diff
@@ -48,9 +48,9 @@ jobs:
         run: ./configure.sh <<< $'n\n'
         shell: bash
       - name: "TF Lite Arm32: Cross-compile and run unit tests in qemu"
-        run: ./bazelisk run larq_compute_engine/tests:cc_tests_arm32_qemu --config=rpi3 --compilation_mode=fastbuild
+        run: ./bazelisk test larq_compute_engine/tests:cc_tests_arm32_qemu --config=rpi3 --compilation_mode=fastbuild --test_output=all
```
AdamHillier, line 51:
```suggestion
        run: ./bazelisk test larq_compute_engine/tests:cc_tests_arm32_qemu --config=rpi3 --test_output=all
```

.github/workflows/unittests.yml
```diff
@@ -48,9 +48,9 @@ jobs:
...
       - name: "TF Lite Aarch64: Cross-compile and run unit tests in qemu"
-        run: ./bazelisk run larq_compute_engine/tests:cc_tests_aarch64_qemu --config=aarch64 --compilation_mode=fastbuild
+        run: ./bazelisk test larq_compute_engine/tests:cc_tests_aarch64_qemu --config=aarch64 --compilation_mode=fastbuild --test_output=all
```
AdamHillier, line 53:
```suggestion
        run: ./bazelisk test larq_compute_engine/tests:cc_tests_aarch64_qemu --config=aarch64 --test_output=all
```

.github/workflows/unittests.yml
AdamHillier: (Test already sets fastbuild)

lgeiger:
> This should have the very nice side benefit of speeding up the end2end test, because previously it was build with opt and so couldn't re-use the compilation of the MLIR filecheck, which was fastbuild.

That makes sense, I was wondering why it now has to rebuilt and is not streamed from the cache :+1:

lgeiger:
@AdamHillier do you have an idea why the ARM tests timeout now?

AdamHillier:
> @AdamHillier do you have an idea why the ARM tests timeout now?

Good question, I'm not sure. Previously they took a bit longer than this run did: https://github.com/larq/compute-engine/runs/567324056.

lgeiger:
:+1: I'll increase the timeout if it fails after this run as well.

Tombana:
In the int8 PR the ARM tests are 'fixed' because the qemu tests only run a small subset of parameters (this should probably be split into a separate PR from the int8 PR), so i think we can simply increase the timeout for now and fix it later.

lgeiger:
> I wonder if bazel allows more time for "medium" tests when running -c opt or something like that.

I think it is because we used `bazel run` before, it probably doesn't check the timeout.

PR #258, opened 2020-03-03
Upgrade to TF 2.2

lgeiger:
TensorFlow 2.2.0 will be released 🔜 . This PR upgrades the TensorFlow dependency to point to the 2.2 release branch of TensorFlow. There've been quite a few changes between the two versions so this requires a thorough review.

## TODO
- [x] Get builds to pass
- [x] Upgrade to TensorFlow 2.2.0 stable
- [x] Speed up CI builds since this doubles build time for the converter and End2End tests #287
- [x] Update ImageNet eval tool
- [x] Benchmark models to ensure we don't have performance regressions
- [x] Switch to new Ruy profiler https://github.com/tensorflow/tensorflow/commit/b0f057d1b4ef78f0bba7e8893debe6b4310711ee

## Benchmarks

I ran some preliminary benchmarks on my personal phone for 500 runs with a QuickNet like model and didn't see any performance regressions:

branch | 1 thread (ms) | 4 threads (ms)
--- | --- | ---
master | 26.6 ± 0.3 | 17.5 ± 0.5
PR | 26.5 ± 0.2 | 17.2 ± 0.8

PR #331, opened 2020-04-09
Upgrade ARM gcc compiler to 9.2.1

lgeiger:
## What do these changes do?
I tried to fix the build failure in #258 by upgrading the [Aarch64 gcc version to 9.2](https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-a/downloads). Unfortunately this didn't fix the build, but since the code is already here, this PR upgrades the compiler version anyway.
I don't think this upgrade has any downsides, but please correct me if I am wrong.

I ran buildifier, so only dfc27501e19cf2474c6263aaa097584bf8f5bae7 is relevant for review.

## How Has This Been Tested?
CI

## Benchmark Results
@Tombana do you think we should re-benchmark on the RP4?

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -25,15 +25,14 @@ http_archive(
 )
 
 http_archive(
-    # This is the latest `aarch64-linux-gnu` compiler provided by ARM
+    # This is the latest `aarch64-none-linux-gnu` compiler provided by ARM
     # See https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-a/downloads
-    # Download size: 260 MB compressed, 1.5 GB uncompressed
-    # The archive contains GCC version 8.3.0
+    # The archive contains GCC version 9.2.1
     name = "aarch64_compiler",
     build_file = "//third_party:arm_compiler.BUILD",
-    sha256 = "8ce3e7688a47d8cd2d8e8323f147104ae1c8139520eca50ccf8a7fa933002731",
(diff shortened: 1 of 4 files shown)


AdamHillier:
Ah this might mean we can use C++17 features like `if constexpr`! When I tried previously it was fine except on Arm.

lgeiger:
I just checked, that our Raspberry Pi has a new enough version of GLIBC require for this compiler to work.

arashb:
> Ah this might mean we can use C++17 features like `if constexpr`! When I tried previously it was fine except on Arm.

Currently, the TF is using C++14 (I checked a few months ago) and we have also hardcoded in LCE  configure [here](https://github.com/larq/compute-engine/blob/f801c05b9373528b29b8732ae55e3cc605d26747/configure.sh#L70) to be compile with c++14. So I am surprised how it has compiled it for you.

AdamHillier:
> Currently, the TF is using C++14 (I checked a few months ago, they might have changed in the meantime) and we have also hardcoded it in LCE configure [here](https://github.com/larq/compute-engine/blob/f801c05b9373528b29b8732ae55e3cc605d26747/configure.sh#L70) to be compiled with c++14. So I am surprised how it has compiled it for you at all.

Yeah on reflection this won't quite allow us to use C++17 because this updates only the Aarch64 compiler, not the 32-bit Arm one, which also doesn't support it.

PR #332, opened 2020-04-12
AArch64 BMLA: Use pairwise addition to accumulate popcount results

lgeiger:
## What do these changes do?
This changes the accumulate of the popcount result in the AArch BMLA to use pairwise addition. See [Figure 7.9 in the arm docs](https://developer.arm.com/docs/den0024/latest/aarch64-floating-point-and-neon/aarch64-neon-instruction-format). The final accumulate is then done using [`uadalp`](https://developer.arm.com/docs/dui0801/j/a64-simd-vector-instructions/uadalp-vector).
This removes the need for the `ins` instructions, while keeping the same amount of arithmetic operations.
I couldn't find the cycle counts for these operations so I am not sure what the theoretical speedup of this should be, let me know if I am missing something here.

## How Has This Been Tested?
CI

## Benchmark Results
~~Initial benchmarks on my phone in the warm easter sun were promising, but I will properly benchmark the changes next week in a controlled environment without my phone overheating.~~

I ran some full model benchmarks on my phone with `num_calls=250`:

| Model          | Threads | master (ms)    | PR (ms)        |
| -------------- | ------- | -------------- | -------------- |
| QuickNet       | 1       | 25.2 ± 0.3     | **24.3 ± 0.3** |
| QuickNet       | 2       | 18.4 ± 0.3     | **17.5 ± 0.5** |
| QuickNet       | 4       | **15.0 ± 0.3** | 15.2 ± 0.4     |
|                |         |                |                |
| QuickNet Large | 1       | 36.8 ± 0.3     | **35.7 ± 0.3** |
| QuickNet Large | 2       | 26.4 ± 0.3     | **25.9 ± 0.3** |
| QuickNet Large | 4       | 21.8 ± 0.3     | **21.7 ± 0.5** |
|                |         |                |                |
| QuickNet XL    | 1       | 66.7 ± 0.3     | **64.2 ± 0.3** |
| QuickNet XL    | 2       | 47.1 ± 0.5     | **44.9 ± 0.4** |
| QuickNet XL    | 4       | 38.9 ± 0.3     | **37.9 ± 0.2** |

PR #333, opened 2020-04-14
AArch64 BMLA: Reduce temporary register usage

lgeiger:
## What do these changes do?
This PR changes the order of the BMLA instructions reduce the usage of temporary registers. Since this changes the data dependency of instructions this PR requires a thorough review.

Register `v31` is now no longer used, do we need to change something [here](https://github.com/larq/compute-engine/blob/24d853fb964ff44204f60bfd6684ed0b8772dbf7/larq_compute_engine/core/bgemm_kernels_arm64.h#L444-L452) as well?

## How Has This Been Tested?
CI

## Benchmark Results
I ran benchmarks on my phone with `num_calls=250`:
| Model          | Threads | master (ms) | PR (ms)    |
| -------------- | ------- | ----------- | ---------- |
| QuickNet       | 1       | 24.0 ± 0.3  | 24.2 ± 0.3 |
| QuickNet       | 2       | 18.2 ± 0.3  | 18.0 ± 0.3 |
|                |         |             |            |
| QuickNet Large | 1       | 36.2 ± 0.4  | 35.3 ± 0.4 |
| QuickNet Large | 2       | 26.3 ± 0.3  | 26.3 ± 0.4 |
|                |         |             |            |
| QuickNet XL    | 1       | 64.4 ± 0.6  | 64.1 ± 0.3 |
| QuickNet XL    | 2       | 45.9 ± 0.4  | 45.5 ± 0.3 |

It looks like changing the order of the instructions doesn't introduce any slow down, but take the exact numbers with a grain of salt since my phone can have quite a high run variation.

## Related issue number
https://github.com/larq/compute-engine/pull/332#issuecomment-613387689

### DIFF
--- a/larq_compute_engine/core/bgemm_kernels_arm64.h
+++ b/larq_compute_engine/core/bgemm_kernels_arm64.h
@@ -9,56 +9,56 @@ using namespace ruy;
 
 // clang-format off
 
-// temporary NEON registers: v28,v29,v30,v31
+// temporary NEON registers: v28,v29,v30
 #define LCE_BMLA(Vd, Vr, Vl1, Vl2, Vl3, Vl4) \
   "eor v28.16b, " #Vr".16b, " #Vl1".16b\n"    \
   "eor v29.16b, " #Vr".16b, " #Vl2".16b\n"    \
-  "eor v30.16b, " #Vr".16b, " #Vl3".16b\n"    \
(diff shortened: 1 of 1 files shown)


lgeiger:
#350 Added a faster 8x4 kernel which will be used in most cases, so optimizing the register usage here won't really matter much and reduces the readability. I am closing this for now.

PR #342, opened 2020-04-29
:arrow_up: lint dependencies

lgeiger:


lgeiger:
Opened an issue at https://github.com/tensorflow/datasets/issues/1977.

I upgraded tfds in 07ebbbb27ff32630bbb49aea8c4520e18c196a4e and switched to using `tf_flowers` for testing. This has the added benefit that we can directly read it from the TFDS public GCS folder and don't need to download and extract it.

PR #348, opened 2020-05-07
Add TF 2.2 to converter test matrix

lgeiger:
## What do these changes do?
This add tests against TF 2.2 to the Python side of our converter.
Currently the 2.2.0 pip package hasn't been uploaded but 2.2.0rc4 includes all relevant Python changes so it is fine to use it for testing.

## How Has This Been Tested?
CI

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -86,7 +86,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        tf-version: [1.14.0, 1.15.2, 2.0.1, 2.1.0]
+        tf-version: [1.14.0, 1.15.2, 2.0.1, 2.1.0, 2.2.0rc4]
     if: "!contains(github.event.head_commit.message, 'ci-skip')"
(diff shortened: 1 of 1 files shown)


.github/workflows/unittests.yml
```diff
@@ -86,7 +86,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        tf-version: [1.14.0, 1.15.2, 2.0.1, 2.1.0]
+        tf-version: [1.14.0, 1.15.2, 2.0.1, 2.1.0, 2.2.0rc4]
```
arashb, line 89: 2.2.0rc4 -> 2.2.0?

lgeiger: See my comment above:
> Currently the 2.2.0 pip package hasn't been uploaded but 2.2.0rc4 includes all relevant Python changes so it should be fine to use it for testing.

I'd be fine to wait for the 2.2.0 stable pip package, but I wanted to check that CI is still happy with it.

PR #349, opened 2020-05-07
Upgrade TensorFlow to nightly version

lgeiger:
## What do these changes do?
This upgrades TensorFlow to version [`a80d96c6634cd005b3841d462030448f9f551a14`](https://github.com/tensorflow/tensorflow/commits/a80d96c6634cd005b3841d462030448f9f551a14). This is required for the upcoming work on int8 quantization aware training.

This includes many changes that depend on each other:
- Ruy was moved to https://github.com/google/ruy
- Ruy changed their spec API
- Replaced grappler constant folding with MLIR pass. This is inline with what TensorFlow does and is required for QAT
- Upgraded LLVM/MLIR
- Updates to MLIR passes according to what TensorFlow does on master
- Truncate op names to ~50 characters since LCE ops can have version long location names. This saves some space in the flatbuffer and still allows for good debugging compared to completely stripping the names

## How Has This Been Tested?
CI and manual testing by @AdamHillier  and me.

## Benchmark Results

Benchmarked with QuickNetLarge on my personal phone:

| branch | 1 thread (ms) | 2 threads (ms) |
| ------ | ------------- | -------------- |
| master | 35.8 ± 0.3    | 25.7 ± 0.2     |
| PR     | 35.6 ± 0.3    | 25.2 ± 0.4     |


AdamHillier:
For reviewers: by far the biggest thing that affects the kernels is Ruy having been upgraded to a version that lives outside of TF. In this new version, all variations of `Spec` have been turned into variations of `MulParams`.

lgeiger:
I don't know what's going on with the TFLite makefile here.
@Tombana could you have a look at it when you have time?

lgeiger:
> I don't know what's going on with the TFLite makefile here.
@Tombana could you have a look at it when you have time?

It looks like it is not possible anymore to overwrite the op registration in the way it is done in our build script. Do we still need to support the make build, or are there platforms we cannot cross compile for?

Tombana:
> > I don't know what's going on with the TFLite makefile here.
> > @Tombana could you have a look at it when you have time?
> 
> It looks like it is not possible anymore to overwrite the op registration in the way it is done in our build script. Do we still need to support the make build, or are there platforms we cannot cross compile for?

I think we can still do it. Originally we had both files (the TFLite one and ours) named `register.cc` (compiled to `register.o`). Then when its put in the static library with the `ar` tool, our `register.o` would overwrite their `register.o`. This way there would be no symbols that were defined twice.

However then we renamed our file to `lce_register.o` and used `ar`'s `-m` flag to make sure that our `lce_register.o` came *after* tflite's `register.o`. This way it would always use our symbols.
Now it seems, however, that there's a problem with having the symbol defined twice, I'm not sure what changed. Maybe their symbol was first a [weak symbol](https://en.wikipedia.org/wiki/Weak_symbol) and now not anymore?

In either case, we should be able to get it working by changing this `-m` (move to lower precedence) flag
https://github.com/larq/compute-engine/blob/c75b32ee36aa2150025ca565a9afdf5a42cb7744/larq_compute_engine/tflite/build_make/Makefile#L136-L138
to a `-d` (delete) flag like
```
$(AR) -d $(LIB_PATH) register.o 
```
to simply disregard their `register.o` file entirely.

larq_compute_engine/core/bconv2d_output_transform.h
```diff
@@ -6,7 +6,7 @@
 #include <limits>
 
 #include "tensorflow/lite/kernels/internal/common.h"
-#include "tensorflow/lite/kernels/internal/round.h"
+#include "tensorflow/lite/kernels/internal/reference/round.h"
```
arashb, line 9: is this a reference impl. for rounding? are there other impl.s for rounding as well?

larq_compute_engine/core/bgemm_kernels_common.h
```diff
@@ -2,30 +2,36 @@
...
 template <typename tAccumScalar, typename tDstScalar>
-struct BinaryBasicSpec {
+struct BinaryMulParams {
   using AccumScalar = tAccumScalar;
   using DstScalar = tDstScalar;
 
   OutputTransform<AccumScalar, DstScalar> output_transform;
 
-  // This is identical to `ruy::BasicSpec`
+ public:
```
arashb, line 20: `structs` attributes/methods are by default `public`. the `public` identifier here  is not necessary.

AdamHillier: Makes sense, but just for reference it was copied from here: https://github.com/google/ruy/blob/bec99d61aa9d1e4d07a85f2be326821a1600593c/ruy/mul_params.h#L125-L145

arashb: wow, they have completely changed the Ruy spec impl. if you look a few lines above now is a `class` and not a `struct`.  We don't need other parts of the new impl or RUY spec/Mul_params?

larq_compute_engine/core/bconv2d_output_transform.h
lgeiger: Good catch, I think this include might be wrong or not necessary, I'll ceck.

larq_compute_engine/core/bgemm_kernels_common.h
arashb: also its not a final class and we can actually subclass it to add our extensions to it without losing the compatibility.

larq_compute_engine/core/bconv2d_output_transform.h
lgeiger: They moved the definition in https://github.com/tensorflow/tensorflow/commit/a4c2fcf05b831c726346d78b4b3dddfd823d93c0 so `reference/round.h` but I guess the correct header was implicitely included in `common.h`.

larq_compute_engine/core/bgemm_kernels_common.h
arashb: @Tombana does it make sense to integrate/extend our `output_transorm` functionality to that?

AdamHillier: Ah interesting. I don't think we were thinking too much about it at the time, just trying to get things to successfully compile. The reason we didn't include the other stuff in the new `MulParams` class is that we have the output multiplier, zero point, clamp min/max etc in other parts of LCE (e.g. the `OutputTransform` struct) and didn't want to have to mess around with that too.

Tombana: Since some time ago, at the PR where we started using our own `OutputTransform`, I think we basically don't use these specs/mulparams anymore, except for passing through our own outputtransform.

AdamHillier: Would you rather extend the `MulParams` class?

Tombana: I think me and @arash discussed in a call whether to extend it (I guess you could subclass it in either direction?) or to have it as a member variable. I think the opinions were rather neutral, and then we opted for the member variable. I'm still neutral.

Tombana: Its a good point that if we make `OutputTransform` a subclass of `MulParams` then we don't have to copy any `MulParams` code, so thats easier to keep up to date.
However this won't work for us since `OutputTransform` is also used in places where we don't have RUY.
We could do it the other way around, having `MulParams` as a subclass of `OutputTransform` (I think that was my original implementation when I pushed the PR), but thats almost exactly the same as the current method.

arashb: For this PR, if everything works and there is no regression I am in favour of continuing with the current code but I think we should take a deep look into the RUY changes and whats the best way to adopt them.

arashb, line 20:
```suggestion
```

lgeiger: Removed in 20a6de5

PR #350, opened 2020-05-09
Add 8x4 Arm64 kernel to improve performance

AdamHillier:
In my spare time I'm been messing around with the optimised Arm64 kernels.

This PR adds an 8x4 kernel (for 64-bit bitpacked input) which improves single-threaded benchmark performance by around 6%.

## What do these changes do?

The current optimal Arm kernel (for 64-bit bitpacking) has a 4x4 layout and uses 32-bit accumulators. This PR additionally introduces an 8x4 version of this kernel that uses 16-bit accumulators.

After the accumulation loop, the first operation is an `unsigned shift left long` instruction which performs the necessary back-transformation left-shift by 1, and writes the result into 32-bit vector registers. This means that all subsequent 'post-processing' (clamping, post-multiply and post-addition) is done in Int32/Float32, as in the existing kernel.

### Performance improvements

Performance is improved (compared to the 4x4 kernel) in two ways:

* **Increased weight sharing**: The 8x4 layout means that the RHS data gets used two times more for each load, or equivalently the number of RHS memory loads is halved.

* **Faster BMLA pseudo-instructions**: With 8 rows and 16-bit accumulators, each column of accumulators fits into a single NEON register. With 16-bit accumulators rather than 32-bit accumulators, there's no need for an extra `uadalp` instruction to pack 16-bit results into 32-bits (specifically, there's no need for the equivalent of [this instruction](https://github.com/larq/compute-engine/blob/f801c05b9373528b29b8732ae55e3cc605d26747/larq_compute_engine/core/bgemm_kernels_arm64.h#L25) in the current kernel). As a result, the proposed `BMLA` macro uses 24 instructions to process eight rows of data, as opposed to 13 instruction to process four rows of data in the current 4x4 kernel.

### Overflow risk with 16-bit accumulators

There is a potential risk of overflowing the 16-bit accumulators if `filter_height` * `filter_width` * `channels_in` >= 2^16. As a result, this PR does not replace the existing 4x4 kernel, and safely falls back to the old kernel if this overflow threshold is breached.

In reality, we will always use the new 8x4 kernel for models of any reasonable size: a `3x3` kernel would have to have > 7000 input channels to present an overflow risk.

## How Has This Been Tested?

All existing kernel tests pass, including the 'big' versions of the tests (that I've run locally, as they don't run by default on CI).

## Benchmark Results

I have benchmarked current master (the 'Baseline') and this PR on my Android phone (a OnePlus 6) and on a Raspberry Pi Model 4B. I used the LCE benchmark utility with `num_runs=250` and report the average and the standard-deviation.

#### Single-threaded

| Device    | Model         | Baseline      | PR            | Speed-up |
|-----------|---------------|---------------|---------------|----------|
| OnePlus 6 | QuickNet      | 19.83 +- 0.28 | 18.72 +- 0.21 | 5.6%     |
| OnePlus 6 | QuickNetLarge | 28.41 +- 0.22 | 26.99 +- 0.22 | 5.0%     |
| OnePlus 6 | QuickNetXL    | 50.65 +- 0.52 | 47.86 +- 0.33 | 5.5%     |
| RPi 4     | QuickNet      | 37.82 +- 0.09 | 34.94 +- 0.17 | 7.6%     |
| RPi 4     | QuickNetLarge | 55.98 +- 0.12 | 52.66 +- 0.15 | 5.9%     |
| RPi 4     | QuickNetXL    | 99.09 +- 0.46 | 92.00 +- 0.26 | 7.2%     |

#### Multi-threaded (4 threads)

| Device    | Model         | Baseline      | PR            | Speed-up |
|-----------|---------------|---------------|---------------|----------|
| OnePlus 6 | QuickNet      | 10.36 +- 0.40 | 10.08 +- 0.31 | 2.7%     |
| OnePlus 6 | QuickNetLarge | 14.65 +- 0.55 | 14.14 +- 0.42 | 3.5%     |
| OnePlus 6 | QuickNetXL    | 26.32 +- 0.53 | 25.34 +- 0.55 | 3.7%     |
| RPi 4     | QuickNet      | 17.73 +- 0.34 | 17.17 +- 0.45 | 3.2%     |
| RPi 4     | QuickNetLarge | 26.56 +- 0.56 | 25.74 +- 0.43 | 3.1%     |
| RPi 4     | QuickNetXL    | 43.63 +- 1.09 | 41.71 +- 1.06 | 4.4%     |

There is an average single-threaded speed-up of 6.1% and multi-threaded speed-up of 3.4%.


larq_compute_engine/core/bgemm_kernels_common.h
```diff
@@ -53,13 +53,13 @@ struct BinaryKernelParams {
   std::int32_t clamp_max;
   std::int32_t backtransform_add;
   std::uint8_t flags;
-  T dst_tmp_buf[LhsCols * RhsCols];
+  float dst_tmp_buf[LhsCols * RhsCols];
```
AdamHillier, line 56: This I think is a mistake on master, as in this case T is the bitpacked type not the destination type.

AdamHillier:
### Prospects for an 8x8 kernel

If we don't use `RUY_OPT_MAX_STREAMING`, then there are enough registers to be able to implement an 8x8 kernel. I have experimented with disabling the `RUY_OPT_MAX_STREAMING` part of this kernel and haven't noticed any significant benchmarked performance difference, so I don't think that would be a problem.

However, I also don't think that going to 8x8 would bring significant futher performance gains in and of itself, because as far as I can tell most of the performance gain in this PR comes from the slightly more optimal `LCE_BMLA` rather than the increased weight sharing. That's just a guess though.

On [this branch](https://github.com/AdamHillier/compute-engine/tree/8x8-arm64-kernel) I have an implementation of an 8x8 kernel that is very close to working (out of the many thousands of kernel tests, it fails 12 of them - specifically it fails for large inputs combined with valid padding combined with large stride combined with large dilation). I haven't been able to successfully debug the issue, and I think @Tombana also had weird specific test failures when he tried to get an 8-column kernel working, so there might be some weird underlying issue. In any case, I'm not going to pursue 8x8 any further any time soon.

arashb:
> ### Prospects for an 8x8 kernel
> If we don't use `RUY_OPT_MAX_STREAMING`, then there are enough registers to be able to implement an 8x8 kernel. I have experimented with disabling the `RUY_OPT_MAX_STREAMING` part of this kernel and haven't noticed any significant benchmarked performance difference, so I don't think that would be a problem.

I observed the same performance behaviour, utilizing the entire register space did not yield noticeable performance differences (it decreases the iteration over depth and as a result number of kernel calls but this is not significant)

Regarding the 8x8 kernels, I also think it will not also yield signifant improvement (for the same reason as stated above for increasing usage of register space) However, if you are still interested to get it working, I have a 8x8 kernel working [here](https://github.com/larq/compute-engine/blob/c6d488a424f02aa97b4fdfd727d26e9033afe50b/larq_compute_engine/tflite/cc/kernels/bgemm_kernels_arm64.h#L45) (with a completely different memory layout and instructions). You might have a bug in corner case where the destination matrix does not fit entirely into 8x8 sub-matrices. You can check against my branch to debug that.

AdamHillier:
We previously had a separate 4x4 kernel for uint32 inputs (that didn't, for example, implement `RUY_OPT_MAX_STREAMING`). As far as I can tell, there's no reason not to write the optimised kernels to just accept uint64, i.e. get rid of the separate uint32 kernel, because we can still use the uint64 kernels with uint32 inputs (by pretending that the inputs are uint64: halving the LHS/RHS stride and the depth in the kernel params).

That is what I have implemented in ab533ea. All the big tests, in particular the `OPT_32` tests, still pass.

The benefit of doing this is that we remove a bunch of code and reduce binary size. Before this PR there were two Arm64 kernels: one 4x4 uint32 kernel and one 4x4 uint64 kernel. After this PR there will still be two Arm64 kernels: one 8x4 uint64 kernel and one 4x4 uint64 kernel. But both will be compatible with uint32 input as well. It also means that using 32-bit bitpacked input still allows us to take advantage of the fastest and most optimised kernels, which is important e.g. for reading bitpacked input directly (which is uint32).

The downside is that ab533ea has made the diff less readable so I'm very happy to revert it and open another PR to discuss those changes.

Tombana:
> That is what I have implemented in [ab533ea](https://github.com/larq/compute-engine/commit/ab533eab50719f76d09e566cb69b977bbcf32c79). All the big tests, in particular the `OPT_32` tests, still pass.

That sounds great! I think i've measured performance differences in the past between the '32-bit' and '64-bit' kernel, even though they should be the same (just some parameters doubled/halved). Could it be that the '64-bit' kernel causes better memory alignment by requesting blocks that are double the size (but half as many) and therefore a speedup? I guess that's the only reason I can think of for having this distinction at all.

AdamHillier:
> That sounds great! I think i've measured performance differences in the past between the '32-bit' and '64-bit' kernel, even though they should be the same (just some parameters doubled/halved). Could it be that the '64-bit' kernel causes better memory alignment by requesting blocks that are double the size (but half as many) and therefore a speedup? I guess that's the only reason I can think of for having this distinction at all.

Ah it's quite possible that alignment might make a difference; the only other concrete difference I can think of is that the uint32 kernels previously didn't have `RUY_OPT_MAX_STREAMING`, though as discussed above I don't think that makes a difference really. Also I guess the actual bitpacking step would be quicker for uint64?

Tombana:
> Ah it's quite possible that alignment might make a difference; the only other concrete difference I can think of is that the uint32 kernels previously didn't have `RUY_OPT_MAX_STREAMING`, though as discussed above I don't think that makes a difference really. Also I guess the actual bitpacking step would be quicker for uint64?

Ah yes the bitpacking is only optimized for uint64. I also didn't realize we didn't have the `RUY_OPT_MAX_STREAMING` in the `uint32` kernel. It might be good to benchmark this stuff, with the bitpacking temporarily commented out.

I'm wondering if it makes sense to keep the 32-bit code path at all. There might be some situations where the 32-bit bitpacking path is faster, when the number of input channels is less than 32, so there's less bitpacking-padding overhead. However I think the assembly kernels always pad it to 128-bits (in the "ruy prepacking step"), so maybe its never faster?

AdamHillier:
> I'm wondering if it makes sense to keep the 32-bit code path at all. There might be some situations where the 32-bit bitpacking path is faster, when the number of input channels is less than 32, so there's less bitpacking-padding overhead. However I think the assembly kernels always pad it to 128-bits (in the "ruy prepacking step"), so maybe its never faster?

Yes I've been thinking about this too. I think the assembly kernels pad to 128-bits only after im2col, so if we have a 3x3 kernel operating on 24 or 32 input channels, for example, and do bitpacking before im2col then there would be a performance gain to be had.

Also, I was under the impression that if/when we implement writing bitpacked activations for the assembly kernels we would do it with 32-bit bitpacked activations as in the reference kernel? Not 100% sure about that though.

Tombana:
> Yes I've been thinking about this too. I think the assembly kernels pad to 128-bits only after im2col, so if we have a 3x3 kernel operating on 24 or 32 input channels, for example, and do bitpacking before im2col then there would be a performance gain to be had.

Good point, there's im2col inbetween indeed.

> Also, I was under the impression that if/when we implement writing bitpacked activations for the assembly kernels we would do it with 32-bit bitpacked activations as in the reference kernel? Not 100% sure about that though.

I'm not sure either. Can we use a destination matrix with \#rows padded to multiples of 64 even when the LHS-\#rows does not agree with that? It might get a bit hacky at some point, but i guess it already is when writing bitpacked activations.

AdamHillier:
> I'm not sure either. Can we use a destination matrix with #rows padded to multiples of 64 even when the LHS-#rows does not agree with that? It might get a bit hacky at some point, but i guess it already is when writing bitpacked activations.

I think we can, because we had to do it to get the CPP kernel working in the first place. As far as I can tell, when writing to the destination we are writing directly to the output tensor that we allocate in `Prepare()`.

arashb:
> We previously had a separate 4x4 kernel for uint32 inputs (that didn't, for example, implement `RUY_OPT_MAX_STREAMING`). As far as I can tell, there's no reason not to write the optimised kernels to just accept uint64, i.e. get rid of the separate uint32 kernel, because we can still use the uint64 kernels with uint32 inputs (by pretending that the inputs are uint64: halving the LHS/RHS stride and the depth in the kernel params).

I think changing the 64-bit kernel to be used for both 32-bit packing and 64-bit packing needs to be in a separate PR, so we bisect it if it causes any bugs later. Regarding, the 32-bit kernels we have them in repo, in case we start to optimize for armv7-a and I think we should keep them.

AdamHillier:
> > We previously had a separate 4x4 kernel for uint32 inputs (that didn't, for example, implement `RUY_OPT_MAX_STREAMING`). As far as I can tell, there's no reason not to write the optimised kernels to just accept uint64, i.e. get rid of the separate uint32 kernel, because we can still use the uint64 kernels with uint32 inputs (by pretending that the inputs are uint64: halving the LHS/RHS stride and the depth in the kernel params).
> 
> I think changing the 64-bit kernel to be used for both 32-bit packing and 64-bit packing needs to be in a separate PR, so we can git bisect it if it causes any bugs later. Regarding, the 32-bit kernels we have them in repo, in case we start to optimize for armv7-a and I think we should keep them.

Makes sense, I'll revert the last commit and open a new PR :+1:

larq_compute_engine/core/bgemm_kernels_arm64.h
```diff
@@ -904,4 +904,584 @@ void BinaryKernelNeonOutOfOrder64BP4x4(
...
+      // them anymore in the rest of the work on the current block. We do it in
+      // parallel with the back-transform shift.
+
+      // Load backtransform add (also in parallel with applying the
+      // back-transform shift).
+      "ldr w1, [%[params], #" RUY_STR(RUY_OFFSET_BACKTRANSFORM_ADD) "]\n"
+
+      // Perform the back-transformation shift with 'unsigned shift left long'
+      // instructions; this performs the necessary left shift and extends the
+      // result, so we get a int16 -> int32 transformation for free.
```
Tombana, line 1255: Awesome

PR #351, opened 2020-05-11
Unify Arm64 kernels, independent of bitpacking bitwidth

AdamHillier:
See comment [here](https://github.com/larq/compute-engine/pull/350#issuecomment-626310229).

~~This depends on #350, so will be a draft PR until that is merged.~~

## What do these changes do?

At the moment there are three optimised assembly kernels for Aarch64: a 4x4 uint32 kernel, a 4x4 uint64 kernel, and an 8x4 uint64 kernel.

I can't see a reason to maintain a distinction between uint32 input and uint64 input at a kernel level, because a kernel that's written to accept uint64 input can accept uint32 input as long as some kernel parameters are adjusted (e.g. kernel depth halved) and the Ruy LHS/RHS packing is done correctly.

This PR removes the separate uint32 4x4 kernel, and makes 8x4 kernel and the remaining 4x4 kernels usable with either uint64 or uint32 input. Whereas there were previously three Aarch64 kernel code-paths and three kernel registrations, now there are two Aarch64 kernel code paths (8x4 and 4x4) and four kernel registrations (uint64 and uint32 for each of the code paths).

The main benefit is reduced code and binary size. However, this will also speed up binary convolutions which use uint32 input, as they were previously unable to use a faster 8x4 kernel.

## How Has This Been Tested?

CI tests. The 'big' tests that don't run on CI have be run locally (and pass).

## Benchmark Results

N/A.

## Related issue number

N/A.

### DIFF
--- a/larq_compute_engine/core/bgemm_kernels_common.h
+++ b/larq_compute_engine/core/bgemm_kernels_common.h
@@ -104,4 +104,54 @@ inline void MakeBinaryKernelParams(
   RUY_DCHECK_LT(params->last_col, params->dst_cols);
 }
 
+// A specialised template for the case when the LHS and RHS are uint32 bitpacked
+// but we're using a kernel designed for uint64 bitpacked inputs.
+template <int LhsCols, int RhsCols, typename AccumScalar>
+inline void MakeBinaryKernelParams(
+    const PackedMatrix<std::uint32_t>& lhs,
+    const PackedMatrix<std::uint32_t>& rhs,
+    const BinaryMulParams<AccumScalar, float>& spec, int start_row,
+    int start_col, int end_row, int end_col, Matrix<float>* dst,
+    BinaryKernelParams<LhsCols, RhsCols, std::uint64_t>* params) {
+  const int depth = lhs.layout.rows;
+  RUY_DCHECK_EQ(start_row % LhsCols, 0);
+  RUY_DCHECK_EQ(start_col % RhsCols, 0);
+  RUY_DCHECK_EQ(end_row % LhsCols, 0);
+  RUY_DCHECK_EQ(end_col % RhsCols, 0);
+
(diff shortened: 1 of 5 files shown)


larq_compute_engine/core/bgemm_kernels_common.h
```diff
@@ -99,4 +99,54 @@ inline void MakeBinaryKernelParams(
...
+  params->flags = flags;
+  params->start_row = start_row;
+  params->start_col = start_col;
+  params->last_row = end_row - LhsCols;
+  params->last_col = end_col - RhsCols;
+  params->lhs_stride = sizeof(std::uint32_t) * lhs.layout.stride;
+  params->rhs_stride = sizeof(std::uint32_t) * rhs.layout.stride;
+  params->dst_stride = sizeof(float) * dst->layout.stride;
+  // We halve the depth to pretend that the input data is uint64.
+  RUY_DCHECK_EQ(depth % 2, 0);
```
Tombana, line 141: It's been some time since I've looked at this, but is `depth` measured in number of `TBitpacked` elements, or is it measured in number of "blocks" where a "block" is what is defined by `using LhsLayout = FixedKernelLayout<Order::kColMajor, 2, 4>;` ? I guess the first one.

What happens if the depth is odd? Does RUY's prepacking somehow make sure that doesn't happen because it knows about that block size thats expected by the kernel?

AdamHillier: It's the number of `TBitpacked` elements, and I believe that the Ruy pre-packing makes sure that that's not an issue by using padding (it knows the depth at compile-time as it's reported in the `FixedKernelLayout`, as you point out).

lgeiger:
Looks like the merge of #346 introduced some merge conflicts here.

PR #352, opened 2020-05-11
Add experimental support for bitpacked output tensors to the converter.

AdamHillier:
## What do these changes do?

This PR adds support to the LCE converter for Int32 output tensors of binary convolutions.

This can happen when we have two binary convolutions, `A` followed by `B` (with nothing inbetween), where `B` is the only consumer of the output of `A`. Then, instead of `A` writing float values and `B` having an initial bitpacking step, we can write bitpacked values from `A` directly to be consumed by `B`. This will significantly reduce intermediate activation memory usage.

To-do:
- [x] Gatekeep this feature behind a Python flag `experimental_enable_bitpacked_activations`.

## How Has This Been Tested?

CI.

## Benchmark Results

N/A.

## Related issue number

N/A.


AdamHillier:
This should now be good to go and ready for review.

larq_compute_engine/mlir/tf_tfl_passes.h
```diff
@@ -8,7 +8,9 @@
 namespace tensorflow {
 
 // Add the TF to TFLite passes into a pass_manager.
-void AddTFToLCETFLConversionPasses(mlir::OpPassManager* pass_manager);
+void AddTFToLCETFLConversionPasses(
+    mlir::OpPassManager* pass_manager,
+    bool experimental_enable_bitpacked_activations = true);
```
lgeiger, line 13: How do you feel about defaulting to `false` which would be consistent to the Python code path, or not setting a default at all?
```suggestion
    bool experimental_enable_bitpacked_activations = false);
```

For the MLIR tests, having the default to true in the `Pass` constructor should be enough.

larq_compute_engine/mlir/transforms/optimize.cc
```diff
@@ -58,12 +68,84 @@ DenseElementsAttr Bitpack(PatternRewriter& builder, Attribute x) {
...
+
+    auto bconv_input_op =
+        dyn_cast_or_null<TF::LceBconv2dOp>(bconv_input.getDefiningOp());
+    if (!bconv_input_op) return matchFailure();
+
+    if (bconv_input_op.write_bitpacked_output() &&
+        bconv_op.read_bitpacked_input())
+      return matchFailure();
+
+    if (bconv_input_op.padding() == "SAME" && bconv_input_op.pad_values() == 0)
```
lgeiger, line 87:
```suggestion
    if (bconv_input_op.padding() == "SAME" && bconv_input_op.pad_values() != 1)
```

larq_compute_engine/mlir/transforms/optimize.cc
```diff
@@ -58,12 +68,84 @@ DenseElementsAttr Bitpack(PatternRewriter& builder, Attribute x) {
...
+
+    const auto channels = inner_tensor_shape[3];
+    const auto packed_channels = (channels + bitwidth - 1) / bitwidth;
+
+    RankedTensorType inner_tensor_type =
+        RankedTensorType::get({inner_tensor_shape[0], inner_tensor_shape[1],
+                               inner_tensor_shape[2], packed_channels},
+                              rewriter.getIntegerType(bitwidth));
+
+    rewriter.replaceOpWithNewOp<TF::LceBconv2dOp>(
```
lgeiger, line 105: Do you use `replaceOpWithNewOp` since you need to replace two ops at once and `replaceOp` won't work?

larq_compute_engine/mlir/transforms/optimize.cc
```diff
@@ -58,12 +68,84 @@ DenseElementsAttr Bitpack(PatternRewriter& builder, Attribute x) {
...
+        bconv_input_op, inner_tensor_type, bconv_input_op.input(),
+        bconv_input_op.filter(), bconv_input_op.post_activation_multiplier(),
+        bconv_input_op.post_activation_bias(),
+        rewriter.getIntegerAttr(rewriter.getIntegerType(32),
+                                bconv_input_op.channels_in()),
+        bconv_input_op.strides(),
+        rewriter.getStringAttr(bconv_input_op.padding()),
```
lgeiger, line 112: Doesn't `bconv_input_op.padding()` work? Why do we need to use `rewriter.getStringAttr`here?

larq_compute_engine/mlir/transforms/prepare_patterns.td
```diff
@@ -52,12 +54,15 @@ def : Pat<(TF_Conv2DOp: $output
             (TF_TransposeOp (TF_DivOp (ConstantOp $filter), (ConstantOp (GetScaleVector $filter))), (ConstantOp ConstantAttr<I32VectorElementsAttr<4>, "{3, 0, 1, 2}">)),
             (ConstantOp (GetScaleVector $filter)),
             (ConstantOp (GetConstantVector<"0.0f"> $filter)),
+            ConstantAttr<I32Attr, "-1">,
```
lgeiger, line 57: Do you think it would be nicer to directly assign the shape here in tablegen using something like [`this`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/mlir/lite/transforms/optimize_patterns.td#L396) and [`this`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/mlir/lite/transforms/optimize.cc#L193-L207)?

If not, feel free to ignore this comment :)

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -199,6 +199,14 @@ TfLiteStatus Prepare(KernelType kernel_type,
   TF_LITE_ENSURE_EQ(context, NumDimensions(post_activation_multiplier), 1);
   TF_LITE_ENSURE_EQ(context, NumDimensions(post_activation_bias), 1);
 
+  // Read the input dimensions. TF Lite has the same input format as TensorFlow:
+  // (B, H, W, Ci).
+  conv_params->batch = input->dims->data[0];
+  conv_params->input_height = input->dims->data[1];
+  conv_params->input_width = input->dims->data[2];
+  if (!conv_params->read_bitpacked_input)
+    conv_params->channels_in = input->dims->data[3];
```
lgeiger, line 208: Just personal preference, feel free to ignore (I feel like I am betraying Python with this suggestion)
```suggestion
  if (!conv_params->read_bitpacked_input) {
    conv_params->channels_in = input->dims->data[3];
  }
```

larq_compute_engine/mlir/ir/lce_ops.td
```diff
@@ -50,12 +50,15 @@ TODO
     TensorOf<[F32]>:$post_activation_multiplier,
     TensorOf<[F32]>:$post_activation_bias,
 
+    DefaultValuedAttr<I32Attr, "-1">:$channels_in,
```
lgeiger, line 53: I am not a fan of this argument, but I guess for now this is the easiest way to preserve the shape :+1:

larq_compute_engine/mlir/tests/optimize.mlir
```diff
@@ -137,21 +135,66 @@ func @do_not_fuse_relu_into_bconv2d_no_post_activation_bias(%arg0: tensor<256x32
...
+  // CHECK: %0 = "tf.LceBconv2d"(%arg0, %cst, %arg1, %arg2) {activation = "NONE", channels_in = -1 : i32, data_format = "NHWC", dilations = [1, 1, 1, 1], filter_format = "OHWI_PACKED", pad_values = 0 : i32, padding = "VALID", read_bitpacked_input = false, strides = [1, 1, 1, 1], write_bitpacked_output = false} : (tensor<256x32x32x3xf32>, tensor<16x3x3x1xi32>, tensor<16xf32>, tensor<16xf32>) -> tensor<256x30x30x16xf32>
   // CHECK-NEXT: return %0
 }
+
+// CHECK-LABEL: @bitpack_activations_between_two_bconv2ds_valid_padding
+func @bitpack_activations_between_two_bconv2ds_valid_padding(%arg0: tensor<256x32x32x3xf32>, %arg1: tensor<65x3x3x3xf32>, %arg2: tensor<8x3x3x65xf32>) -> tensor<256x28x28x8xf32> {
+  %post_activation_multiplier_1 = constant dense<3.14> : tensor<65xf32>
+  %post_activation_bias_1 = constant dense<1.5> : tensor<65xf32>
+  %post_activation_multiplier_2 = constant dense<[1.0, -1.0, 1.0, -1.0, 1.0, -1.0, 1.0, -1.0]> : tensor<8xf32>
+  %post_activation_bias_2 = constant dense<-1.5> : tensor<8xf32>
```
lgeiger, line 162: Since we don't actually check any of the values, I think it would be more readable to move those into arguments of the test function `%arg{3,4,5,6}`

larq_compute_engine/mlir/tests/optimize.mlir
```diff
@@ -137,21 +135,66 @@ func @do_not_fuse_relu_into_bconv2d_no_post_activation_bias(%arg0: tensor<256x32
...
+  %0 = "tf.LceBconv2d"(%arg0, %arg1, %post_activation_multiplier_1, %post_activation_bias_1) {activation = "NONE", channels_in = -1 : i32, data_format = "NHWC", dilations = [1, 1, 1, 1], filter_format = "OHWI", pad_values = 1 : i32, padding = "SAME", read_bitpacked_input = false, strides = [1, 1, 1, 1], write_bitpacked_output = false} : (tensor<256x32x32x3xf32>, tensor<65x3x3x3xf32>, tensor<65xf32>, tensor<65xf32>) -> tensor<256x32x32x65xf32>
+  %1 = "tf.LceBconv2d"(%0, %arg2, %post_activation_multiplier_2, %post_activation_bias_2) {activation = "NONE", channels_in = -1 : i32, data_format = "NHWC", dilations = [1, 1, 1, 1], filter_format = "OHWI", pad_values = 0 : i32, padding = "VALID", read_bitpacked_input = false, strides = [1, 1, 1, 1], write_bitpacked_output = false} : (tensor<256x32x32x65xf32>, tensor<8x3x3x65xf32>, tensor<8xf32>, tensor<8xf32>) -> tensor<256x30x30x8xf32>
+  return %1 : tensor<256x30x30x8xf32>
```
lgeiger, line 180: I think it would be good to add a test that checks that we don't set `write_bitpacked_output` if the output has multiple users (like in a skip connection). I think this can be easily simulated by returning `%0` from the function as well.

larq_compute_engine/mlir/ir/lce_ops.td
AdamHillier: Yes at some point we can have some nice bitpacked tensor type that preseves this, but it's the only easy way for now.

larq_compute_engine/mlir/transforms/optimize.cc
AdamHillier: I think this is correct because we're checking in order to return `matchFailure`.

larq_compute_engine/mlir/transforms/optimize.cc
AdamHillier: I'm not sure specifically why it's necessary, but it doesn't compile without it.

larq_compute_engine/mlir/transforms/optimize.cc
lgeiger: I agree, this is correct. But for the case that `bconv_input_op.pad_values()` is not `0` or `1` (which should never happen if the rest of the code does the right thing) it would pass as well. So checking agains ` != 1` should be more conservative.

larq_compute_engine/mlir/transforms/optimize.cc
lgeiger: Hmm OK, let's leave it like this then :+1:

larq_compute_engine/mlir/transforms/optimize.cc
AdamHillier: Ah sorry I misread the suggestion :)

larq_compute_engine/mlir/transforms/optimize.cc
AdamHillier: Ah I wasn't actually aware of `replaceOp` because I copied the boilerplate for this pass from a TF example. I can't find a clear definition of `replaceOp` though and it seems to only accept two arguments; do you know how it's meant to be used?

lgeiger: :+1: I'll check if there is a qay to simplify this.

lgeiger: I wonder if the two `replaceOpWithNewOp` statements can just be replaced by:
```cpp
bconv_input_op.setAttr("write_bitpacked_output", rewriter.getBoolAttr(true))
bconv_op.setAttr("read_bitpacked_input", rewriter.getBoolAttr(true))
```

Tombana:
Now that we set the tensortype to INT32, there's no longer the need for the attributes. Do we want to remove the attributes already in this PR or do we want to add them now and then delete them in a next PR? (since it'd also require some changes to the `Init()` and `Prepare()` functions). I'm neutral on this.

AdamHillier:
> Now that we set the tensortype to INT32, there's no longer the need for the attributes. Do we want to remove the attributes already in this PR or do we want to add them now and then delete them in a next PR? (since it'd also require some changes to the `Init()` and `Prepare()` functions). I'm neutral on this.

Ah that's a very good point, I hadn't thought about that. I'm happy to remove them in this PR :+1:

larq_compute_engine/mlir/tests/optimize.mlir
AdamHillier: That's a good idea, should be done in e905a83.

larq_compute_engine/mlir/tests/optimize.mlir
AdamHillier: Done in e905a83.

larq_compute_engine/mlir/transforms/optimize.cc
AdamHillier: Now that we're removing the two boolean flags entirely and relying on the types, I suspect that the `replaceOpWithNewOp` will be required now as I doubt you can change the output type? I could be wrong though.

larq_compute_engine/mlir/tests/optimize.mlir
```diff
@@ -137,21 +135,65 @@ func @do_not_fuse_relu_into_bconv2d_no_post_activation_bias(%arg0: tensor<256x32
...
+  %1 = "tf.LceBconv2d"(%0, %arg4, %arg5, %arg6) {activation = "NONE", channels_in = -1 : i32, data_format = "NHWC", dilations = [1, 1, 1, 1], filter_format = "OHWI", pad_values = 0 : i32, padding = "VALID", strides = [1, 1, 1, 1]} : (tensor<256x32x32x65xf32>, tensor<8x3x3x65xf32>, tensor<8xf32>, tensor<8xf32>) -> tensor<256x30x30x8xf32>
+  return %1 : tensor<256x30x30x8xf32>
+
+  // CHECK: %0 = "tf.LceBconv2d"(%arg0, %arg1, %arg2, %arg3) {activation = "NONE", channels_in = -1 : i32, data_format = "NHWC", dilations = [1, 1, 1, 1], filter_format = "OHWI", pad_values = 0 : i32, padding = "SAME", strides = [1, 1, 1, 1]}
+  // CHECK-NEXT: %1 = "tf.LceBconv2d"(%0, %arg4, %arg5, %arg6) {activation = "NONE", channels_in = -1 : i32, data_format = "NHWC", dilations = [1, 1, 1, 1], filter_format = "OHWI", pad_values = 0 : i32, padding = "VALID", strides = [1, 1, 1, 1]}
```
lgeiger, line 186: I think it would be good to check the tensor output types to make sure that the activations are not bitpacked.

larq_compute_engine/mlir/tests/optimize.mlir
```diff
@@ -137,21 +135,65 @@ func @do_not_fuse_relu_into_bconv2d_no_post_activation_bias(%arg0: tensor<256x32
...
+  %1 = "tf.LceBconv2d"(%0, %arg4, %arg5, %arg6) {activation = "NONE", channels_in = -1 : i32, data_format = "NHWC", dilations = [1, 1, 1, 1], filter_format = "OHWI", pad_values = 0 : i32, padding = "VALID", strides = [1, 1, 1, 1]} : (tensor<256x32x32x65xf32>, tensor<8x3x3x65xf32>, tensor<8xf32>, tensor<8xf32>) -> tensor<256x30x30x8xf32>
+  return %0, %1: tensor<256x32x32x65xf32>, tensor<256x30x30x8xf32>
+
+  // CHECK: %0 = "tf.LceBconv2d"(%arg0, %arg1, %arg2, %arg3) {activation = "NONE", channels_in = -1 : i32, data_format = "NHWC", dilations = [1, 1, 1, 1], filter_format = "OHWI", pad_values = 1 : i32, padding = "SAME", strides = [1, 1, 1, 1]}
+  // CHECK-NEXT: %1 = "tf.LceBconv2d"(%0, %arg4, %arg5, %arg6) {activation = "NONE", channels_in = -1 : i32, data_format = "NHWC", dilations = [1, 1, 1, 1], filter_format = "OHWI", pad_values = 0 : i32, padding = "VALID", strides = [1, 1, 1, 1]}
```
lgeiger, line 197: See above

larq_compute_engine/mlir/transforms/prepare_patterns.td
lgeiger: @AdamHillier How do you feel about always setting `channel_in`? Another way would be to set the default to `None` wihch might be cleaner.

larq_compute_engine/mlir/tests/optimize.mlir
AdamHillier: Ah yeah sorry hadn't seen that I'd missed those out. Done in 09ea77a.

larq_compute_engine/mlir/tests/optimize.mlir
AdamHillier: See above.

larq_compute_engine/mlir/transforms/prepare_patterns.td
AdamHillier: Done in 09ea77a.

larq_compute_engine/mlir/transforms/optimize.cc
```diff
@@ -58,12 +68,80 @@ DenseElementsAttr Bitpack(PatternRewriter& builder, Attribute x) {
...
+    if (!bconv_input.hasOneUse()) return matchFailure();
+
+    auto bconv_input_op =
+        dyn_cast_or_null<TF::LceBconv2dOp>(bconv_input.getDefiningOp());
+    if (!bconv_input_op) return matchFailure();
+
+    const auto inner_tensor_type = bconv_input_op.getType().cast<ShapedType>();
+
+    // We can only apply this transformation if the inner tensor type is F32.
+    if (!inner_tensor_type.getElementType().isF32()) return matchFailure();
```
Tombana, line 86: Just to check if I understood it right:
This comment is true because this pass is happening *before* any int8-quantization related passes?

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -199,6 +170,35 @@ TfLiteStatus Prepare(KernelType kernel_type,
...
+    return kTfLiteError;
+  }
+
+  // Read the input dimensions. TF Lite has the same input format as TensorFlow:
+  // (B, H, W, Ci).
+  conv_params->batch = input->dims->data[0];
+  conv_params->input_height = input->dims->data[1];
+  conv_params->input_width = input->dims->data[2];
+  if (!conv_params->read_bitpacked_input) {
+    conv_params->channels_in = input->dims->data[3];
```
Tombana, line 190: I would opt for always assuming `channels_in` is set now, and giving an error if it isn't. This will also make sure that "the kernel version matches the converter version" so to speak. I guess if `channels_in` isn't set then we could also throw an error in `Init` already.
```suggestion
    TF_LITE_ENSURE_EQ(context, conv_params->channels_in, input->dims->data[3]);
```

larq_compute_engine/mlir/transforms/optimize.cc
AdamHillier: Well that's a very good question, to which I don't know the answer. @lgeiger do Int8 passes happen after LCE optimise passes?

larq_compute_engine/tflite/kernels/bconv2d.cc
AdamHillier: Out of interest, do we have a policy on which kernel versions support which converter versions?

larq_compute_engine/mlir/transforms/optimize.cc
Tombana: I guess they *should* happen after, since the int8 stats simply don't make sense when the layers aren't fused. (batchnorm, etc)

larq_compute_engine/tflite/kernels/bconv2d.cc
Tombana: I guess not. It might be a good idea to start doing that. I think we are planning to change the attributes to be more in line with the tflite conv, (i.e. strings -> enums). When we have that, maybe we could label that as "version 1" and then use tflite's op versioning system whenever we have a breaking change from that point.

larq_compute_engine/mlir/transforms/optimize.cc
AdamHillier: In any case, I think 47e1084 solves the issue.

larq_compute_engine/mlir/transforms/optimize.cc
```diff
@@ -58,12 +68,81 @@ DenseElementsAttr Bitpack(PatternRewriter& builder, Attribute x) {
...
+        dyn_cast_or_null<TF::LceBconv2dOp>(bconv_input.getDefiningOp());
+    if (!bconv_input_op) return matchFailure();
+
+    const auto inner_tensor_type = bconv_input_op.getType().cast<ShapedType>();
+
+    // We use 32-bit bitpacking.
+    constexpr int bitwidth = 32;
+
+    // We can only apply this transformation if the inner tensor type is F32.
+    if (inner_tensor_type.getElementType().isInteger(bitwidth))
```
lgeiger, line 89: Do we need to check if `bconv_op.channels_in()` is set, or will the previous pattern already catch that?

larq_compute_engine/mlir/ir/lce_ops.td
```diff
@@ -50,6 +50,7 @@ TODO
     TensorOf<[F32]>:$post_activation_multiplier,
     TensorOf<[F32]>:$post_activation_bias,
 
+    DefaultValuedAttr<I32Attr, "0">:$channels_in,
```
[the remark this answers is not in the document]
AdamHillier: Sounds good, I'd prefer to require it.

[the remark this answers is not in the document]
Tombana: I also prefer to require it.

larq_compute_engine/mlir/transforms/optimize.cc
AdamHillier: I don't think this is an issue if we're requiring the attribute to be set?

larq_compute_engine/mlir/ir/lce_ops.td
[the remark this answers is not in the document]
AdamHillier: Done in 34cb96f.

larq_compute_engine/mlir/transforms/optimize.cc
lgeiger: Yeah I think you're right. I was checking the pattern and it will always be set in the prepare pass.

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -199,6 +170,35 @@ TfLiteStatus Prepare(KernelType kernel_type,
...
+  }
+
+  // Read the input dimensions. TF Lite has the same input format as TensorFlow:
+  // (B, H, W, Ci).
+  conv_params->batch = input->dims->data[0];
+  conv_params->input_height = input->dims->data[1];
+  conv_params->input_width = input->dims->data[2];
+  if (!conv_params->read_bitpacked_input) {
+    TF_LITE_ENSURE_EQ(context, conv_params->channels_in, input->dims->data[3]);
+  } else if (conv_params->channels_in == 0) {
```
lgeiger, line 191: Since `0` is not the default argument in the converter anymore, I think we should check against null, or remove this check.

arashb: channels in can never be `0` can it?

AdamHillier: 0 is the default argument in the `conv_params` constructor, so the value will be 0 if it's not otherwise set (in `Init` we set the value, but only if it's non-null in the flatbuffer). The only reason this check would be relevent would be if you tried to run with an old model that was converted with an old converter and so didn't have the `channels_in` attribute set, but that's still probably something we should catch.

AdamHillier: On a related note, I don't think we have a proper way of versioning the ops/the converter?

lgeiger: > On a related note, I don't think we have a proper way of versioning the ops/the converter?

Currently not, though builtin TFLite ops do. I think we should investigate if we can get some sort of versioning at some point.

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -199,6 +170,35 @@ TfLiteStatus Prepare(KernelType kernel_type,
...
+    return kTfLiteError;
+  }
+
+  // Read the input dimensions. TF Lite has the same input format as TensorFlow:
+  // (B, H, W, Ci).
+  conv_params->batch = input->dims->data[0];
+  conv_params->input_height = input->dims->data[1];
+  conv_params->input_width = input->dims->data[2];
+  if (!conv_params->read_bitpacked_input) {
+    TF_LITE_ENSURE_EQ(context, conv_params->channels_in, input->dims->data[3]);
```
lgeiger, line 190: I just ran a sanity check with the benchmarking binary and this check fails on a model that was converted with an old version of the converter, should we remove this check to stay backwards compatible?
`larq_compute_engine/tflite/kernels/bconv2d.cc:190 conv_params->channels_in != input->dims->data[3] (0 != 64)`

larq_compute_engine/tflite/kernels/bconv2d.cc
Tombana: When registering an op you can provide a version, so from the runtime side thats very easy to do. I'm not sure how easy it is from the mlir/flatbuffer side.

lgeiger: > I'm not sure how easy it is from the mlir/flatbuffer side.

I can take a look at it later this week. TF 2.2 added support for this in https://github.com/tensorflow/tensorflow/commit/42a4ae6cbef9505fd7f7f7f7d17e869da75b9583

larq_compute_engine/tflite/kernels/bconv2d.cc
AdamHillier: I don't have a strong opinion about this

PR #356, opened 2020-05-12
Add support for int8 quantization aware training 

### LINKED ISSUE #344: 8-bit quantization for non-binary layers
Is it possible to use the converter in a way that would enable using tflite's default 8-bit post training quantization on layers and activations that were not quantized by larq? (Such as first, last conv layers).

It seems that currently the public conversion API externalizes only `convert_keras_model` and is not customizable by the user (as opposed to tflite's default converter)

lgeiger:
## What do these changes do?
This PR adds support for quantization aware training. It PR is built on top of #349 so only the last two commits are relevant for review.

## How Has This Been Tested?
CI and manual testing by @AdamHillier and me.

## Related issue number
Adresses #344

### DIFF
--- a/larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
+++ b/larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
@@ -40,7 +40,7 @@ pybind11::bytes ConvertGraphDefToTFLiteFlatBuffer(
     const std::vector<string>& input_arrays,
     const std::vector<string>& input_dtypes,
     const std::vector<std::vector<int>>& input_shapes,
-    const std::vector<string>& output_arrays, const bool& should_quantize,
+    const std::vector<string>& output_arrays, const bool should_quantize,
     const bool experimental_enable_bitpacked_activations) {
   GraphDef graphdef;
   if (!tensorflow::LoadProtoFromBuffer(std::string(graphdef_bytes), &graphdef)
--- a/larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
+++ b/larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
@@ -11,6 +11,7 @@
 #include "pybind11/pytypes.h"
 #include "pybind11/stl.h"
 #include "tensorflow/compiler/mlir/lite/flatbuffer_export.h"
+#include "tensorflow/compiler/mlir/lite/quantization/quantization_config.h"
 #include "tensorflow/compiler/mlir/lite/transforms/passes.h"
 #include "tensorflow/compiler/mlir/op_or_arg_name_mapper.h"
 #include "tensorflow/compiler/mlir/tensorflow/translate/import_model.h"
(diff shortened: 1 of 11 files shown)


lgeiger:
> LGTM (after including the test we discussed in the call)

Great! I extended the tests in 43d303a139859eb7a8853ceed44e268d63802760

Tombana:
> Great! I extended the tests in [43d303a](https://github.com/larq/compute-engine/commit/43d303a139859eb7a8853ceed44e268d63802760)

Is it correct that you changed the original test, and not made a copy? Now its not doing the original test anymore.. Other than that the test looks good to me.

lgeiger:
> Is it correct that you changed the original test, and not made a copy? Now its not doing the original test anymore.

The original test is a subset of the new one, so I think it is fine to have only a single one here.

larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
```diff
@@ -38,7 +40,7 @@ pybind11::bytes ConvertGraphDefToTFLiteFlatBuffer(
     const std::vector<string>& input_arrays,
     const std::vector<string>& input_dtypes,
     const std::vector<std::vector<int>>& input_shapes,
-    const std::vector<string>& output_arrays,
+    const std::vector<string>& output_arrays, const bool& should_quantize,
```
AdamHillier, line 43: What's the difference between `const bool& foo` and `const bool foo`?

larq_compute_engine/mlir/transforms/quantize.cc
```diff
@@ -0,0 +1,70 @@
...
+
+#include "larq_compute_engine/mlir/transforms/generated_quantize.inc"
+
+void HybridQuantizePass::runOnFunction() {
+  OwningRewritePatternList patterns;
+  auto func = getFunction();
+  auto* ctx = func.getContext();
+  TFL::populateWithGenerated(ctx, &patterns);
+  patterns.insert<TFLHybridQuantization>(ctx, /*enable_numeric_verify=*/false,
+                                         /*error_tolerance=*/5.0,
```
AdamHillier, line 54: What does this error tolerance mean?

larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
Tombana: :+1: For primitive types we can just use `const bool foo`, no need to pass it by reference.

larq_compute_engine/mlir/transforms/quantize.cc
lgeiger: These are just the defaults, passed into the pattern by TensorFlow: https://github.com/tensorflow/tensorflow/blob/686908251a6711212cc7fad6de3d929c6c0c1921/tensorflow/compiler/mlir/lite/transforms/quantize.cc#L37-L55

larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
lgeiger: :+1: Changed it to not pass it by reference.

larq_compute_engine/mlir/transforms/quantize.cc
lgeiger: I guess they are only used if `enable_numeric_verify=true`.

AdamHillier: Ah makes sense, thanks.

PR #357, opened 2020-05-12
Allow to set fake default ranges to enable latency test of int8 models 

lgeiger:
## What do these changes do?
This allows for passing default int8 quantization ranges to the converter to enable latency testing without needing to correctly have fake-quant nodes in the graph.

This currently relies on a fork of TensorFlow to include https://github.com/tensorflow/tensorflow/pull/39171.
#363 still needs some work which would include the relevant commits, but this PR can already be reviewed until we incorporate the Ruy API changes of #363 

## How Has This Been Tested?
CI and manual testing.

### DIFF
--- a/larq_compute_engine/mlir/python/converter.py
+++ b/larq_compute_engine/mlir/python/converter.py
@@ -52,7 +52,7 @@ def _contains_training_quant_op(graph_def):
 def convert_keras_model(
     model: tf.keras.Model,
     *,  # Require remaining arguments to be keyword-only.
-    default_ranges: Optional[Tuple[float, float]] = None,
+    experimental_default_int8_range: Optional[Tuple[float, float]] = None,
     experimental_enable_bitpacked_activations: bool = False,
 ) -> bytes:
     """Converts a Keras model to TFLite flatbuffer.
@@ -66,9 +66,9 @@ def convert_keras_model(
 
     # Arguments
         model: The model to convert.
-        default_ranges: Tuple of integers representing `(min, max)` range values for all
-            arrays without a specified range. Intended for experimenting with
-            quantization via "dummy quantization". (default None)
+        experimental_default_int8_range: Tuple of integers representing `(min, max)`
+            range values for all arrays without a specified range. Intended for
+            experimenting with quantization via "dummy quantization". (default None)
         experimental_enable_bitpacked_activations: Enable an experimental
             converter optimisation that attempts to reduce intermediate
             activation memory usage by bitpacking the activation tensor between
@@ -82,9 +82,9 @@ def convert_keras_model(
             "Graph mode is not supported. Please enable eager execution using "
             "tf.enable_eager_execution() when using TensorFlow 1.x"
         )
-    if default_ranges:
+    if experimental_default_int8_range:
         warnings.warn(
-            "Using `default_ranges` as fallback quantization stats. "
+            "Using `experimental_default_int8_range` as fallback quantization stats. "
             "This should only be used for latency tests."
         )
     func = concrete_function_from_keras_model(model)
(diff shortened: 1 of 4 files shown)


larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
```diff
@@ -80,6 +81,11 @@ pybind11::bytes ConvertGraphDefToTFLiteFlatBuffer(
       // which default to mean=0.0 and std=1.0.
       quant_specs.input_ranges.push_back({-128.0, 127.0});
     }
+    if (!default_ranges.is_none()) {
+      quant_specs.inference_input_type = tensorflow::DT_QINT8;
```
lgeiger, line 85: Setting `inference_input_type=int8` will get rid of the need for Quantize / Dequantize ops at beginning and end of the network.

Tombana: Thats great because they were blowing up the MIAM for some models.

WORKSPACE
```diff
@@ -47,10 +47,10 @@ arm_compiler_configure(
 #    reliable downloads.
 http_archive(
     name = "org_tensorflow",
-    sha256 = "8e86547c8238637cdc895ec53f02adedad4dfa2a7fcfee06dbdeb9f0e3be6168",
-    strip_prefix = "tensorflow-a80d96c6634cd005b3841d462030448f9f551a14",
+    sha256 = "13fd904f7da15b1b90f5016a48e88f0b82a3a2472fc957bf7afbcfbb0c7cc6a7",
+    strip_prefix = "tensorflow-0670e04d5664ee6e1485becd8abe410a74e2f4e7",
     urls = [
-        "https://github.com/tensorflow/tensorflow/archive/a80d96c6634cd005b3841d462030448f9f551a14.tar.gz",
+        "https://github.com/lgeiger/tensorflow/archive/0670e04d5664ee6e1485becd8abe410a74e2f4e7.tar.gz",
```
Tombana, line 53: Is this required or accidentally comitted?

larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
Tombana: So now this is only activated when default ranges are supplied? Should it not be a separate argument to the convert function whether to have those ops?

WORKSPACE
lgeiger: This is required since this PR relies on https://github.com/tensorflow/tensorflow/pull/39171

Upgrading to an upstream version requires #363 to be merged before. There is no rush, so we can wait with merging until #363 is resolved.

Tombana: Ah makes sense. Having default ranges is super useful to have right now, so I'd rather merge this now and then 'fix' the tensorflow dependency in #363 again. @arashb what do you think?

larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
lgeiger: > So now this is only activated when default ranges are supplied?

Correct

> Should it not be a separate argument to the convert function whether to have those ops?

In general yes, though I'd rather wait with implementing this until TF has proper support for this. See https://github.com/tensorflow/tensorflow/issues/38285

AdamHillier:
Blocked on #363

larq_compute_engine/mlir/python/converter.py
```diff
@@ -50,6 +52,7 @@ def _contains_training_quant_op(graph_def):
 def convert_keras_model(
     model: tf.keras.Model,
     *,  # Require remaining arguments to be keyword-only.
+    default_ranges: Optional[Tuple[float, float]] = None,
```
lgeiger, line 55: Internally we ran into https://github.com/tensorflow/tensorflow/issues/40055

@AdamHillier how do you feel about marking `default_ranges` as experimental for now as well?

AdamHillier: Yes, I think we should definitely do that. `experimental_default_int8_range`?

lgeiger: Sounds good. Changed it to `experimental_default_int8_range` in 483781ac4fd2eb194360c7b39ed40747c92ee08b

PR #362, opened 2020-05-13
Update Tensorflow submodule to tag v2.2.0

Tombana:
## What do these changes do?
Probably due to all the rebasing, the tf submodule got reverted to `v2.2.0-rc0`. This PR updates it to `v2.2.0`.


Tombana:
> This will conflict with #349
> I don't have a strong opinion, but #346 will need another rebase if we merge this.

Good point, I didn't think about that. I don't have a strong opinion either, I'll try to review #346 soon.

lgeiger:
Closed via #346

PR #363, opened 2020-05-13
Upgrade TensorFlow to eaacee173897b77cdb6afd22d5e78154177a10f3

lgeiger:
## What do these changes do?
This PR upgrades the TensorFlow dependency in order to include [`eaacee173897b77cdb6afd22d5e78154177a10f3`](https://github.com/tensorflow/tensorflow/commit/eaacee173897b77cdb6afd22d5e78154177a10f3).

Ruy heavily changed their internal API which requires quite a few code changes. A full diff of the changes in ruy can be seen [here](https://github.com/google/ruy/compare/9f53ba413e6fc879236dcaa3e008915973d67a4f...1b313682ef8b8fc8ed08719c610d1c3503b016bf). This is a first stab at trying to adapt our kernels to the API changes.

## How Has This Been Tested?
CI

## Related issue number
This is a required update to enable easy int8 benchmarking in #357

### DIFF
--- a/larq_compute_engine/core/bgemm_impl_ruy.h
+++ b/larq_compute_engine/core/bgemm_impl_ruy.h
@@ -53,18 +53,21 @@ struct BGemmImplUsingRuy {
     rhs.set_data(rhs_data);
     dst.set_data(dst_data);
 
-    ruy::Mat<LhsScalar> internal_lhs = ruy::ToInternal(lhs);
-    ruy::Mat<RhsScalar> internal_rhs = ruy::ToInternal(rhs);
+    // We have to make this a `const` matrix because otherwise gcc will try to
+    // use the non-const versions of `matrix.data()`
+    ruy::Mat<LhsScalar> internal_lhs =
+        ruy::ToInternal((const ruy::Matrix<LhsScalar>)lhs);
+    ruy::Mat<RhsScalar> internal_rhs =
+        ruy::ToInternal((const ruy::Matrix<RhsScalar>)rhs);
     ruy::Mat<DstScalar> internal_dst = ruy::ToInternal(dst);
 
     BinaryMulParams<AccumScalar, DstScalar> mul_params;
     mul_params.output_transform = output_transform;
 
-    constexpr auto BGemmCompiledPaths = ruy::kAllPaths & ~ruy::Path::kReference;
+    constexpr auto BGemmCompiledPaths = ruy::kAllPaths;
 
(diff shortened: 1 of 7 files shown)


larq_compute_engine/core/bgemm_impl_ruy.h
```diff
@@ -99,41 +97,16 @@ struct BGemmImplUsingRuy {
...
-                                    rounded_dims[side]);
-        binary_trmul_params.is_prepacked[side] = true;
-      }
-    }
+        transposed_lhs, internal_rhs, mul_params, ruy_ctx, &internal_dst,
+        bgemm_runtime_path, &binary_trmul_params);
 
-    ruy::TrMul(&binary_trmul_params, ruy_context);
+    HandlePrepackedCaching(&binary_trmul_params, ruy_ctx);
+    ruy::TrMul(&binary_trmul_params, ruy_ctx);
```
lgeiger, line 109: ⚠️ Review with care. I am not familiar with this code so I am not sure what these changes do, but for now at least they seem to compile.

/cc @Tombana @arashb

arashb: here I was using the RUY's advanced API but it is completely removed in [this](https://github.com/google/ruy/commit/57e64b4c8f32e813ce46eb495d13ee301826e498) commit. I need to dedicate more time to figure out how to replicate that.

Tombana:
The `is_mutable_` error turned out to be an issue with `const`ness of the lhs and rhs matrices. In the tflite/ruy code, a non-const `Matrix` object was created, and then passed to the `ruy::Mul` function as a `const Matrix`. However, since we don't have that function call, our matrix was non-`const`. This is now fixed.

Then a second error got thrown, because RUY removed the `kReference` path (now there's only `kStandardCpp`). For legacy reasons, they set `kReference = kStandardCpp`. We therefore can simply remove the check for not doing the reference path.

Tombana:
Rebased on master.

Tombana:
Just benchmarked this branch on the pixel. Quicknet got `18.0 ms` so there is no performance degradation (just to confirm we are not messing up the ruy prepacking or caching or anything). From my side it can be merged.

lgeiger:
> Just benchmarked this branch on the pixel. Quicknet got 18.0 ms so there is no performance degradation (just to confirm we are not messing up the ruy prepacking or caching or anything). From my side it can be merged.

Thanks for integrating all the changes and fixing the PR

Issue #67, opened 2019-10-15
Adhere to TensorFlow custom op naming scheme


lgeiger:
External custom ops should be prefixed with a namespace according to https://github.com/tensorflow/community/blob/master/rfcs/20190726-custom-ops.md#out-of-tree-ops-must-be-namespaced

I think we should adhere to that.

PR #364, opened 2020-05-14
Add converter support for a binarised maxpool op `LceBMaxPool2d`.

AdamHillier:
## What do these changes do?

This PR adds converter support for a new op `LceBMaxPool2d`: a 2D max-pooling op that operates on bitpacked (`int32`) input. This will allow a sequence of binary convolutions interspersed with `MaxPool2D` ops to still have bitpacked activation optimisations applied (see #352).

This PR *does not* include a TFLite registration or implementation of the op; only converter support. Converted models not be runnable until the op is actually implemented. However, this converter optimisation lives behind an off-by-default experimental flag, so that should not be a problem.

## How Has This Been Tested?

The PR includes new MLIR filecheck tests that will run on CI.

## Benchmark Results
N/A.

## Related issue number
N/A.


AdamHillier:
From a naming perspective, I would much rather this op were named `LCEBinaryMaxPool2D` because that would fit in much better with what TFL calls their ops (e.g. `DepthwiseConv2D`, `MaxPool2D`) but then wouldn't match our bconv op (`LceBconv2d`).

@larq/compute-engine am I right in thinking that we can't retrospectively change an op name without breaking compatibility with existing converted models?

arashb:
> From a naming perspective, I would much rather this op were named `LCEBinaryMaxPool2D` because that would fit in much better with what TFL calls their ops (e.g. `DepthwiseConv2D`, `MaxPool2D`) but then wouldn't match our bconv op (`LceBconv2d`).
> 
> @larq/compute-engine am I right in thinking that we can't retrospectively change an op name without breaking compatibility with existing converted models?

the current naming scheme is the result of the discussion [here](https://github.com/larq/compute-engine/issues/67). `Lce` at the beginning of the LCE ops is considered technically as a namespace and is not part of the Op name. I think we should not use different naming scheme for each op. If we decided to change to the naming scheme I would be in favour of doing it in a separate PR and for all other LCE ops as well to be consistent.

Tombana:
> the current naming scheme is the result of the discussion [here](https://github.com/larq/compute-engine/issues/67). `Lce` at the beginning of the LCE ops is considered technically as a namespace and is not part of the Op name. I think we should not use different naming scheme for each op. If we decided to change the naming scheme I would be in favour of doing it in a separate PR and for all other LCE ops as well to be consistent.

Good point. Back then we couldn't use the `>` symbol because we supported the op in Tensorflow 1.x . Now we don't have that problem anymore (we don't even have tensorflow ops in the first place) we could rename them to `Lce>BinaryConv2D` and `Lce>BinaryMaxPool2D`. I'm in favor of this and I agree this should be a separate PR. I also think it has no priority, and we should try to synchronize it with other breaking changes such as the attribute-cleanup.

Tombana:
I'd rather wait with merging this until the op kernels are merged, and then we can add end2end tests to this PR. (Rather than merging this first and adding those tests to the kernel PR).

AdamHillier:
Yep, very happy to wait until the C++ op is added. I've now ammended this PR to also allow float input for the op.

Tombana:
> Yep, very happy to wait until the C++ op is added. I've now ammended this PR to also allow float input for the op.

And that means also int8 when using QAT, right? I've added the op now in #365 :)

AdamHillier:
> > Yep, very happy to wait until the C++ op is added. I've now ammended this PR to also allow float input for the op.
> 
> And that means also int8 when using QAT, right? I've added the op now in #365 :)

I'm think that F32 implies Int8, but we'll find out when we write end-to-end tests after #365 is merged.

larq_compute_engine/mlir/transforms/optimize.cc
```diff
@@ -97,37 +97,43 @@ struct SetBconvReadWriteBitpacked : public OpRewritePattern<TF::LceBconv2dOp> {
             : dyn_cast_or_null<TF::LceBconv2dOp>(
                   outer_bconv_op.input().getDefiningOp());
 
-    if (!inner_bconv_op) return failure();
-
-    if (inner_bconv_op.padding() == "SAME" && inner_bconv_op.pad_values() != 1)
+    if (inner_bconv_op && inner_bconv_op.padding() == "SAME" &&
+        inner_bconv_op.pad_values() != 1)
       return failure();
```
Tombana: I don't yet understand this part of the code, but in case there is an `inner_bconv_op` but it has the wrong padding, then we should still use binarymaxpool right? But simply not with int32 input. I would code it like this:
```suggestion
    if (inner_bconv_op && inner_bconv_op.padding() == "SAME" &&
        inner_bconv_op.pad_values() != 1)
      inner_bconv_op = nullptr;
```

AdamHillier: Ah yes good point, I'll fix that

larq_compute_engine/tests/end2end_test.py
```diff
@@ -56,9 +56,10 @@ def toy_model_sequential(**kwargs):
                 gamma_initializer=tf.keras.initializers.RandomNormal(1.0),
                 beta_initializer="uniform",
             ),
-            tf.keras.layers.ReLU(),
+            # This will be converted to a float->bitpacked binary max pool.
+            tf.keras.layers.MaxPooling2D((2, 2)),
```
AdamHillier, line 60: This max pool will test the `float -> bitpacked` case.

larq_compute_engine/tests/end2end_test.py
```diff
@@ -71,8 +72,11 @@ def toy_model_sequential(**kwargs):
                 gamma_initializer=tf.keras.initializers.RandomNormal(1.0),
                 beta_initializer="uniform",
             ),
+            # This will be converted to a bitpacked->bitpacked binary max pool.
+            # Test some funky filter/stride combination.
+            tf.keras.layers.MaxPooling2D((3, 2), strides=(1, 2)),
```
AdamHillier, line 75: This will test the `bitpacked -> bitpacked` case.

larq_compute_engine/mlir/transforms/optimize.cc
```diff
@@ -71,44 +71,117 @@ DenseElementsAttr Bitpack(PatternRewriter& builder, Attribute x) {
...
-                              rewriter.getIntegerType(bitwidth));
-
-    rewriter.replaceOpWithNewOp<TF::LceBconv2dOp>(
-        bconv_input_op, new_inner_tensor_type, bconv_input_op.getOperands(),
-        bconv_input_op.getAttrs());
-    rewriter.replaceOpWithNewOp<TF::LceBconv2dOp>(bconv_op, bconv_op.getType(),
-                                                  bconv_op.getOperands(),
-                                                  bconv_op.getAttrs());
+    if (intermediate_maxpool_op &&
+        (!intermediate_maxpool_op.input().hasOneUse() ||
```
Tombana, line 88: If I understand correctly, this part tries to match   `bconv2d( maxpool(x) )`, and here its important that the output of `maxpool(x)` has no other uses. However this line of code tests if `x`, the input of maxpool has one use. Why should that matter?

AdamHillier: Ah good point, I think this is from previously when we didn't support just `maxpool -> bconv`. I'll fix it.

PR #366, opened 2020-05-17
Simplify WORKSPACE and remove old third_party files.

AdamHillier:
## What do these changes do?
This PR updates the `WORKSPACE` file to more closely match the TensorFlow `WORKSPACE` file. It also removes a lot of of legacy code from the `third_party` directory that was inherited from the original TensorFlow `custom-op` configuration but that we don't use anymore (it is mostly build configurations for GPU kernels and for dynamic linking against `libtensorflow`, neither of which is relevant to us). Finally, it removes the LCE core bitpacking and bgemm benchmarking utilities which are no longer used (I'm happy to add either or both of those back in if they're still considered useful).

## How Has This Been Tested?
CI.

## Benchmark Results
N/A.

## Related issue number
N/A.


WORKSPACE
```diff
@@ -61,9 +54,10 @@ http_archive(
     ],
 )
 
-# START: Upstream TensorFlow dependencies
-# TensorFlow build depends on these dependencies.
-# Needs to be in-sync with TensorFlow sources.
+# The remainder of this file is derived from, and should be kept in sync with,
+# the TensorFlow WORKSPACE file:
+# https://github.com/tensorflow/tensorflow/blob/master/WORKSPACE
```
AdamHillier, line 59: The new `WORKSPACE` file will be easy to update with new TF versions, because you can basically paste in the TF `WORKSPACE` file below this line and apply the replacement `//tensorflow..` -> `@org_tensorflow//tensorflow...`.

configure.sh
```diff
@@ -83,6 +76,16 @@ common --experimental_repo_remote_exec
...
 
+# By default, build TF in C++ 14 mode.
+build:android --cxxopt=-std=c++14
+build:android --host_cxxopt=-std=c++14
+build:linux --cxxopt=-std=c++14
+build:linux --host_cxxopt=-std=c++14
+build:macos --cxxopt=-std=c++14
+build:macos --host_cxxopt=-std=c++14
+build:windows --cxxopt=/std:c++14
+build:windows --host_cxxopt=/std:c++14
```
AdamHillier, line 87: This reflects the current TF `.bazelrc`: C++14 is now the default.

configure.sh
```diff
@@ -164,6 +167,14 @@ build:windows --copt=/w
 # On windows, we never cross compile
 build:windows --distinct_host_configuration=false
 
+build --define=use_fast_cpp_protos=true
+build --define=allow_oversize_protos=true
+
+# Enable using platform specific build settings, except when cross-compiling for
+# mobile platforms.
+build --enable_platform_specific_config
+build:android --noenable_platform_specific_config
```
AdamHillier, line 176: Also copied from the TF `.bazelrc`.

WORKSPACE
```diff
@@ -74,42 +68,33 @@ http_archive(
...
-)  # https://github.com/bazelbuild/rules_apple/releases
+# Please add all new TensorFlow dependencies in workspace.bzl.
+tf_repositories()
 
-http_archive(
-    name = "build_bazel_rules_swift",
-    sha256 = "18cd4df4e410b0439a4935f9ca035bd979993d42372ba79e7f2d4fafe9596ef0",
-    urls = ["https://github.com/bazelbuild/rules_swift/releases/download/0.12.1/rules_swift.0.12.1.tar.gz"],
-)  # https://github.com/bazelbuild/rules_swift/releases
+register_toolchains("@local_config_python//:py_toolchain")
```
lgeiger, line 78: Are we using this toolchain anywhere?

configure.sh
lgeiger, line 87: This is already defined [here](https://github.com/larq/compute-engine/pull/366/files#diff-735003ab6f5797c8b8bdecbbbb9322a8R141-R142), so we should get rid of one or the other.

configure.sh
```diff
@@ -164,6 +167,14 @@ build:windows --copt=/w
 # On windows, we never cross compile
 build:windows --distinct_host_configuration=false
 
+build --define=use_fast_cpp_protos=true
+build --define=allow_oversize_protos=true
+
+# Enable using platform specific build settings, except when cross-compiling for
+# mobile platforms.
+build --enable_platform_specific_config
```
lgeiger, line 175: Nice I've been meaning to do this as well.
Since this will automatically set the correct config depending on the platform, I think we should get rid of [those lines](https://github.com/larq/compute-engine/pull/366/files#diff-735003ab6f5797c8b8bdecbbbb9322a8R207-R211), specifically the `--config=windows` is now unnecessary.

WORKSPACE
AdamHillier: I don't believe so, I copied it from the TF `WORKSPACE` but will remove it.

AdamHillier:
> In a future PR I think we should also adopt the way TensorFlow handles their `.bazelrc` and check it into git. Dynamic configuration can then be managed using something like [this](https://github.com/tensorflow/tensorflow/blob/master/.bazelrc#L486-L492), which would make user specific bazel configurations easier to maintain (e.g. when using different locations for the Android SDK).

We should definitely do this, I've wanted to for a while too.

configure.sh
```diff
@@ -192,9 +201,3 @@ build --action_env ANDROID_SDK_API_LEVEL="29"
 build --action_env ANDROID_SDK_HOME="/tmp/lce_android"
 
 EOM
-
-if is_windows; then
-  write_to_bazelrc "build --config=v2 --config=windows"
-else
-  write_to_bazelrc "build --config=v2 --config=noaws --config=nogcp --config=nohdfs --config=nonccl"
```
lgeiger, line 199: I think we should still add `--config=noaws --config=nogcp --config=nohdfs --config=nonccl` to the `build:linux` and `build:macos` definitions since it will slightly speed up compile time by disabling features we don't need.

AdamHillier: Do those config options not work on Windows?

lgeiger: Yeah for some reason I got a conflict between `--config=windows` and `--config=noaws` in one of the TensorFlow internal build rules, I didn't spent time debugging the root cause, but removing removing them from the windows build fixed it.

PR #367, opened 2020-05-19
Simplify .bazelrc and configure.sh

lgeiger:
## What do these changes do?
This simplifies `configure.sh` by checking `.bazelrc` into `git`. This follows the approach taken by TensorFlow, allows to include user specific configurations and prevents issues that could arise when the upstream config changes but users forgot to rerun `./configure.sh`.

## How Has This Been Tested?
CI

## Related issue number
https://github.com/larq/compute-engine/pull/366#issuecomment-630099301

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -21,20 +21,15 @@ alias python='python3'
 PLATFORM="$(uname -s | tr 'A-Z' 'a-z')"
 
 function write_to_bazelrc() {
-  echo "$1" >> .bazelrc
+  echo "$1" >> .lce_configure.bazelrc
 }
 
 function is_linux() {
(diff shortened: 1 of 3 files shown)


configure.sh
```diff
@@ -56,151 +51,10 @@ fi
 if is_linux; then
   write_to_bazelrc "build:manylinux2010 --crosstool_top=//third_party/toolchains/preconfig/ubuntu16.04/gcc7_manylinux2010-nvcc-cuda10.1:toolchain"
```
lgeiger, line 52: I think we could further simplify `./configure.sh` or completely get rid of it, but let's keep it like this for now so we don't need to change the docs.

PR #372, opened 2020-05-21
Fix bitpacked activation pass when interacting with fused BN/activation.

### LINKED ISSUE #371: `BConv -> BN -> MaxPool -> BConv` pattern not being applied correctly with bitpacked activations.
Here is an example of a network sub-graph without our LCE optimisations being applied (hence no fusing of the ReLU and multiply/add):

image

In theory, with our optimisation pass (and `experimental_enable_bitpacked_activations=True`) this should be transformed to `-(float)-> BConv -(bitpacked)-> BinaryMaxpool -(bitpacked)-> BConv -(float)->`.

At the moment, we instead get `-(float)-> BConv -(float)-> BinaryMaxpool -(bitpacked)-> BConv -(float)->`. This is because the ReLU/BN fusing seems to happen after the bitpacked activation pass, so the unfused ops block the `bconv -> maxpool -> bconv` pattern being matched for the bitpacked activation pass, which means only `maxpool -> bconv` gets bitpacked.

This should be an easy fix by splitting the three-layer bitpacked pattern match into two two-layer pattern matches.

AdamHillier:
Closes #371.

This has the benefit of being a much cleaner implementation as well: it replaces the previous three-level pattern match with a single two-level pattern match (templated to accept both bconv and bmaxpool). As a result, any of the following will be bitpacked `bconv -> bconv`, `bconv -> bmaxpool`, `bmaxpool -> bmaxpool`, `bmaxpool -> bconv`.

With this PR, the rewrite pattern now works nicely even when there are fused activations and batch-norms that previously interfered:

image

### DIFF
--- a/larq_compute_engine/mlir/transforms/optimize.cc
+++ b/larq_compute_engine/mlir/transforms/optimize.cc
@@ -1,5 +1,6 @@
 #include "larq_compute_engine/core/packbits.h"
 #include "larq_compute_engine/mlir/ir/lce_ops.h"
+#include "llvm/ADT/Optional.h"
 #include "mlir/Dialect/StandardOps/IR/Ops.h"
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Pass/Pass.h"
@@ -68,24 +69,18 @@ DenseElementsAttr Bitpack(PatternRewriter& builder, Attribute x) {
 
 #include "larq_compute_engine/mlir/transforms/generated_optimize.inc"
 
-/**
- * Returns `false` if some assumption does not hold (the consumer should then
- * consider the rewrite pattern match to have failed).
- */
-bool getBitpackedType(PatternRewriter& rewriter, ShapedType existing_type,
-                      RankedTensorType& bitpacked_type) {
-  if (existing_type.getElementType().isInteger(32)) return false;
+llvm::Optional<RankedTensorType> getBitpackedType(PatternRewriter& rewriter,
+                                                  ShapedType existing_type) {
+  if (existing_type.getElementType().isInteger(32)) return llvm::None;
 
   const auto existing_shape = existing_type.getShape();
-
-  if (existing_shape.size() != 4) return false;
+  if (existing_shape.size() != 4) return llvm::None;
 
   const auto channels = existing_shape[3];
   const auto packed_channels = (channels + 32 - 1) / 32;
-  bitpacked_type = RankedTensorType::get({existing_shape[0], existing_shape[1],
-                                          existing_shape[2], packed_channels},
-                                         rewriter.getIntegerType(32));
-  return true;
+  return RankedTensorType::get({existing_shape[0], existing_shape[1],
+                                existing_shape[2], packed_channels},
+                               rewriter.getIntegerType(32));
 }
 
 template <typename BinaryOp>
@@ -107,16 +102,13 @@ struct SetBitpackedActivations : public OpRewritePattern<BinaryOp> {
         return failure();
       }
 
-      RankedTensorType bitpacked_type;
-      if (!getBitpackedType(rewriter,
-                            inner_bconv_op.getType().cast<ShapedType>(),
-                            bitpacked_type)) {
-        return failure();
-      }
+      auto maybe_bitpacked_type = getBitpackedType(
+          rewriter, inner_bconv_op.getType().cast<ShapedType>());
+      if (!maybe_bitpacked_type.hasValue()) return failure();
 
       rewriter.replaceOpWithNewOp<TF::LceBconv2dOp>(
-          inner_bconv_op, bitpacked_type, inner_bconv_op.getOperands(),
-          inner_bconv_op.getAttrs());
+          inner_bconv_op, maybe_bitpacked_type.getValue(),
+          inner_bconv_op.getOperands(), inner_bconv_op.getAttrs());
 
       return success();
     }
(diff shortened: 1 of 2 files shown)


larq_compute_engine/mlir/transforms/optimize.cc
```diff
@@ -68,123 +68,88 @@ DenseElementsAttr Bitpack(PatternRewriter& builder, Attribute x) {
...
+    if (!input_op || !input_op->hasOneUse()) return failure();
 
-    // If applicable, compute the new type for the inner bconv op.
+    // Try and match `input_op` to a binary convolution.
+    auto inner_bconv_op = dyn_cast_or_null<TF::LceBconv2dOp>(input_op);
     if (inner_bconv_op) {
-      ShapedType inner_bconv_op_type =
-          inner_bconv_op.getType().cast<ShapedType>();
+      if (inner_bconv_op.padding() == "SAME" &&
+          inner_bconv_op.pad_values() != 1) {
```
lgeiger, line 106: I guess in the future (maybe if we have our own dialect) we could generalize and simplify those checks by using [`OpInterfaces`](https://mlir.llvm.org/docs/Interfaces/) similar to how [TFLite handles setting a channel axis](https://github.com/tensorflow/tensorflow/blob/3d9ec6298a5d167c93e260605d3d2957d294fcb2/tensorflow/compiler/mlir/lite/ir/tfl_ops.td#L780-L783) for different ops.

larq_compute_engine/mlir/transforms/optimize.cc
```diff
@@ -68,123 +68,88 @@ DenseElementsAttr Bitpack(PatternRewriter& builder, Attribute x) {
...
-        (intermediate_maxpool_op && inner_bconv_op &&
-         !intermediate_maxpool_op.input().hasOneUse())) {
-      return failure();
-    }
+/**
+ * Returns `false` if some assumption does not hold (the consumer should then
+ * consider the rewrite pattern match to have failed).
+ */
+bool getBitpackedType(PatternRewriter& rewriter, ShapedType existing_type,
+                      RankedTensorType& bitpacked_type) {
```
lgeiger, line 76: How do you feel about changing this to something like:
```cpp
#include "llvm/ADT/Optional.h"

llvm::Optional<RankedTensorType> (PatternRewriter& rewriter, ShapedType existing_type) {
```

I think this would make the checks easier to read and would not require us to initialize `RankedTensorType` before using this function.

larq_compute_engine/mlir/transforms/optimize.cc
```diff
@@ -68,123 +68,88 @@ DenseElementsAttr Bitpack(PatternRewriter& builder, Attribute x) {
...
+
+  if (existing_shape.size() != 4) return false;
+
+  const auto channels = existing_shape[3];
+  const auto packed_channels = (channels + 32 - 1) / 32;
+  bitpacked_type = RankedTensorType::get({existing_shape[0], existing_shape[1],
+                                          existing_shape[2], packed_channels},
+                                         rewriter.getIntegerType(32));
+  return true;
+}
```
lgeiger, line 89: I think something like this should work:
```cpp
llvm::Optional<RankedTensorType> (PatternRewriter& rewriter, ShapedType existing_type) {
  if (existing_type.getElementType().isInteger(32)) return llvm::None;

  const auto existing_shape = existing_type.getShape();

  if (existing_shape.size() != 4) return llvm::None;

  const auto channels = existing_shape[3];
  const auto packed_channels = (channels + 32 - 1) / 32;
  return RankedTensorType::get({existing_shape[0], existing_shape[1],
                                          existing_shape[2], packed_channels},
                                         rewriter.getIntegerType(32));
}
```

larq_compute_engine/mlir/transforms/optimize.cc
```diff
@@ -68,123 +69,77 @@ DenseElementsAttr Bitpack(PatternRewriter& builder, Attribute x) {
...
-    if (intermediate_maxpool_op) {
-      ShapedType intermediate_maxpool_op_type =
-          intermediate_maxpool_op.getType().cast<ShapedType>();
+      rewriter.replaceOpWithNewOp<TF::LceBconv2dOp>(
+          inner_bconv_op, maybe_bitpacked_type.getValue(),
+          inner_bconv_op.getOperands(), inner_bconv_op.getAttrs());
 
-      // Don't apply this transformation if the type is already I32.
-      if (intermediate_maxpool_op_type.getElementType().isInteger(bitwidth)) {
+      return success();
```
lgeiger, line 113: I had hoped that we can use something like the following here so we don't need to call `hasValue` and `getValue`, if this this doesn't work, I'm happy with the current solution.

```cpp
if (auto bitpacked_type = getBitpackedType(rewriter, inner_bconv_op.getType().cast<ShapedType>())) {
  ...
  return success();
}
return failure();

Comment (AdamHillier) [reply larq_compute_engine/mlir/transforms/optimize.cc:113]: When using this `bitpacked_type` is still the optional type, so you have to call `getValue()` inside the if block.

Given that's the case, I think I'd rather be explicit and use the `hasValue()` check rather than relying on the truthiness of `llvm::None` being false.

Comment (lgeiger) [reply larq_compute_engine/mlir/transforms/optimize.cc:113]: Makes sense, I wasn't aware of that.

Though judging from how it is [used in TensorFlow](https://github.com/tensorflow/tensorflow/blob/e164659e5b607f2e7fc54fbc090894c78745d544/tensorflow/compiler/mlir/lite/flatbuffer_export.cc#L1211-L1213) one should be able to do:
```cpp
if (auto bitpacked_type = getBitpackedType(rewriter, inner_bconv_op.getType().cast<ShapedType>())) {
  rewriter.replaceOpWithNewOp<TF::LceBconv2dOp>(
          inner_bconv_op, *bitpacked_type,
          inner_bconv_op.getOperands(), inner_bconv_op.getAttrs());
  return success();
}
```
If you pass it using `*bitpacked_type` it might work, but I migth miss something. In both cases this PR looks good to me :+1:

AdamHillier: Nice, you're right it does work with the dereference.

PR #373, opened 2020-05-22
Rename BConv attributes to follow TFLite more closely

lgeiger:
## What do these changes do?

This is a follow up on #370 and changes the attributes of the binary convolution to follow TFLite more closely, this should also make handling of the attributes in the kernel a bit easier.

The main changes are:
- Replace `dilations` array of length 4 with two integers called `dilation_height_factor` and `dilation_width_factor` defining the spatial dilation_factors
- Replace `strides` array of length 4 with two integers called  `stride_height` and `stride_width` defining the spatial strides
- Rename `activations` to `fused_activation_function` following the naming convention of TFLite (I don't feel strongly about this change but I think it is nice to stay consistent in this case)
- Ordered attributes alphabetically to make MLIR test consistent since in the IR they will always be ordered alphabetically anyway.

~~I didn't touch the kernels yet, but it would be good to get this reviewed and agree on the name change before I put any work into this.~~ The PR also doesn't remove the `filter_format` attribute in order to keep the diff at a sensible size.

## How Has This Been Tested?
LIT tests on CI


larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -52,6 +52,15 @@ inline void decide_bitpack_before_im2col(KernelType kernel_type,
   }
 }
 
+#define LCE_ENSURE_PARAM(conv_params, context, a)                       \
```
lgeiger, line 55: I didn't use `TFLITE_ENSURE` here since it would return `kTfliteError` instead of `conv_params`. Not sure if this is the best way to handle the errors though.

Tombana: This looks like a nice solution to me.

Side note: There is no solution we can copy from tflite, because they don't have to read a flexbuffer in `Init`: their `struct` with parameters is already filled, and correctness is ensured by the Interpreter and the flatbuffer schema. Therefore they never have to report errors from `Init`.

Tombana: We could have the following to speed up non-test builds, but since its only for initialization I think we don't have to.
```cpp
#ifdef NDEBUG
#define LCE_ENSURE_PARAM(conv_params, context, a) 
#else
#define LCE_ENSURE_PARAM(conv_params, context, a)  do { if(!(a)) { ... } } while(0)
#endif
```

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -71,23 +80,17 @@ void* Init(TfLiteContext* context, const char* buffer, std::size_t length) {
...
   // reading dilations
-  auto dilation_vector = m["dilations"].AsTypedVector();
-  if (dilation_vector.size() != 4) {
-    context->ReportError(context, "Dilations vector should have size 4.");
-    return conv_params;
-  }
-  for (std::size_t i = 0; i < dilation_vector.size(); ++i)
-    conv_params->dilations[i] = dilation_vector[i].AsInt64();
+  conv_params->dilation_height_factor = m["dilation_height_factor"].AsInt64();
+  conv_params->dilation_width_factor = m["dilation_width_factor"].AsInt64();
```
Tombana, line 93: Nice, this is a lot cleaner!

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -71,23 +80,17 @@ void* Init(TfLiteContext* context, const char* buffer, std::size_t length) {
...
 
   // reading padding
   if (m["padding"].ToString() == "VALID" ||
```
Tombana, line 96: I'm not sure why we originally had both the upper and lowercase checks here. I think we can just remove one of them? I know its not entirely relevant to this PR, but since we're updating the attributes I think we can do this here as well.

larq_compute_engine/tflite/kernels/bconv2d.cc
lgeiger: Thanks for the explanation.

> We could have the following to speed up non-test builds, but since its only for initialization I think we don't have to.

I don't think we need to do this, since `TF_LITE_ENSURE` uses a very similar macro which is used often throughout the codebase so optimizing compile time of `LCE_ENSURE_PARAM` will probably not be noticable.

On a different note, if we really need to reduce binary size, we can replace calls to `context->ReportError` with [`TF_LITE_KERNEL_LOG`](https://github.com/tensorflow/tensorflow/blob/f44bb87843cababb97401e93ae9d91e6a296c293/tensorflow/lite/c/common.h#L153-L171) which allows to strip out message strings if the `TF_LITE_STRIP_ERROR_STRINGS` flag is set.

larq_compute_engine/tflite/kernels/bconv2d.cc
lgeiger: I can check this out in a future PR and see if it is cleaner to directly pass the TFLite padding enum (same can be done for the activation strings then as well).

larq_compute_engine/tflite/kernels/bconv2d.cc
Tombana: Agreed. And that `TF_LITE_KERNEL_LOG` is a good idea, lets do that in a separate PR.

larq_compute_engine/tflite/kernels/bconv2d_params.h
```diff
@@ -25,10 +25,12 @@ typedef struct {
   std::int64_t channels_out{0};
 
   // strides
-  std::int64_t strides[4] = {};
+  std::int64_t stride_height{0};
+  std::int64_t stride_width{0};
```
Tombana, line 29: Now that we are changing this anyway, we should change it to `std::int32_t` to match the MLIR spec (as well as the dilations). And update the `.AsInt64()` flexbuffer call to `.AsInt32()`.

Originally we had this as 64-bit, I think because the flexbuffer function `AsInt64()` seemed more general than the `.AsInt32` one, and with the old non-MLIR converter we weren't sure what the actual data-type in the flexbuffer was, and `.AsInt64()` would convert correctly from any flexbuffer type.

lgeiger: Good point, I changed it to `int32` in b1a0b2f61df035a9e159192c7f857bbd5ae3223b

PR #375, opened 2020-05-22
Add pattern to ensure post_activation_multipliers are always positiv

lgeiger:
## What do these changes do?
This PR makes sure that `post_activation_multiplier`s will always  be positive if no fused activation is present by absorbing the sign in the binary weights. This will enable future optimizations to precompute thresholds in the converter when writing bitpacked output.

## How Has This Been Tested?
CI

### DIFF
--- a/larq_compute_engine/mlir/tests/optimize.mlir
+++ b/larq_compute_engine/mlir/tests/optimize.mlir
@@ -144,6 +144,18 @@ func @do_not_fuse_relu_into_bconv2d_no_post_activation_multiplier(%arg0: tensor<
   // CHECK-NEXT: return %1
 }
 
+// CHECK-LABEL: @tranform_negative_multipliers
+func @tranform_negative_multipliers(%arg0: tensor<256x32x32x3xf32>, %arg1: tensor<16xf32>) -> tensor<256x30x30x16xf32> {
+  %multiplier = constant dense<-3.0> : tensor<16xf32>
+  %filter = constant dense<1.0> : tensor<16x3x3x3xf32>
+  %0 = "tf.LceBconv2d"(%arg0, %filter, %multiplier, %arg1) {activation = "NONE", channels_in = 3 : i32, filter_format = "OHWI", padding = "VALID", strides = [1, 1, 1, 1]} : (tensor<256x32x32x3xf32>, tensor<16x3x3x3xf32>, tensor<16xf32>, tensor<16xf32>) -> tensor<256x30x30x16xf32>
+  return %0 : tensor<256x30x30x16xf32>
+
+  // CHECK: %cst_0 = constant dense<3.000000e+00> : tensor<16xf32>
+  // CHECK: %0 = "tf.LceBconv2d"(%arg0, %cst, %cst_0, %arg1)
+  // CHECK-NEXT: return %0
+}
+
 // CHECK-LABEL: @do_not_change_multiplier_if_fused_activation
 func @do_not_change_multiplier_if_fused_activation(%arg0: tensor<256x32x32x3xf32>, %arg1: tensor<16xf32>) -> tensor<256x30x30x16xf32> {
   %multiplier = constant dense<-3.0> : tensor<16xf32>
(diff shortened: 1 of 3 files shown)


larq_compute_engine/mlir/tests/optimize.mlir
```diff
@@ -145,12 +145,14 @@ func @do_not_fuse_relu_into_bconv2d_no_post_activation_multiplier(%arg0: tensor<
...
+func @bitpack_bconv2d_filters(%arg0: tensor<256x32x32x3xf32>, %arg1: tensor<16xf32>) -> tensor<256x30x30x16xf32> {
+  %multiplier = constant dense<-3.0> : tensor<16xf32>
+  %filter = constant dense<1.0> : tensor<16x3x3x3xf32>
+  %0 = "tf.LceBconv2d"(%arg0, %filter, %multiplier, %arg1) {activation = "NONE", channels_in = 3 : i32, filter_format = "OHWI", padding = "VALID", strides = [1, 1, 1, 1]} : (tensor<256x32x32x3xf32>, tensor<16x3x3x3xf32>, tensor<16xf32>, tensor<16xf32>) -> tensor<256x30x30x16xf32>
   return %0 : tensor<256x30x30x16xf32>
 
-  // CHECK: %0 = "tf.LceBconv2d"(%arg0, %cst, %arg1, %arg2) {activation = "NONE", channels_in = 3 : i32, dilations = [1, 1, 1, 1], filter_format = "OHWI_PACKED", pad_values = 0 : i32, padding = "VALID", strides = [1, 1, 1, 1]} : (tensor<256x32x32x3xf32>, tensor<16x3x3x1xi32>, tensor<16xf32>, tensor<16xf32>) -> tensor<256x30x30x16xf32>
+  // CHECK: %cst_0 = constant dense<3.000000e+00> : tensor<16xf32>
```
lgeiger, line 154: I wanted to also check that the filter is properly inverted, but due to bitpacking this is a bit more tricky than expected. I manually verified that the pattern is correct.

larq_compute_engine/mlir/transforms/optimize_patterns.td
```diff
@@ -41,9 +41,23 @@ multiclass FuseMulOrDivWithBConv2D<dag binaryOp> {
...
+def : Pat<(TF_LceBconv2dOp $input, (ConstantOp F32ElementsAttr:$filter),
+            (ConstantOp HasNegativeValues:$post_activation_multiplier),
+            $post_activation_bias, $channels_in, $strides, $padding,
+            $pad_values, $dilations, $filter_format, $act),
+          (TF_LceBconv2dOp $input,
+            (TFL_MulOp (ConstantOp $filter), (ConstantOp (ComputeBSignAndExpandTo4D $post_activation_multiplier)), TFL_AF_None),
+            (TFL_AbsOp (ConstantOp $post_activation_multiplier)),
+            $post_activation_bias, $channels_in, $strides, $padding,
+            $pad_values, $dilations, $filter_format, $act),
+          [], (addBenefit 90)>;
```
Tombana, line 59: Really nice that you can just use the `MulOp` for this and don't have to manually iterate through the weights.
Some questions:
- The `addBenefit` numbers that you changed, are those to make sure that this happens after the batchnorm has been merged?
- How do you make sure this pattern happens before bitpacking the filters?
- I think this pattern should only be applied when there is no activation function, so the `$act` should be replaced by `TFL_AF_None`.

lgeiger: > Really nice that you can just use the MulOp for this and don't have to manually iterate through the weights.

Yes, if we would implement constant folding for the bsignop, we could've also used it here as well.

> The `addBenefit` numbers that you changed, are those to make sure that this happens after the batchnorm has been merged?

Its main goal is to make sure that the transformation happens before the bitpacking the weights, since we can't apply the transformation after it. In most cases this will also make the pattern run after batchnorm fusing, but this wouldn't be strickly necessary since the pattern will be applied again if batchnorm fusing would reintroduce negative multipliers.

> How do you make sure this pattern happens before bitpacking the filters?

`F32ElementsAttr:$filter` makes sure it doesn't run afterwards and `addBenefit` makes sure that it runs before. I guess in the future it probably makes sense to extract the bitpacking related transforms into their own pass that runs afterwards which would also make testing much easier.

> I think this pattern should only be applied when there is no activation function, so the `$act` should be replaced by `TFL_AF_None`.

Good catch, thanks for noticing that! I fixed it in 1d4b29ba6cfd0d9865035fb398ca86d07664ade3
I was distracted by some bitpacking related bugs so this slipped by when pushing the PR.

larq_compute_engine/mlir/tests/optimize.mlir
lgeiger: Fixed in dcf5db16fe0cee65cbd9b15fe81baba362051837

PR #377, opened 2020-05-26
Fix weight bitpacking which could lead to non-deterministic behaviour 

lgeiger:
## What do these changes do?

While writing tests for #375 I this covered that the previous implementation could lead to non-deterministic bitpacking of the weights during conversion

## How Has This Been Tested?
Added a filecheck test which will run on CI.

This bugfix could influence network accuracy, so I will double check if this is an issue in 0.3 as well and will backport the fix.


lgeiger:
This bug is also present in 0.2 and 0.3. I will backport the fix to 0.3 and release 0.3.1 after this has been merged.

PR #380, opened 2020-05-26
Add testing script for faster model conversions debugging

lgeiger:
## What do these changes do?

This PR adds a small script that converts a model and saves the generated flatbuffer on disk. This enable easier debugging when developing new conversion features that need to be tested on a full model without the need to build the pip package or modify the end2end tests to write the file to disk (which I found myself manually doing quite often).

## How Has This Been Tested?
Tested manually with quicknet.

### DIFF
--- a/larq_compute_engine/tests/BUILD
+++ b/larq_compute_engine/tests/BUILD
@@ -29,6 +29,14 @@ py_test(
     ],
 )
 
+py_test(
+    name = "convert_model",
+    srcs = ["convert_model.py"],
+    deps = [
+        "//larq_compute_engine/mlir:converter",
+    ],
+)
+
 test_suite(
     name = "cc_tests",
     tests = [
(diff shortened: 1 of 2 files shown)


larq_compute_engine/tests/BUILD
```diff
@@ -29,6 +29,14 @@ py_test(
     ],
 )
 
+py_test(
```
lgeiger, line 32: I used `py_test` instead of `py_binary` in order to enable running this via `bazel test` which makes sure that the same compile flags are set so local build caches don't get nuked.

lgeiger:
> Maybe in the usage example add a note that it should be used with absolute paths, because the working directory of the python script will be something like bazel-bin/larq_compute_engine/tests/convert_model.runfiles/ which is usually not where you want the result stored.

The command will print the absolute path of the saved model at the end, but I added a note to the help text in 2c05b37606661cecb6417f8ffe8a9cc52dd5b856

> I think it makes sense to add convert_model.py to .gitignore to prevent accidental committing..

I'd prefer not to add it to `.gitignore` because one of the reasons why this script would be useful for me is to have a place to temporarily commit test models on long lived branches like we did for the int8 quantization without the risk of introducing merge conflicts when rebasing those branches.

PR #381, opened 2020-05-27
Upgrade github/actions

lgeiger:
## What do these changes do?
The main reason for this upgrade is to use `actions/checkout@v2` and `actions/cache@v2`  which should improve CI time.

## How Has This Been Tested?
CI

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -14,12 +14,12 @@ jobs:
       image: tensorflow/tensorflow:custom-op-ubuntu16
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
         with:
           submodules: true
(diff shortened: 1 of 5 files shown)


.github/workflows/unittests.yml
```diff
@@ -100,11 +100,13 @@ jobs:
     if: "!contains(github.event.head_commit.message, 'ci-skip')"
 
     steps:
-      - uses: actions/checkout@v1
-      - uses: actions/setup-python@v1
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
AdamHillier, line 105: What does fetch-depth do?

lgeiger: It fetches the whole history including the tags, which we need because the script call `git describe --tags`

PR #382, opened 2020-05-27
Remove filter_format attribute from IR and bconv kernel

lgeiger:
## What do these changes do?
This is a follow up to #373 and removes the `filter_format` attribute from the IR and the bconv kernels since we only support `OHWI` and `OHWI_PACKED` which can be infered by the dtype in `Prepare`

## How Has This Been Tested?
CI

## Related issue number
This targets the branch of #373 but can be reviewed independently.


larq_compute_engine/mlir/transforms/optimize_patterns.td
```diff
@@ -117,10 +117,10 @@ def Bitpack : NativeCodeCall<"Bitpack($_builder, $0)">;
 def : Pat<(TF_LceBconv2dOp $input, (ConstantOp Conv2DFilter:$filter),
             $post_activation_multiplier, $post_activation_bias,
             $channels_in, $dilation_height_factor, $dilation_width_factor,
-            ConstantAttr<StrAttr, "OHWI">, $fused_activation_function,
+            $fused_activation_function,
```
Tombana, line 120: Running the pattern only once is now ensured by the filter tensor type, instead of this argument, right?

lgeiger: Yes, it is should now be handled by this check:
https://github.com/larq/compute-engine/blob/f69edc70ad1e5efda5bcac1d22d2744d861e05f4/larq_compute_engine/mlir/transforms/optimize.cc#L41
inside the `Conv2DFilter` attr constraint

PR #385, opened 2020-05-28
Support dilated convolutions in converter

lgeiger:
## What do these changes do?
@koenhelwegen noticed that we currently do not correctly convert dilated convolutions since they may be represented using `SpaceToBatchND` on the TF side instead of properly set `dilation` arguments.

This PR copies the `ConvertTFDilatedConvOp` pattern from the `TFPass` to the `LCEPass` so it happens earlier which allows us to properly convert dilated binary convolutions.

## How Has This Been Tested?
mlir tests

### DIFF
--- a/larq_compute_engine/mlir/BUILD
+++ b/larq_compute_engine/mlir/BUILD
@@ -126,6 +126,7 @@ cc_library(
         ":larq_compute_engine",
         "@llvm-project//mlir:StandardOps",
         "@org_tensorflow//tensorflow/compiler/mlir/lite:tensorflow_lite",
+        "@org_tensorflow//tensorflow/compiler/mlir/lite:tensorflow_lite_legalize_tf",
         "@org_tensorflow//tensorflow/compiler/mlir/lite:validators",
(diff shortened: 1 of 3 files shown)


lgeiger:
> Agreed, LGTM when an end2end test is added. Ideally two of them, one for which TF uses SpaceToBatch and one where TF doesn't. Would that be possible, or can we not control when TF does that?

I don't think that adding an end2end test will help much in this case since even before that PR the network ran properly with TFLite (just without our op) so dilated binary convolutions would just translate `Bsign` and `Conv2D` which wouldn't be caught by the end2end tests. I manually verified that with this PR I can properly convert `BinaryDenseNet37Dilated` though.

AdamHillier:
Ah okay I see what you mean, interesting. I guess you could hackily verify this by asserting a max length of the flatbuffer, as if it's not all bconvs the weights will be much bigger. But that's ugly and brittle, and as this isn't an issue that would affect accuracy I think you're right that manual verification is fine for now.

lgeiger:
Rebased and resolved the merge conflicts 🚀

PR #384, opened 2020-05-28
Use custom larq MLIR dialect for our ops

lgeiger:
TensorFlow added a `tfl.custom` op to the `TFL` dialect in https://github.com/tensorflow/tensorflow/commit/fb7ea8f0e66899587e71311adbd76536d571f068. This allows us to decouple our ops from the `TF` dialect which was previously necessary to allow for flatbuffer serialization.

Essentially this PR does two things: It moves our op definition to a custom larq dialect and introduces a legalize pass at the end that translates our ops to `tfl.custom` ops which can be correctly serialized to the TFLite flatbuffer.

This has a few advantages over the current approach of adding our ops to the `TF` dialect:
- Full compatibility with MLIR passes and patterns
   - Proper type checking in the IR and during transformation
      - should prevent bugs that could lead to invalid flatbuffers
      - will allow us to introduce strict verification of the correcness of the IR
   - Proper handling of MLIR traits which
      - MLIR can properly check if ops have side effects and remove trailing ones without the need for custom op cleanup patterns
      - will allow us to potentially use [quantization traits](https://github.com/tensorflow/tensorflow/blob/ea8e87c8e945c5424b130d66a4fbf5fca7e5c9cd/tensorflow/compiler/mlir/lite/quantization/quantization.td#L59-L92) in the future if necessary for easier customization of bias quantizations 
- Full flexibility of attribute sersialization and deserialization (If really necessary we could even get rid of flexbuffers and roll with our own attribute serialization)
- Autogenerated docs for our IR

I marked this as a draft PR since it is built ontop of #373 and #382

<details><summary><b>Click to expand autogenerated docs</b></summary>
<p>

# 'lq' Dialect

Types and operations for Larq dialect

This dialect contains operations for Larq. This dialect will be used in
conjunction with the TensorFlow dialects for converting & optimizing
TF graphs to be deployed on Larq Compute Engine.

[TOC]

### `lq.BMaxPool2d` (TF::BMaxPool2dOp)

Binary MaxPool2D op.
  

Computes a MaxPool2D operation and outputs bitpacked binary values, for consumption by a binary convolution.

#### Attributes:

| Attribute | MLIR Type | Description |
| :-------: | :-------: | ----------- |
`padding` | StringAttr | padding enum
`stride_width` | IntegerAttr | 32-bit signless integer attribute
`stride_height` | IntegerAttr | 32-bit signless integer attribute
`filter_width` | IntegerAttr | 32-bit signless integer attribute
[...]

PR #386, opened 2020-05-29
Set padding and fused activations as enums

lgeiger:
## What do these changes do?
This changes the padding and fused activation attributes to be enums instead of strings to more closely follow TFLite and simplify the kernel implementation.

~~I marked it as a draft PR since it is built ontop of #384~~

## How Has This Been Tested?
CI

### DIFF
--- a/larq_compute_engine/mlir/ir/lce_ops.cc
+++ b/larq_compute_engine/mlir/ir/lce_ops.cc
@@ -1,9 +1,25 @@
 #include "larq_compute_engine/mlir/ir/lce_ops.h"
 
 #include "flatbuffers/flexbuffers.h"
+#include "tensorflow/lite/c/builtin_op_data.h"
 
 namespace mlir {
 namespace TF {
+
+static TfLitePadding ConvertTfLitePaddingAttr(llvm::StringRef str) {
+  return llvm::StringSwitch<TfLitePadding>(str)
+      .Case("SAME", kTfLitePaddingSame)
+      .Case("VALID", kTfLitePaddingValid);
+}
+
+static TfLiteFusedActivation ConvertTfLiteFusedActivationAttr(
+    llvm::StringRef str) {
+  return llvm::StringSwitch<TfLiteFusedActivation>(str)
+      .Case("NONE", kTfLiteActNone)
+      .Case("RELU", kTfLiteActRelu)
+      .Case("RELU_N1_TO_1", kTfLiteActRelu1)
+      .Case("RELU6", kTfLiteActRelu6);
+}
 
 #define GET_OP_CLASSES
 #include "larq_compute_engine/mlir/ir/lce_ops.cc.inc"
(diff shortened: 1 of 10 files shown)


larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -80,34 +80,10 @@ void* Init(TfLiteContext* context, const char* buffer, std::size_t length) {
...
-  } else if (m["padding"].ToString() == "SAME" ||
-             m["padding"].ToString() == "same") {
-    conv_params->padding_type = kTfLitePaddingSame;
-  } else {
-    context->ReportError(context, "Invalid padding attribute.");
-    return conv_params;
-  }
+  conv_params->padding_type = (TfLitePadding)m["padding"].AsInt32();
+  conv_params->fused_activation_function =
+      (TfLiteFusedActivation)m["fused_activation_function"].AsInt32();
```
lgeiger, line 85: Not sure if we need to do some error checking here.

larq_compute_engine/mlir/ir/lce_ops.cc
```diff
@@ -1,10 +1,26 @@
...
 #include "flatbuffers/flexbuffers.h"
+#include "tensorflow/lite/c/builtin_op_data.h"
 
 namespace mlir {
 namespace TF {
 
+static TfLitePadding ConvertTfLitePaddingAttr(llvm::StringRef str) {
+  return llvm::StringSwitch<TfLitePadding>(str)
+      .Case("SAME", kTfLitePaddingSame)
+      .Case("VALID", kTfLitePaddingValid);
```
lgeiger, line 12: I am not 100% sure if this is the enum TFLite uses in the flatbuffer @Tombana  do you know?

AdamHillier:
How do you feel about getting rid of `pad_values` and merging that into a custom padding enum?

lgeiger:
> How do you feel about getting rid of pad_values and merging that into a custom padding enum?

Yeah, thought about that too. I don't really have strong opinions but the nice things now would be that we can reuse the default TFLite enum

larq_compute_engine/mlir/ir/lce_ops.cc
Tombana: I think this is correct. The tflite conv2d kernel gets an `OpData` struct with a parameter `padding` which is directly passed into a function that expects these `kTfLitePaddingXXX` values.

larq_compute_engine/tflite/kernels/bconv2d.cc
Tombana: I'd opt for adding an `LCE_ENSURE_PARAM(conv_params, context, !m["fused_activation_function"].IsNull());` to the list thats slightly above these lines.

lgeiger: Good point, added a check in f092d5b31a33c0d64d886869269526a7f7e5061c

PR #387, opened 2020-05-29
Use thresholds for bitpacked output

Tombana:
## What do these changes do?
This changes the output transformation for the case of bitpacked output. The accumulator is now simply compared with a single threshold, that is assumed to be present in the flatbuffer.

(This is required for a big cleanup that we're doing on the microcontroller side)

## How Has This Been Tested?
Unittests have been updated.

## Benchmark Results
Only relevant for microcontrollers, where it gives a speedup of up to 10%.

## Related
We should only merge this synced with @AdamHillier's upcoming PR that does the matching converter update.


larq_compute_engine/tflite/tests/utils.h
```diff
@@ -97,9 +97,9 @@ struct GetPostType<float> {
 
 template <>
 struct GetPostType<std::int8_t> {
-  // We can also choose to have this in float, the kernel supports both.
-  // using type = float;
-  using type = std::int8_t;
+  // The converter curretly uses float, the kernel supports both
```
Tombana, line 100:
```suggestion
  // The converter currently uses float, the kernel supports both
```

AdamHillier:
(Just rebased this on master)

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -150,22 +150,28 @@ TfLiteStatus Prepare(KernelType kernel_type,
...
   if (!conv_params->conv_params_initialized) return kTfLiteError;
 
-  TF_LITE_ENSURE(context, node->inputs->size == 4);
+  TF_LITE_ENSURE(context, node->inputs->size == 5);
 
   const auto* input = GetInput(context, node, 0);
   const auto* filter = GetInput(context, node, 1);
   const auto* post_activation_multiplier = GetInput(context, node, 2);
   const auto* post_activation_bias = GetInput(context, node, 3);
+  const auto* thresholds = GetInput(context, node, 4);
```
lgeiger, line 159: What is the benefit of passing the thresholds as an operand rather than a kernel attribute?

Currently the threshold will always be a constant that is computed during the conversion process and cannot be the output of an other op unless we change the pattern to compute the threshold using TFLite ops. These ops would then be constant folded in the normal case but could be part of the graph if multiplier and bias are dynamic. However this is probably not practical due to all the edge cases we handle in the current conversion.

larq_compute_engine/mlir/ir/lce_ops.td
```diff
@@ -57,8 +60,9 @@ TODO
   let arguments = (ins
     TensorOf<[F32]>:$input,
     TensorOf<[F32, I32]>:$filter,
-    TensorOf<[F32]>:$post_activation_multiplier,
-    TensorOf<[F32]>:$post_activation_bias,
+    TensorOfOrNone<[F32]>:$post_activation_multiplier,
+    TensorOfOrNone<[F32]>:$post_activation_bias,
+    TensorOfOrNone<[I32]>:$write_bitpacked_threshold,
```
lgeiger, line 65: I think I would prefer
```tablegen
    TensorOfOrNone<[I32]>:$output_threshold,
```

AdamHillier: Yeah I'm happy with that name

larq_compute_engine/tflite/kernels/bconv2d.cc
AdamHillier: Yeah you're right, it probably could be an attribute rather than an operand, I hadn't thought about that. How strongly do you feel about this?

lgeiger: I don't feel strongly either way, I might even prefer it being an operand. I was just curious about the reasoning.
This came up when looking at the attribute name and I got confused that this PR didn't mention `write_bitpacked_threshold` before was #387 merged and the unittests still pass...

Tombana: I don't feel strongly about it. So far I had been using the idea that variable-length data (i.e. depending on input shape) would be input and constant-size data would be attributes. But I'm fine with changing it if that makes more sense.

lgeiger: > 

I don't feel strongly about it. So far I had been using the idea that variable-length data (i.e. depending on input shape) would be input and constant-size data would be attributes. But I'm fine with changing it if that makes more sense.

I agree, I think it makes sense to keep it an operand 👍

PR #389, opened 2020-06-01
Remove RUY_ASM_FLAG_HAS_BIAS

lgeiger:
## What do these changes do?

This PR removes the handling of cases where `post_activation_bias` can be null from the optimized convolution kernels.
In the converter `post_activation_bias` and `post_activation_multiplier` will always be set, so since we don't check for the existence of `post_activation_multiplier` in the optimized assembly we can also get rid of the check for `post_activation_bias`. This check has been added moved over from TFLite which allows biases to be null. This is not the case here so I think this can be savely removed from the optimized code path.

## How Has This Been Tested?

This has purely been tested on CI since I originally made these changes when experiment with #263 only now rebased them onto the latest master.

## Benchmark Results
I didn't ran any benchmarks for these changes, but I am happy to do so if you think it would be useful

## Related issue number
Removing these checks might make implementing AArch64 support for #387 a bit easier.

### DIFF
--- a/larq_compute_engine/core/bgemm_kernels_common.h
+++ b/larq_compute_engine/core/bgemm_kernels_common.h
@@ -57,7 +57,6 @@ struct BinaryKernelParams {
   std::int32_t clamp_min;
   std::int32_t clamp_max;
   std::int32_t backtransform_add;
-  std::uint8_t flags;
   float dst_tmp_buf[LhsCols * RhsCols];
 };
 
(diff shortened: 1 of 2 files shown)


larq_compute_engine/core/bgemm_kernels_common.h
```diff
@@ -130,9 +127,6 @@ inline void MakeBinaryKernelParams(
   params->post_activation_multiplier =
       spec.output_transform.post_activation_multiplier;
   params->post_activation_bias = spec.output_transform.post_activation_bias;
-  if (params->post_activation_multiplier && params->post_activation_bias) {
-    flags |= RUY_ASM_FLAG_HAS_BIAS;
-  }
   params->backtransform_add = spec.output_transform.backtransform_add;
   params->flags = flags;
```
AdamHillier, line 131: Should we remove the `flags` param altogether (this was the only flag being used)?

lgeiger: Good point, removed them in a0ca403b78562fcd6c5c5e34324011bdb5fcafbc

I guess if we need them at some point we can add them back rather easily.

PR #391, opened 2020-06-01
Compute bitpack thresholds in the converter

AdamHillier:
## What do these changes do?

When using bitpacked activations, we currently perform the output transformation (back-transform, clamping, post-multiply, post-add) and then take the sign of the result to decide whether to write a 1-bit or a 0-bit. It would be more efficient to simply compare the xor-popcount accumulator against a pre-computed threshold in order to make this decision.

In #387 @Tombana is modifying the kernels to instead support this behaviour. This PR extends the converter to support pre-computing the thresholds.

Doing this involves adding a fifth input to the `bconv` op. However, when not writing bitpacked activations the threshold input won't be used, and conversely when writing bitpacked input neither the post-multiply nor the post-addition will be used. As a result, to avoid bulking up the flatbuffer unnecessarily, this PR makes all three of those inputs optional, and only includes them when necessary for the computation.

To make implementation easier, the pattern that bitpacks the weights has been moved out of the optimise pass and into its own dedicated pass that runs afterwards.


AdamHillier:
If/when this is approved, please don't merge it just yet.

larq_compute_engine/mlir/transforms/op_removal_patterns.td
```diff
@@ -22,18 +22,31 @@ def : Pat<(TF_PlaceholderWithDefaultOp $arg), (TF_IdentityOp $arg)>;
...
 def : Pat<(TF_LceBsignOp:$op $x), (replaceWithValue $x), [(HasNoUseOf:$op)]>;
-def : Pat<(TF_LceBconv2dOp:$op $input, $filter, $post_activation_multiplier,
-            $post_activation_bias, $channels_in,
-            $dilation_height_factor, $dilation_width_factor,
-            $fused_activation_function, $pad_values, $padding,
-            $stride_height, $stride_width),
+def : Pat<(TF_LceBconv2dOp:$op
+              $input,
+              $filter,
+              $post_activation_multiplier,
```
AdamHillier, line 33: I should add that this PR also standardises all the table-gen patterns to use one line per attribute, which should hopefully make future rebases/additions easier.

larq_compute_engine/mlir/transforms/optimize.cc
```diff
@@ -130,23 +213,117 @@ struct SetBitpackedActivations : public OpRewritePattern<BinaryOp> {
...
+        const auto filter_shape = filter_type.getShape();
+        const std::int32_t backtransform_add =
+            filter_shape[1] * filter_shape[2] * filter_shape[3];
+        const auto clamp_min_max =
+            llvm::StringSwitch<std::pair<std::int32_t, std::int32_t>>(
+                inner_bconv_op.fused_activation_function())
+                .Case("RELU", {0, backtransform_add})
+                .Case("RELU_N1_TO_1", {-1, 1})
+                .Case("RELU6", {0, 6})
+                .Default({-1 * backtransform_add, backtransform_add});
```
AdamHillier, line 243: I've used this rather than the minimum/maximum representable Int32 value because I was for some reason getting some overflow errors and ended up with `clamp_min = clamp_max = min_int32`. This is valid though, as the xor-popcount accumulator will never be outside this range.

larq_compute_engine/mlir/transforms/optimize.cc
```diff
@@ -130,23 +213,117 @@ struct SetBitpackedActivations : public OpRewritePattern<BinaryOp> {
...
+            rewriter.getIntegerAttr(rewriter.getIntegerType(32),
+                                    inner_bconv_op.dilation_width_factor()),
+            rewriter.getStringAttr(inner_bconv_op.fused_activation_function()),
+            rewriter.getIntegerAttr(rewriter.getIntegerType(32),
+                                    inner_bconv_op.pad_values()),
+            rewriter.getStringAttr(inner_bconv_op.padding()),
+            rewriter.getIntegerAttr(rewriter.getIntegerType(32),
+                                    inner_bconv_op.stride_height()),
+            rewriter.getIntegerAttr(rewriter.getIntegerType(32),
+                                    inner_bconv_op.stride_width()));
```
AdamHillier, line 314: Unfortunately this rather explicit operand/attribute format was necessary to get non-`opt` builds to work properly - the previous method with `ValueRange` lists of operands didn't work. It's a tad uglier, but it works.

AdamHillier:
Thanks for taking a look at this :)

> I like that the weight bitpacking is now it's own pass, but how do you feel about moving the bitpacked activation transforms to this pass as well which would reduce the size of the optimize pass? 

In principle I agree that it would be nicer to slim down the optimise pass, but the reason I moved bitpacking weights into its own pass in the first place was because it was interfering with the activation bitpacking pass (my attempts at getting `addBenefit` to work with the C++ pass were fruitless). If they were in the same pass I think the same issue would re-emerge. Would you rather have activation bitpacking in its own pass that runs after optimise but before weight bitpacking?

> How do you feel about doing the refactor and splitting out of the bitpack weights pass in a separate PR to keep the diff managable? If it is too much work to split it out I am happy to keep it like this but due to the refactor and tablegen formatting it is hard to review the relevant changes in-depth.
 
I can definitely do that :+1: I'll move the changes into a new PR, though that'll have to be merged before this one can be.

lgeiger:
> In principle I agree that it would be nicer to slim down the optimise pass, but the reason I moved bitpacking weights into its own pass in the first place was because it was interfering with the activation bitpacking pass (my attempts at getting addBenefit to work with the C++ pass were fruitless). If they were in the same pass I think the same issue would re-emerge. Would you rather have activation bitpacking in its own pass that runs after optimise but before weight bitpacking?

Ah OK, I see.

> I can definitely do that 👍 I'll move the changes into a new PR, though that'll have to be merged before this one can be.

Thanks I will review this ASAP

AdamHillier:
Apologies for the ugly diff. I've now rebased both this and the target branch on top of master (which includes #393) so hopefully it'll be slightly better now.

larq_compute_engine/mlir/transforms/optimize.cc
```diff
@@ -34,34 +39,152 @@ bool IsConstantValue(Attribute values, float expected_value) {
...
-    results.push_back(APFloat(result));
-  }
+/**
+ * =================================================
+ * Computing thresholds for writing bitpacked output
+ * =================================================
+ *
+ * Consider a single output element of a binary convolution, y. We have that:
+ *
+ *                       y = clamp(a - 2x, c, C) * μ + β
```
lgeiger, line 51: This is a super nit-picky suggestion, but do you mind changing mu to gamma to make it consistent with the formulas in the original batch norm paper? I can't help it but mu, always makes me think of a mean instead of a multiplier.

larq_compute_engine/mlir/transforms/optimize_patterns.td
```diff
@@ -12,96 +12,148 @@ def HasOneUse : Constraint<CPred<"$0.hasOneUse()">>;
...
-            TFL_AF_None, $pad_values, $padding,
-            $stride_height, $stride_width),
-          (TF_LceBconv2dOp $input,
-            (TFL_MulOp (ConstantOp $filter), (ConstantOp (ComputeBSignAndExpandTo4D $post_activation_multiplier)), TFL_AF_None),
-            (TFL_AbsOp (ConstantOp $post_activation_multiplier)),
-            $post_activation_bias, $channels_in,
-            $dilation_height_factor, $dilation_width_factor,
-            TFL_AF_None, $pad_values, $padding,
-            $stride_height, $stride_width),
-          [], (addBenefit 90)>;
```
lgeiger, line 67: Just curious, what is the benefit of moving this transform into the activation bitpacking pattern and not doing this before hand? Is it due to the handling of the fused activation?

larq_compute_engine/mlir/ir/lce_ops.td
```diff
@@ -57,8 +60,9 @@ TODO
   let arguments = (ins
     TensorOf<[F32]>:$input,
     TensorOf<[F32, I32]>:$filter,
-    TensorOf<[F32]>:$post_activation_multiplier,
-    TensorOf<[F32]>:$post_activation_bias,
+    TensorOfOrNone<[F32]>:$post_activation_multiplier,
+    TensorOfOrNone<[F32]>:$post_activation_bias,
+    TensorOfOrNone<[I32]>:$write_bitpacked_threshold,
```
lgeiger, line 65: I am not sure if I like the name `write_bitpacked_threshold`, it seems to suggest that it is a boolean that decides wether to write a bitpacked threshold or not. How do you feel just calling it `threshold`?
@koenhelwegen you are usually good in coming up with names for those things. Do you think `threshold` is clear enough?

This is not a blocker for this PR, since it will be easier to change once it is merged into #387.

AdamHillier: Yeah I agree the name is quite clunky, definitely open to naming suggestions.

larq_compute_engine/mlir/transforms/optimize_patterns.td
AdamHillier: Yeah, the issue is that we need to know inside the activation bitpacking pass which of the weights have been flipped, because for channels with flipped weights we need to invert the clamp min/max for computing the correct threshold.

larq_compute_engine/mlir/transforms/optimize.cc
AdamHillier: Sure, happy to :+1:

larq_compute_engine/mlir/tests/optimize.mlir
```diff
@@ -1,266 +1,293 @@
...
+  %cst = constant unit
+  %0 = "tf.LceBconv2d"(%arg0, %filter, %post_activation_multiplier, %post_activation_bias, %cst) {channels_in = 1 : i32, dilation_height_factor = 1 : i32, dilation_width_factor = 1 : i32, fused_activation_function = "NONE", pad_values = 1 : i32, padding = "SAME", stride_height = 1 : i32, stride_width = 1 : i32} : (tensor<256x32x32x1xf32>, tensor<8x2x2x1xf32>, tensor<8xf32>, tensor<8xf32>, none) -> tensor<256x32x32x8xf32>
+  %1 = "tf.LceBconv2d"(%0, %arg1, %arg2, %arg3, %cst) {channels_in = 8 : i32, dilation_height_factor = 1 : i32, dilation_width_factor = 1 : i32, fused_activation_function = "NONE", pad_values = 1 : i32, padding = "SAME", stride_height = 1 : i32, stride_width = 1 : i32} : (tensor<256x32x32x8xf32>, tensor<8x3x3x8xf32>, tensor<8xf32>, tensor<8xf32>, none) -> tensor<256x32x32x8xf32>
+  return %1 : tensor<256x32x32x8xf32>
+
+  // The new filter values. Check that the values for the first four filters
+  // have had their signs flipped. The syntax is very ugly because filecheck
+  // tries to perform string substitution when it encounters `[[` and so we have
+  // to wrap the whole thing in a regex, and escape the square brackets.
```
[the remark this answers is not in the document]
AdamHillier: Yeah, I'm not sure it's a bug per-se, just clunky, non-intuitive behaviour. I guess we could make a feature request to change it, but it's just an unfortunate conflict between their array syntax and their string substitution syntax.

larq_compute_engine/mlir/transforms/optimize_patterns.td
lgeiger: I see, makes sense :+1:

larq_compute_engine/mlir/ir/lce_ops.td
koenhelwegen: `threshold` sounds reasonable to me, or `quantization_threshold`/`binarization_threshold`/`output_threshold`? But I'm on unfamiliar territory here so not sure what works in this context. Perhaps `bitpacking_threshold` (instead of `bitpacked`) would avoid the confusion @lgeiger raised?

Tombana:
I'd opt for `output_threshold`. (That's also how I named it in the kernel
at some point.)

larq_compute_engine/mlir/ir/lce_ops.td
lgeiger: I think I'd prefer `output_threshold` as well. I only just now realized that the naming decision isn't really that important here since the kernel will just use the 4th operand and doesn't rely on the name at all since it is not an attribute.

> But I'm on unfamiliar territory here so not sure what works in this context.

That's the reason why I asked for your opinion, since I thought this is an attribute name that would be exposed to users :)

PR #393, opened 2020-06-04
Move weight bitpacking into its own converter pass.

AdamHillier:
## What do these changes do?

Weight bitpacking in the converter is moved out of the optimise pass and into its own dedicated pass that runs after the optimise pass.

## How Has This Been Tested?

CI.

## Benchmark Results

N/A.

## Related issue number

This is a prerequisite for #391.

### DIFF
--- a/larq_compute_engine/mlir/transforms/bitpack_weights.cc
+++ b/larq_compute_engine/mlir/transforms/bitpack_weights.cc
@@ -1,0 +1,77 @@
+#include "larq_compute_engine/core/packbits.h"
+#include "larq_compute_engine/mlir/ir/lce_ops.h"
+#include "mlir/Dialect/StandardOps/IR/Ops.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Pass/Pass.h"
+
+namespace mlir {
+namespace TFL {
+
+namespace {
+
+struct BitpackWeightsLCE : public PassWrapper<BitpackWeightsLCE, FunctionPass> {
+  void runOnFunction() override;
+};
+
+bool IsConv2DFilter(Attribute filter) {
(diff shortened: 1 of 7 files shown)


larq_compute_engine/mlir/transforms/bitpack_weights_patterns.td
```diff
@@ -0,0 +1,36 @@
+include "mlir/Dialect/StandardOps/IR/Ops.td"
+include "tensorflow/compiler/mlir/lite/ir/tfl_ops.td"
```
lgeiger, line 2: I don't think we actually need it here
```suggestion
```

larq_compute_engine/mlir/transforms/bitpack_weights.cc
```diff
@@ -0,0 +1,77 @@
...
+  auto func = getFunction();
+
+  TFL::populateWithGenerated(ctx, &patterns);
+  applyPatternsAndFoldGreedily(func, patterns);
+}
+
+}  // namespace
+
+// Creates an instance of the TensorFlow dialect BitpackWeights pass.
+std::unique_ptr<OperationPass<FuncOp>> CreateBitpackWeightsLCEPass() {
```
lgeiger, line 69: I think this needs to be added to [`mlir/transforms/passes.h`](https://github.com/larq/compute-engine/blob/master/larq_compute_engine/mlir/transforms/passes.h) and [`mlir/tf_tfl_passes.cc`](https://github.com/larq/compute-engine/blob/master/larq_compute_engine/mlir/tf_tfl_passes.cc) as well.

AdamHillier: Ah thanks, I hadn't copied them properly from the other PR

PR #399, opened 2020-06-06
Slightly optimize docker image size

lgeiger:
## What do these changes do?
This swiches to the headless version of openjdk to slightly reduce the size of the docker image and installes git and zip so we can also build the AAR package


lgeiger:
CI failure is unrelated due to https://github.com/agronholm/typeguard/issues/134

PR #402, opened 2020-06-09
Fix trailing quantization op with experimental_bitpacking fusion

lgeiger:
## What do these changes do?

This PR replaces the int8 hybrid quantization pass with a manual TabelGen pattern for the BConv and the default TFLite full quantization pass. The old approach left trailing quantize ops in the graph when using bitpacked activations, which can be reproduced by the following toy model:
```python
model = tf.keras.Sequential(
    [
        lq.layers.QuantConv2D(
            32,
            3,
            input_quantizer="ste_sign",
            kernel_quantizer="ste_sign",
            use_bias=False,
            input_shape=(32, 32, 3),
        ),
        tf.keras.layers.BatchNormalization(),
        lq.layers.QuantConv2D(
            32,
            3,
            input_quantizer="ste_sign",
            kernel_quantizer="ste_sign",
            use_bias=False,
        ),
    ]
)
```
I couldn't get this to work with the previous approach, but in any case I think this approach is to be simpler and has less moving parts.

## How Has This Been Tested?
FileCheck tests and manually checked QuickNet.


larq_compute_engine/mlir/tests/quantize.mlir
```diff
@@ -1,294 +1,7 @@
...
-  %4 = "tfl.quantize"(%3) {qtype = tensor<4x!quant.uniform<u8:f32, 1.0>>} : (tensor<4xf32>) -> tensor<4x!quant.uniform<u8:f32, 1.0>>
-
-  return %4 : tensor<4x!quant.uniform<u8:f32, 1.0>>
-
-// CHECK: %[[dq1:.*]] = "tfl.dequantize"(%arg0)
-// CHECK-NEXT: %[[dq2:.*]] = "tfl.dequantize"(%arg1)
-// CHECK-NEXT: %[[pow:.*]] = tfl.pow %[[dq1]], %[[dq2]]
-// CHECK-NEXT: %[[q:.*]] = "tfl.quantize"(%[[pow]])
-// CHECK-NEXT: return %[[q]]
-}
```
lgeiger, line 288: These tests are now unnecessary since quantization for ops other than BConv will be handled by the TFLite quantization pass.

larq_compute_engine/mlir/transforms/quantize_patterns.td
```diff
@@ -33,3 +32,66 @@ def : Pat<(LQ_Bconv2dOp
...
+              $dilation_height_factor,
+              $dilation_width_factor,
+              $fused_activation_function,
+              $pad_values,
+              $padding,
+              $stride_height,
+              $stride_width),
+          [(HasOneUse $input)]>;
+
+def : Pat<(TFL_QuantizeOp
```
lgeiger, line 68: We could merge this pattern into the one handling `post_activation_multiplier` and `post_activation_bias` but I opted to keep them separate to make sure that `post_activation_multiplier` and `post_activation_bias` will always converted correctly even if this pattern fails to match.

larq_compute_engine/mlir/tf_tfl_passes.cc
```diff
@@ -20,7 +20,8 @@ namespace tensorflow {
 void AddQuantizationPasses(const mlir::TFL::QuantizationSpecs& quant_specs,
                            mlir::OpPassManager* pass_manager) {
   pass_manager->addPass(mlir::TFL::CreatePrepareQuantizePass(quant_specs));
-  pass_manager->addPass(mlir::TFL::CreateHybridQuantizePass());
+  pass_manager->addPass(mlir::TFL::CreateLCEQuantizePass());
+  pass_manager->addPass(mlir::TFL::CreateQuantizePass());
```
Tombana, line 24: Just out of curiosity, does the order of `LCEQuantizePass` and `QuantizePass` matter? Since our op is in a different dialect, the `QuantizePass` should leave it alone right?

lgeiger: > Just out of curiosity, does the order of LCEQuantizePass and QuantizePass matter? 

If I do it the other way around I ran into some problems where the multiplier and bias were casted to int8 (since we only remove the quantize dequantize pairs in the lce quantize pass), which now fails the op verification in our dialect.

> Since our op is in a different dialect, the QuantizePass should leave it alone right?

The quantize pass will also run on our op as well, the different dialect is not a problem.

PR #405, opened 2020-06-11
Upgrade TensorFlow to tensorflow@9a70ab8

lgeiger:
## What do these changes do?
This upgrade our TF dependency to https://github.com/tensorflow/tensorflow/commit/9a70ab8813048db1aacd3df5361dae42bf4ea31f

## How Has This Been Tested?
CI


lgeiger:
Looks like the ARM builds are failing since we are missing the `armeabi` config in the `cpuinfo` dependencies. Looks like TensorFlow needs to pull in https://github.com/google/ruy/pull/107 into their build defs as well.

AdamHillier:
What is the error that an optimised build fixes?

lgeiger:
> What is the error that an optimised build fixes?

On CI I ran into:
```
/bin/larq_compute_engine/tests/end2end_test.runfiles/larq_compute_engine/larq_compute_engine/mlir/python/converter.py", line 6, in <module>
    from larq_compute_engine.mlir._graphdef_tfl_flatbuffer import (
ImportError: /home/runner/.cache/bazel/_bazel_runner/3412a614bd4c8f82c864c40f9d166608/execroot/larq_compute_engine/bazel-out/k8-fastbuild/bin/larq_compute_engine/tests/end2end_test.runfiles/larq_compute_engine/larq_compute_engine/mlir/_graphdef_tfl_flatbuffer.so: undefined symbol: _ZN10tensorflow4data12experimental19SnapshotDatasetV2Op18kFileFormatVersionE
```
I couldn't reproduce this on my local machine, so I am just trying to see if the latest commit makes CI happy

.github/workflows/unittests.yml
```diff
@@ -75,9 +75,9 @@ jobs:
       - name: Install pip dependencies
         run: pip install tensorflow==2.1.0 larq~=0.9.4 larq_zoo==1.0.0 pytest tensorflow_datasets==3.1.0 --no-cache-dir
       - name: Run FileCheck tests
-        run: bazelisk test larq_compute_engine/mlir/tests:all --test_output=all --distinct_host_configuration=false $([ -z "$GOOGLE_APPLICATION_CREDENTIALS" ] || echo "--remote_http_cache=https://storage.googleapis.com/plumerai-bazel-cache/lce-ubuntu --google_default_credentials")
+        run: bazelisk test larq_compute_engine/mlir/tests:all --test_output=all --copt=-O2 --distinct_host_configuration=false $([ -z "$GOOGLE_APPLICATION_CREDENTIALS" ] || echo "--remote_http_cache=https://storage.googleapis.com/plumerai-bazel-cache/lce-ubuntu --google_default_credentials")
```
AdamHillier, line 78:
```suggestion
        run: bazelisk test larq_compute_engine/mlir/tests:all --test_output=all --distinct_host_configuration=false $([ -z "$GOOGLE_APPLICATION_CREDENTIALS" ] || echo "--remote_http_cache=https://storage.googleapis.com/plumerai-bazel-cache/lce-ubuntu --google_default_credentials")
```

In the last commit that failed CI before you added in the `--copt=-02` (f8ff469) the filecheck test passed and it was just the end-to-end test that failed, so would you be okay to revert this for the filecheck test?

lgeiger: >In the last commit that failed CI before you added in the --copt=-02 (f8ff469) the filecheck test passed and it was just the end-to-end test that failed, so would you be okay to revert this for the filecheck test?

If I understand bazel caching correclty, we shouldn't do this.
If we revert this flag in the filecheck test the local cache won't be shared between the two commands so the end2end tests would need to rebuild parts of the code that was already compiled for the filecheck tests since `--copt` changed.

AdamHillier: > If I understand bazel caching correclty, we shouldn't do this.
> If we revert this flag in the filecheck test the local cache won't be shared between the two commands so the end2end tests would need to rebuild parts of the code that was already compiled for the filecheck tests since `--copt` changed.

Ah I see, fair enough

PR #415, opened 2020-06-29
Upgrade TensorFlow to 2.3.0rc0

lgeiger:
## What do these changes do?
This PR upgrades our TensorFlow dependency to 2.3.0rc0.

## How Has This Been Tested?
Let's see what CI thinks.

### DIFF
--- a/larq_compute_engine/mlir/tf_tfl_passes.cc
+++ b/larq_compute_engine/mlir/tf_tfl_passes.cc
@@ -21,12 +21,6 @@ void AddQuantizationPasses(const mlir::TFL::QuantizationSpecs& quant_specs,
                            mlir::OpPassManager* pass_manager) {
   pass_manager->addPass(mlir::TFL::CreatePrepareQuantizePass(quant_specs));
   pass_manager->addPass(mlir::TFL::CreateLCEQuantizePass());
-  pass_manager->addPass(mlir::TFL::CreateQuantizePass());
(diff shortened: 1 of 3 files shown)


larq_compute_engine/mlir/tf_tfl_passes.cc
```diff
@@ -21,23 +21,19 @@ void AddQuantizationPasses(const mlir::TFL::QuantizationSpecs& quant_specs,
...
     pass_manager->addPass(mlir::TFL::CreateLCEQuantizePass());
-    pass_manager->addPass(mlir::TFL::CreateQuantizePass());
-    pass_manager->addPass(
-        mlir::TFL::CreatePostQuantizePass(emit_quant_adaptor_ops));
   }
+  pass_manager->addPass(mlir::TFL::CreateQuantizePass());
+  bool emit_quant_adaptor_ops =
+      quant_specs.inference_type != quant_specs.inference_input_type;
+  pass_manager->addPass(
+      mlir::TFL::CreatePostQuantizePass(emit_quant_adaptor_ops));
```
lgeiger, line 36: From https://github.com/tensorflow/tensorflow/commit/22f6939be9cc23d14610c3e21fd30f7345c1b125
> By this order, the default quantization parameter is only applied on the
activations and the weight quantization parameter will use the parameters from
the weight content.

PR #426, opened 2020-07-14
Refactor OutputTransform with template types

Tombana:
## What do these changes do?
This adds an extra template argument to `OutputTransform` to allow for cleaner code in lce-micro.

## How Has This Been Tested?
CI


lgeiger:
I restarted CI, but the ARM tests are still failing. Not sure what's going on here.

lgeiger:
> I'm not sure why this is suddenly popping up so much while it did not pop up before. In any case, I think we can increase the tolerance to 1e-3 here

This sounds good. I just rebased the PR, otherwise we can just increase the tolerance

Issue #124, opened 2019-12-10
Future optimizations


Tombana:
Things we know that might be faster, but that we left for now because it simplifies the code

- [ ] Move the `bitpack_array` C++ loop inside the assembly part. Gives speedup of about 5% for the `bitpack_array` function on Raspberry Pi.
- [ ] Remove the `bitpack_matrix` loop when there's no padding (i.e. `bitpack_matrix` becomes a single `bitpack_array` call).
- [ ] Decouple bitpacking bitwidth and bgemm bitwidth: we can use the 64-bit bitpacking code for 32-bit bgemm code.
- [ ] Cache RUY prepacking for weights

lgeiger:
Closed via #443

PR #443, opened 2020-07-29
Add optimised canonical bitpacking for Aarch64. Closes #435.

### LINKED ISSUE #435: Optimize canonical bitbacking
We've been discussing seperating `BSign` from `BConv` some time ago. In order to do this eventually we need to standardise on canonical bitpacking across our ops. `BConv` on AArch64 internally uses an optimized bitpacking implementation with different ordering, in order to be able to change this to canonical ordering our canonical bitpacking implementation needs to be optimized so we don't loose performance.

@Tombana mentioned previously:
> The non-optimized bitpacking is about `0.17 ms` compared to `0.10 ms` for optimized bitpacking (for the first bconv layer in quicknet). It would bring the total time of quicknet from `18.3` ms to `18.7 ms`. However, @AdamHillier mentioned a de-interleaving ARM instruction that would allow us to get almost all the speed of the optimized bitpacking but get the canonical-order, completely alleviating this problem.

AdamHillier:
## What do these changes do?

This PR removes the optimised non-canonical bitpacking for Aarch64 and replaces it with an optimised canonical bitpacking implementation instead. This means that all bitpacking is now canonical, as discussed in #435. As a result, the bitpacking code is a lot simpler, and a bunch of code templating is removed. 

## How has this been tested?

The 'big' kernel tests for Aarch64 (that don't run on CI) pass locally. The Aarch64 bitpacking tests have been strengthened to test for canonical bit ordering.

## Benchmark results

I have benchmarked current master (the 'Baseline') and this PR on my Android phone (a OnePlus 6). I used the LCE benchmark utility with `num_runs=250` and report the average and the standard-deviation.

#### 1 thread

| Model         |    Baseline   |       PR      |
|---------------|:-------------:|:-------------:|
| QuickNet      | 16.53 +- 0.06 | 16.53 +- 0.06 |
| QuickNetLarge | 24.21 +- 0.16 | 24.20 +- 0.08 |
| QuickNetXL    | 42.26 +- 0.28 | 42.09 +- 0.18 |

#### 4 threads

| Model         |    Baseline   |       PR      |
|---------------|:-------------:|:-------------:|
| QuickNet      | 6.92 +- 0.26  | 6.86 +- 0.54  |
| QuickNetLarge | 9.99 +- 0.23  | 10.01 +- 0.19 |
| QuickNetXL    | 17.01 +- 0.23 | 17.05 +- 0.33 |

These results show that there is no performance regression with this PR.

(I don't currently have my Raspberry Pi to hand so I can't run the benchmarks there, but when I did so for an earlier version of this PR I didn't see any significant difference from the baseline on that device, either.)

## Related issue number

Closes #435.


larq_compute_engine/core/packbits.h
```diff
@@ -382,19 +353,30 @@ inline void packbits_array(const TIn* input_array, const std::size_t n,
...
+template <class TIn, class TOut_>
 inline void packbits_matrix(const TIn* input, const std::size_t input_num_rows,
                             const std::size_t input_num_cols, TOut_* output,
                             const std::int32_t zero_point = 0) {
   // Force the types to be unsigned so that the function can be called on signed
   // types as well
   using TOut = typename std::make_unsigned<TOut_>::type;
 
+#ifdef __aarch64__
+  if (!CE_IS_BIG_ENDIAN && std::is_same<TIn, float>::value &&
```
AdamHillier, line 370: The optimised bitpacking is only correct on little-endian CPUs, hence this test here. That is I guess technically a regression as the previous optimised bitpacking I think did work on big-endian CPUs too (though we didn't test that behaviour).

I'm not sure it's worth bothering to support the faster code path on big-endian CPUs - a lot of effort for minimal gain. Thoughts?

larq_compute_engine/core/packbits.h
```diff
@@ -382,19 +353,30 @@ inline void packbits_array(const TIn* input_array, const std::size_t n,
...
+      std::is_same<TOut, std::uint64_t>::value && input_num_cols % 64 == 0 &&
+      zero_point == 0) {
+    packbits_aarch64_64(reinterpret_cast<const float*>(input),
+                        input_num_rows * (input_num_cols / 64),
+                        reinterpret_cast<std::uint64_t*>(output));
+    return;
+  }
+#endif
```
AdamHillier, line 378: One key difference is that previously the optimised assembly bitpacking code bitpacked 64 floats into a single uint64, and was invoked multiple times from nested C++ loops. Instead, in this PR the optimised assembly bitpacking code bitpacks a (long, multiple of 64) sequence of floats into a sequence of uint64s. Doing this makes it easier to schedule the memory accesses for the next block (loads into registers) optimally alongside the computation for the current block.

However, there is technically a regression here in that previously an attempt to bitpack a tensor of 129 channels would use optimised assembly for the first 128 channels, and then slower reference code for the single remaining channel. In this PR, such a tensor would be bitpacked only by reference code. I'm not sure how much of a problem that is - most models use multiples of 64 input channels pretty much always (e.g. all the QuickNet family of models).

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -849,45 +842,45 @@ TfLiteStatus Eval(TfLiteContext* context, TfLiteNode* node) {
...
+TfLiteRegistration* Register_BCONV_2D_32_OPT() {
   static TfLiteRegistration r = {
       bconv2d::Init, bconv2d::Free,
       bconv2d::Prepare<bconv2d::KernelType::kRuyOptimized, 32>,
       bconv2d::Eval<bconv2d::KernelType::kRuyOptimized>};
   return &r;
 }
 
-TfLiteRegistration* Register_BCONV_2D64_OPT() {
+TfLiteRegistration* Register_BCONV_2D_64_OPT() {
```
AdamHillier, line 861: I'm sneaking renaming the registrations into this PR...

larq_compute_engine/core/packbits.h
lgeiger: > I'm not sure it's worth bothering to support the faster code path on big-endian CPUs - a lot of effort for minimal gain. Thoughts?

I agree, but I wouldn't really know if we are excluding a major device here either. It looks like [Android is always little-endian](https://developer.android.com/ndk/guides/abis) and I think the Raspberry PIs are too, so I think we should be fine.

Tombana: I guess all that matters is that the weights and activations are bitpacked using the same code (regardless of what order that is). If we'd ever want to support big-endian CPUs, we would just add a `Prepare` step to unpack and repack the weights, then they'd always be in the native order of that platform, right?
I'm not sure if the old code would've worked for big-endian CPUs either.

larq_compute_engine/core/packbits.h
Tombana: How about we do this optimization more generally, regardless of aarch64: if `num_cols % bitwidth == 0` then call the packbits function with just one set of blocks. Else, call the packbits function in a loop. Then the aarch64 optimized code would be used for both.
(I had wanted to do that long ago: https://github.com/larq/compute-engine/issues/124 )

larq_compute_engine/core/packbits.h
```diff
@@ -6,6 +6,7 @@
 #include <cstring>
 #include <limits>
 
+#include "larq_compute_engine/core/macros.h"
```
Tombana, line 9: The current `macros.h` file, which we weren't using anywhere, uses a very bad version of determining endianness, by calling some networking-related function. This not only will not compile on microcontroller platforms, it might also incur actual runtime overhead to do that check. 
I therefore suggest we update `macros.h` by using `FLATBUFFERS_LITTLEENDIAN` or so instead, since all our projects always have flatbuffers as a dependency anyway.
This has nothing to do with this PR by the way, but I realized that adding this include file would break our microcontroller builds.

AdamHillier: Ah interesting, I thought this was a compile-time constant that would get optimised away. I will update it, thanks :+1:

larq_compute_engine/core/types.h
```diff
@@ -10,8 +10,6 @@ enum class FilterFormat { Unknown, HWIO, OHWI, OHWI_PACKED };
 // defines the operating dimension
 enum class Axis { RowWise, ColWise };
 
-enum class BitpackOrder { Canonical, Optimized };
-
```
Tombana, line 14: Since we're refactoring this anyway, we can maybe also remove the `Axis` parameter, since we only ever use one of them I think. How do you feel about this?

AdamHillier: Yep, that makes sense to me :+1:

Tombana:
Now that all the ordering is canonical, bitpacking into 2 `uint32`s is the same as bitpacking into 1 `uint64`. So if we have a matrix with 2 rows and 32 elements per row, we can just bitpack this as a single `uint64`.
I propose:
- the `packbits_matrix` function needs the `TBitpacked` type to distinguish block sizes 32 and 64 in order to properly do padding. It will then either call the `packbits_array` function many times (in case there's padding at each row) or just once: this is an optimization Adam already implemented here.
- However the `packbits_array` function (whether it was called on the flat array or on individual rows) should just ignore the `TBitpacked` type almost completely: it should just bitpack in whatever the most efficient way is, which might be blocks of size 128, even if `TBitpacked` was `uint8`. The only thing it needs `TBitpacked` for, is if there's padding, to know how much to pad.
- This could simplify a lot of code: instead of having those awfully long generic `pack_bitfield` functions, we could just have the 32-bit or 64-bit generic function and use only that one.

We can do this either in this PR or in a separate one, but I think it would be good to do. What do you think @AdamHillier ?

AdamHillier:
> We can do this either in this PR or in a separate one, but I think it would be good to do. What do you think @AdamHillier ?
> 
> EDIT: I just saw #446, that makes it even better: we can remove `TBitpacked` completely (and still use any blocksize like 64 (or 128 if we want) in the `packbits_array` part).

Haha yeah your comment here reminded me that I needed to write-up #446. But yeah, I definitely agree that there's more to optimise here, but I think we can do that as part of the process of standardising 32-bit bitpacking.

larq_compute_engine/core/tests/packbits_aarch64_tests.cc
```diff
@@ -15,17 +15,10 @@ using namespace compute_engine::core;
 template <void (*packing_func)(const float*, std::size_t, std::uint64_t*),
           std::size_t blocksize>
 void test_bitpacking() {
-  constexpr std::size_t n = 512;
```
Tombana, line 18: Lets make this `64 * 15` or so, so that we also test the "`n=3` blocks left" part of the assembly code.

larq_compute_engine/core/packbits.h
```diff
@@ -6,11 +6,15 @@
...
```
Tombana, line 9: This was already discussed, but the `macros.h` file can no be entirely removed I think.

larq_compute_engine/core/packbits.h
```diff
@@ -369,46 +342,69 @@ inline void packbits_array(const TIn* input_array, const std::size_t n,
...
   const TIn* input_ptr = input;
   TOut* output_ptr = reinterpret_cast<TOut*>(output);
-  for (size_t row_index = 0; row_index < input_num_rows; ++row_index) {
-    packbits_array<bitpack_order>(input_ptr, input_num_cols, output_ptr,
-                                  zero_point);
-    input_ptr += input_num_cols;
-    output_ptr += output_num_cols;
+
+  if (input_num_cols % bitwidth == 0) {
+    // If each column can be bitpacked without any padding, then we can treat
```
AdamHillier, line 392:
```suggestion
    // If each row can be bitpacked without any padding, then we can treat
```

larq_compute_engine/core/packbits_aarch64.h
```diff
@@ -10,53 +10,299 @@
...
+      // Now, we can combine the two unzip results for each register pair, with
+      // 'shift right and insert' SRI instructions.
+      //
+      // Continuing with the example from above, after the first SRI instruction
+      // below, v0 will contain:
+      //                                v0
+      //                     [0/1  2/3  4/5 ... 30/31]
+      //
+      // Where `0/1` is a byte that has the high-bit equal to the high-bit of
+      // byte 0, and the second highest bit equal to the high-bit of byte 1.
```
Tombana, line 134: I think the highest bit would be the high-bit of byte 1, and the second highest bit would be the high-bit of byte 0:

```
bit pos:  7  6 5 4 3 2 1 0
value:   s1 s0 _ _ _ _ _ _
```
where the `_` are irrelevant dummy bits.
They'll keep getting right-shifted (i.e. in the direction of bit-position 0) until `s0` is at the least-significant-bit.

larq_compute_engine/core/tests/packbits_aarch64_tests.cc
```diff
@@ -35,30 +28,35 @@ void test_bitpacking() {
...
+TEST(BitpackingAarch64, 2x64) { test_bitpacking<packbits_aarch64_64, 64, 2>(); }
+TEST(BitpackingAarch64, 3x64) { test_bitpacking<packbits_aarch64_64, 64, 3>(); }
+TEST(BitpackingAarch64, 4x64) { test_bitpacking<packbits_aarch64_64, 64, 4>(); }
+TEST(BitpackingAarch64, 8x64) { test_bitpacking<packbits_aarch64_64, 64, 8>(); }
+TEST(BitpackingAarch64, 17x64) {
+  test_bitpacking<packbits_aarch64_64, 64, 17>();
+}
+TEST(BitpackingAarch64, 23x64) {
+  test_bitpacking<packbits_aarch64_64, 64, 23>();
 }
```
Tombana, line 60: :+1: Great

larq_compute_engine/core/packbits_aarch64.h
AdamHillier: Ah yes, you're right, thanks for spotting that. Will update :+1:

PR #449, opened 2020-08-03
Support full int8 QAT models in converter

lgeiger:
## What do these changes do?

This PR adds support for full integer quantization via the converter with the ability to customize input and output types. This adapts https://github.com/tensorflow/tensorflow/commit/ec94fabb838b0934eeaf46f1fd22c36b2b3ec333 

## How Has This Been Tested?
Added unittest to verify input and output dtypes.


larq_compute_engine/mlir/BUILD
```diff
@@ -293,14 +293,20 @@ pybind_extension(
 
 py_library(
     name = "converter",
-    srcs = ["python/converter.py"],
+    srcs = [
+        "python/converter.py",
+        "python/schema_py_generated.py",
```
lgeiger, line 298: I added an autogenerated tflite flatbuffer schema file since it is not included in the pip package of earlier versions of TensorFlow and is also missing on the macos builds of TensorFlow.

I tried using https://github.com/jackwish/tflite as well, but it doesn't include the object-based flatbuffer API for mutations and adding the API to the package would conflict with their import structure.

larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
```diff
@@ -82,7 +82,6 @@ pybind11::bytes ConvertGraphDefToTFLiteFlatBuffer(
       quant_specs.input_ranges.push_back({-128.0, 127.0});
     }
     if (!default_ranges.is_none()) {
-      quant_specs.inference_input_type = tensorflow::DT_QINT8;
```
lgeiger, line 85: This is now properly handled by post-processing the flatbuffer.

larq_compute_engine/mlir/python/util.py
```diff
@@ -0,0 +1,201 @@
...
+        for op in output_dequant_ops:
+            subgraph.outputs[subgraph.outputs == op.outputs[0]] = op.inputs[0]
+            remove_tensors_idxs.add(op.outputs[0])
+            operators.remove(op)
+
+    # Remove tensors marked for deletion.
+    _remove_tensors_from_model(model, remove_tensors_idxs)
+
+    # Convert the model to a bytearray
+    return _convert_model_from_object_to_bytearray(model)
```
lgeiger, line 201: This is the only function relevant for review in this file and was adapted from [here](https://github.com/tensorflow/tensorflow/commit/ec94fabb838b0934eeaf46f1fd22c36b2b3ec333#diff-84a6aaef794e9a9191371241d009e3bbR685-R780). Main changes are use of native `tf.dtypes` and support for `int8` only.

larq_compute_engine/tests/end2end_test.py
```diff
@@ -143,6 +143,7 @@ def toy_model_int8(**kwargs):
     )(x)
     x = quant(x)
     x = tf.keras.layers.GlobalAveragePooling2D()(x)
+    x = quant(x)
```
lgeiger, line 146: This fake quant layer was preventing full int8 output previously.

setup.py
```diff
@@ -30,7 +30,7 @@ def has_ext_modules(self):
     packages=find_packages(),
     ext_modules=ext_modules,
     url="https://larq.dev/",
-    install_requires=["packaging>=19"],
+    install_requires=["flatbuffers>=1.12", "packaging>=19"],
```
lgeiger, line 33: We are already depending implicetly on `flatbuffers` through TensorFlow.

larq_compute_engine/tests/end2end_test.py
```diff
@@ -192,5 +193,24 @@ def test_simple_model(model_cls):
...
+@pytest.mark.parametrize("inference_input_type", [tf.int8, tf.float32])
+@pytest.mark.parametrize("inference_output_type", [tf.int8, tf.float32])
+def test_int8_input_output(model_cls, inference_input_type, inference_output_type):
+    model_lce = convert_keras_model(
+        model_cls(),
+        inference_input_type=inference_input_type,
+        inference_output_type=inference_output_type,
+        experimental_default_int8_range=(-6.0, 6.0) if model_cls == toy_model else None,
+    )
+    interpreter = tf.lite.Interpreter(model_content=model_lce)
```
lgeiger, line 206: I don't allocate any tensors here and just use the interpretor to get input and output types, thus we don't need to register our custom ops here.

larq_compute_engine/mlir/python/converter.py
```diff
@@ -52,6 +53,8 @@ def _contains_training_quant_op(graph_def):
 def convert_keras_model(
     model: tf.keras.Model,
     *,  # Require remaining arguments to be keyword-only.
+    inference_input_type: tf.DType = tf.float32,
+    inference_output_type: tf.DType = tf.float32,
```
AdamHillier, line 57: I don't know what they do in TFL, but it seems that we could probably drop the `inference_` from the argument names here? I don't really have a strong preference though.

larq_compute_engine/mlir/python/util.py
AdamHillier: I think it might be nice to keep the [logging line](https://github.com/tensorflow/tensorflow/commit/ec94fabb838b0934eeaf46f1fd22c36b2b3ec333#diff-84a6aaef794e9a9191371241d009e3bbR719-R722) from the commit in TF, for ease of debugging potential Int8 issues. How do you feel about that?

setup.py
AdamHillier: If we're depending on it implicitly, why do we have to include it here? Is there something special about `>=1.12`?

larq_compute_engine/mlir/python/converter.py
lgeiger: I don't have a strong preference either, but I went for the `inference_` prefix since [TFLite does the same](https://www.tensorflow.org/api_docs/python/tf/lite/TFLiteConverter#attributes), but I'd be happy to just use `input_type` and `output_type` instead since we also don't adhere to the type conventione the TFLite converter uses here.

Tombana: I also don't have a preference. This part of the code is where training connects to inference-only, so writing `inference_` sort-of shows that its meant for the final inference-only graph and not for the training graph. That might be the rationale behind it.

larq_compute_engine/mlir/python/util.py
lgeiger: I'm not sure if this logging line is relevant for us, since we only support `tf.float32` and `tf.int8` and don't have the complex handling of the different quantization code paths.

setup.py
lgeiger: > If we're depending on it implicitly, why do we have to include it here?

Actually my comment here was wrong and the current stable versions of TensorFlow don't depend on `flatbuffers` yet.

>  Is there something special about `>=1.12`?

This is the version [TensorFlow 2.4](https://github.com/tensorflow/tensorflow/blob/572d7a71e3fbd202148e231e5a18d89ec411a046/tensorflow/tools/pip_package/setup.py#L57) will depend on, so I think it is nice to have the same version constraints.

PR #456, opened 2020-08-06
Automatically upload benchmark binaries to new releases

lgeiger:
## What do these changes do?
This PR will upload the build benchmark to the published release.

## How Has This Been Tested?
This change hasn't been tested since it would require pushing a new release. Since this is the last build job of the pipeline even a failing job wouldn't break our release build.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,6 +51,36 @@ jobs:
         with:
           name: Benchmark-Binaries
           path: benchmark-binaries
+      - name: Upload Release Asset AArch64
+        if: github.event_name == 'release'
+        uses: actions/upload-release-asset@v1
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -51,6 +51,36 @@ jobs:
         with:
           name: Benchmark-Binaries
           path: benchmark-binaries
+      - name: Upload Release Asset AArch64
+        if: github.event_name == 'release'
```
Tombana, line 55: Isn't the fact that this is in `release.yml` enough to only be triggered on releases?

lgeiger: This workflow can also be triggered by push, so this check prevent's the step from running:
https://github.com/larq/compute-engine/blob/1756d2969d17ccd68a711a742e331c18d3eea3b0/.github/workflows/release.yml#L8-L10

Issue #446, opened 2020-07-31
Standardise on 32-bit bitpacking


AdamHillier:
At the moment, we perform either 64-bit bitpacking or 32-bit bitpacking. Once #443 is merged, the only difference between the two will be whether we are bitpacking into uint32s or uint64s, i.e. whether we pad the number of channels to a multiple of 64 or a multiple of 32.

Since #351 was merged, both 64-bit bitpacked bconvs and 32-bit bitpacked bconvs have used the exact same assembly kernel for execution on Aarch64. This means that there is no performance benefit anymore from doing 64-bit bitpacking, other than the difference in latency for the bitpacking process itself. However, the optimised 64-bit bitpacking code introduced in #443 can easily be adapted to do 32-bit bitpacking instead with no performance impact.

Standardising on uint32 bitpacking will reduce code complexity and templating quite significantly, and should yield performance improvements when the number of input channels is a multiple of 32 rather than 64 (in the most extreme case, a 3x3 bconv with 32 input channels should run approximately twice as quickly).

PR #457, opened 2020-08-06
Split BSign into Quantize/Dequantize and separate from BConv

### LINKED ISSUE #447: Separate our 'bsign' and 'bconv' ops
At the moment, our `bconv` op can read either float or int8 or bitpacked inputs, and can write either float or int8 or bitpacked outputs. When the input is not bitpacked, the first step that the op performs is calling a bitpacking procedure. Conceptually, this is almost like a 'sub-op', which doesn't need to be the responsibility of the bconv. We could instead make the `bconv` op accept only bitpacked inputs, and insert dedicated `bsign` ops where needed that take float or int8 inputs and produce bitpacked ouptuts. A `bsign` op already exists in our code base, but we generally fuse it into a subsequent `bconv` and so don't use it directly.

This would have several advantages:
* Simplifying the `bconv` op implementation. This would reduce nine combinations of input/output types to just three, improve the 'separation of concerns' in our ops, and improve codebase readability and approachability. 
* Reducing binary size. This is an extension of the above point - there would be significantly less code templating. 
* Make certain optimisations in the converter easier and cleaner to implement. We would be able to treat the `bsign` op similar to a TFL cast or quantise op. This should make things like bitpacked activation optimisations (for both `bconv` and `bmaxpool`) easier, because the rewrite patterns can likely then be done in TableGen rather than C++.

A prerequesite for doing this is #446, so this is blocked for now.

lgeiger:
## What do these changes do?
This PR implements "Option 1" of #447 by moving the bitpacking out of the BConv and into separate quantize/dequantize ops.

## TODO:
- [x] Adapt converter
- [ ] Rebase ontop of #446
- [ ] Kernel changes
   - [ ] Change `bsign` op to `quantize`
   - [ ] Implement `dequantize` op
   - [ ] Remove bitpacking from `bconv`
   - [ ] Remove bitpacking from `bmaxpool`
   - [ ] Adapt TFLite unittests

## How Has This Been Tested?
Unittests and manually verified that QuickNet and our toy int8 model convert correctly.

## Benchmark Results
TODO

## Related issue number
Closes #447


larq_compute_engine/tflite/kernels/bmaxpool.cc
```diff
@@ -122,30 +78,10 @@ TfLiteStatus Eval(TfLiteContext* context, TfLiteNode* node) {
...
-  } else {
-    TF_LITE_ENSURE_EQ(context, input->type, kTfLiteInt32);
-    packed_input_shape.ReplaceWith(4, GetTensorShape(input).DimsData());
-    packed_input_data = GetTensorData<TBitpacked>(input);
-  }
+  packed_input_shape.ReplaceWith(4, GetTensorShape(input).DimsData());
+  const TBitpacked* packed_input_data = GetTensorData<TBitpacked>(input);
 
   ce::ref::BMaxPool(*poolparams, packed_input_shape, packed_input_data,
                     GetTensorShape(output), GetTensorData<TBitpacked>(output));
```
Tombana, line 87: Now you can even remove the lines with `packed_input_shape` and `packed_input_data` (above the lines in this suggestion) entirely :smiley: 
```suggestion
  ce::ref::BMaxPool(*poolparams, GetTensorShape(input), GetTensorData<TBitpacked>(input),
                    GetTensorShape(output), GetTensorData<TBitpacked>(output));
```

lgeiger: Good point, that's a lot nicer :+1:

PR #461, opened 2020-08-13
Standardise on 32-bit bitpacking. Closes #446.

### LINKED ISSUE #446: Standardise on 32-bit bitpacking

AdamHillier:
## What do these changes do?

In LCE at the moment we support two different bitpacking types: uint32 and uint64. We use uint64 wherever possible, and uint32 otherwise. This used to make sense because we previously used a special optimised non-canonical 64-bit bitpacking procedure that was very fast on Aarch64. However, #443 removed non-canonical bitpacking (without performance loss), which laid the groundwork for switching completely from 64-bit bitpacking to 32-bit bitpacking. For the motivation for standardising on 32-bit bitpacking, see #446.

I'm very sorry for the size of this PR - it touches almost every file in the repo - but at least it removes twice as much code as it adds!

## Detailed changes

Concretely, this PR standardises on int32 - *not* uint32 - bitpacking. This is because TFLite doesn't have an unsigned 32-bit tensor type, but does have a signed 32-bit tensor type. Aligning with the TFLite tensor type makes some things such as bitpacked activations easier, and as we only perform bitwise operations on the bitpacked data, the sign of the datatype doesn't matter.

One nice benefit is that we now have only two opt registrations: `BCONV_2D_REF` and `BCONV_2D_OPT`.

Most of the code that's removed is function/struct templating that was previously necessary to support multiple bitpacking types. It was previously common to have a templated `TBitpacked` type parameter; this PR removes such templating, and instead defines a single, global `TBitpacked` in `types.h`, that is then imported anywhere the type definition is needed. This approach means that it should still be possible to change the bitpacking type in the future without too much trouble.

Additionally, a lot of code has been either simplified or stripped out if no longer relevant. This includes:
* The bitpacking code itself has been heavily simplified, including the bitpacking tests (which have mostly been re-written).
* Functions in `bconv2d.cc` that select the bitpacking type have been removed.
* Lots of assertions in the various kernel implementations that the LHS and RHS types had to be the same, et cetera (without datatype templating this doesn't make sense) have been removed.
* Code for handling 32-bit bitpacked data with uint64 optimised kernels has been removed.
* The optimised assembly kernels have been modified to accept int32 rather than uint64 inputs (this change is relatively minor; it mostly involves adjusting matrix depth calculations).
* Perhaps surprisingly, the `bgemm_functor` code (used in our reference kernel) and associated tests have also been heavily simplified, because there were previously function overloads for many different datatypes, which is now unnecessary.

## How Has This Been Tested?

The 'big' kernel tests that don't run on CI pass locally for Aarch64 and Arm32. The bitpacking tests have been re-written, as have the bgemm functor tests.

## Benchmark Results

This PR itself introduces a performance regression, because although our bitpacking and bgemm are as fast as ever, the Ruy 'packing' step is slower when the datatype is int32 (as opposed to uint64).

However, this PR should be viewed in conjunction with #462, which builds upon this PR to introduce optimised Ruy packing for int32 datatypes and in fact improves benchmark performance. Full benchmark results are in that PR (comparing against the master branch prior to this PR being merged).

## Related issue number

Closes #446.

### DIFF
--- a/larq_compute_engine/core/tests/BUILD
+++ b/larq_compute_engine/core/tests/BUILD
@@ -6,7 +6,6 @@ cc_test(
     name = "bgemm_tests",
     size = "small",
     srcs = ["bgemm_tests.cc"],
-    copts = ["-Iext/gtest/include"],
     deps = [
         "//larq_compute_engine/core:bgemm_functor",
         "@com_google_googletest//:gtest_main",
@@ -17,9 +16,16 @@ cc_test(
     name = "packbits_tests",
     size = "small",
     srcs = ["packbits_tests.cc"],
-    copts = ["-Iext/gtest/include"],
+    linkopts = select({
+        "@org_tensorflow//tensorflow:windows": [],
+        "//conditions:default": [
+            "-lm",
+            "-lrt",
+        ],
+    }),
     deps = [
         "//larq_compute_engine/core:packbits",
+        "@com_google_absl//absl/strings",
         "@com_google_googletest//:gtest_main",
     ],
 )
@@ -38,7 +44,6 @@ cc_test(
     name = "packbits_aarch64_tests",
     size = "small",
     srcs = ["packbits_aarch64_tests.cc"],
-    copts = ["-Iexternal/gtest/include"],
     deps = [
         "//larq_compute_engine/core:packbits",
         "@com_google_googletest//:gtest_main",
--- a/larq_compute_engine/core/tests/packbits_tests.cc
+++ b/larq_compute_engine/core/tests/packbits_tests.cc
@@ -1,101 +1,109 @@
-#include <gmock/gmock.h>
+#include <gtest/gtest.h>
 
 #include <array>
 #include <cstdint>
+#include <random>
 #include <vector>
 
+#include "absl/strings/str_cat.h"
 #include "larq_compute_engine/core/packbits.h"
-
-// From @flatbuffers, used for the FLATBUFFERS_LITTLEENDIAN macro
-#include "flatbuffers/base.h"
 
 namespace compute_engine {
 namespace testing {
 
 namespace ce = compute_engine;
 
-// tests with non-uniform data verify the correctness of the bitpacking
-// operation itself while the uniform tests are used to verify the bitpacking
-// with different bitwidths
-template <class TIn>
-void test_bitpacking_nonuniform_input() {
-  // input matrix (row-major memory laytout)
-  const std::size_t num_rows = 2, num_cols = 8;
-  std::array<TIn, 16> input{1, 1,  -1, 1,  -1, -1, -1, 1,
-                            1, -1, 1,  -1, -1, -1, 1,  1};
+using ce::core::bitpacking_bitwidth;
+using ce::core::TBitpacked;
 
-  // expected output matrix after bitpacking
-  const std::size_t expected_num_rows = 2, expected_num_cols = 1;
-  std::vector<std::uint8_t> expected;
-  if (!FLATBUFFERS_LITTLEENDIAN)
-    expected = {0b00101110, 0b01011100};
-  else
-    expected = {0b01110100, 0b00111010};
+class BitpackingTest
+    : public ::testing::TestWithParam<std::tuple<int, int, std::int32_t>> {};
 
-  std::vector<std::uint8_t> output(
-      ce::core::GetPackedMatrixSize<std::uint8_t>(num_rows, num_cols));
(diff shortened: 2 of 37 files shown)


larq_compute_engine/core/types.h
```diff
@@ -1,9 +1,18 @@
...
+#include <type_traits>
+
 namespace compute_engine {
 namespace core {
 
+// Define these once here, so they can be included everywhere.
+using TBitpacked = std::int32_t;
+constexpr std::size_t bitpacking_bitwidth =
+    std::numeric_limits<typename std::make_unsigned<TBitpacked>::type>::digits;
+
```
AdamHillier, line 15: Just to highlight for reviewers, this is the most important addition, where `TBitpacked` and `bitpacking_bitwidth` are defined.

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -213,10 +212,9 @@ TfLiteStatus Prepare(KernelType kernel_type,
       return kTfLiteError;
     }
 
-    // TODO: more intelligent selection of the parameters `bitpacking_bitwidth`
-    //       and `bitpack_before_im2col` based on benchmarking results
-    //       (as in https://github.com/larq/compute-engine/issues/290).
-    conv_params->bitpacking_bitwidth = default_bitpacking_bitwidth;
```
lgeiger, line 219: Should we also remove `bitpacking_bitwidth` from `TfLiteBConv2DParams`?

https://github.com/larq/compute-engine/blob/2a4d5132104cf4dea3debb44eed96b76d1c40531/larq_compute_engine/tflite/kernels/bconv2d_params.h#L80

AdamHillier: Good spot, thanks :+1:

larq_compute_engine/core/tests/BUILD
```diff
@@ -17,9 +16,16 @@ cc_test(
...
     size = "small",
     srcs = ["packbits_tests.cc"],
-    copts = ["-Iext/gtest/include"],
+    linkopts = select({
+        "@org_tensorflow//tensorflow:windows": [],
+        "//conditions:default": [
+            "-lm",
+            "-lrt",
+        ],
+    }),
```
AdamHillier, line 25: For some reason I had to add this in order for the (re-written) test to compile, I think because of the uniform random sampling? It's very weird. I don't understand why on earth you'd have to include linker arguments in order to successfully compile something from the C++ standard library, that seems unnecessary.

larq_compute_engine/core/tests/packbits_tests.cc
```diff
@@ -1,101 +1,110 @@
...
+  for (auto i = 0; i < num_rows; i++) {
+    // First test all the 'true' values...
+    for (auto j = 0; j < bitpacking_bitwidth * num_packed_cols; j++) {
+      const bool packed_bit =
+          output_matrix.at(i * num_packed_cols + j / bitpacking_bitwidth) &
+          (TBitpacked(1) << (j % bitpacking_bitwidth));
+
+      if (j < num_cols) {
+        // If this bit position corresponds to an actual value, against the sign
+        // of that value...
```
AdamHillier, line 57:
```suggestion
    for (auto j = 0; j < bitpacking_bitwidth * num_packed_cols; j++) {
      const bool packed_bit =
          output_matrix.at(i * num_packed_cols + j / bitpacking_bitwidth) &
          (TBitpacked(1) << (j % bitpacking_bitwidth));

      if (j < num_cols) {
        // If this bit position corresponds to an actual value, compare against
        // the sign of that value...
```

PR #462, opened 2020-08-13
Improve performance with optimised Ruy matrix packing.

AdamHillier:
## What do these changes do?

This PR introduces an optimised Ruy 'packing' routine for 4x4 column-major kernel layouts. This is the kernel layout that is used for the RHS (input activations) matrix in all our optimised assembly kernels, and the Ruy packing for this layout currently falls back to slow reference code. The new packing routine therefore boosts Aarch64 and Arm32 performance.

This PR also enables caching of the packed LHS (weights) matrix. This is done by enabling the `use_caching` flag within our op, and setting the Ruy cache-policy to 'always cache' for the LHS matrix. Note that the RHS matrix would not benefit from caching because it is not static and changes every inference.

This PR should be viewed (and reviewed) in conjunction with #461, upon which it is re-based. Only the final commit is relevant.

## How Has This Been Tested?

CI.

## Benchmark Results

I have benchmarked and this PR and the baseline on my Android phone (a OnePlus 6) and a Raspberry Pi 4B. I used the LCE benchmark utility with `num_runs=250` and report the average and the standard-deviation.

It's important to note that the baseline used here is current master, prior to #461 (specifically, this commit: 30e9b02eceb).

#### OnePlus 6 (1 thread)

| Model         |    Baseline   |       PR      | Change |
|---------------|:-------------:|:-------------:|:------:|
| QuickNet      | 16.45 +- 0.05 | 14.58 +- 0.06 | -11.4% |
| QuickNetLarge | 24.03 +- 0.08 | 21.73 +- 0.08 | -9.6%  |
| QuickNetXL    | 42.03 +- 0.10 | 37.82 +- 0.08 | -10.0% |

#### OnePlus 6 (4 threads)

| Model         |    Baseline   |       PR      | Change |
|---------------|:-------------:|:-------------:|:------:|
| QuickNet      | 6.91 +- 0.23  | 6.39 +- 0.31  | -7.5%  |
| QuickNetLarge | 10.05 +- 0.38 | 9.47 +- 0.33  | -5.8%  |
| QuickNetXL    | 17.06 +- 0.38 | 16.05 +- 0.39 | -5.9%  |

#### Raspberry Pi 4B (1 thread)

| Model         |    Baseline   |       PR      | Change |
|---------------|:-------------:|:-------------:|:------:|
| QuickNet      | 31.68 +- 0.05 | 30.45 +- 0.05 | -3.9%  |
| QuickNetLarge | 48.25 +- 0.49 | 46.49 +- 0.08 | -3.6%  |
| QuickNetXL    | 83.76 +- 0.12 | 80.47 +- 0.19 | -3.9%  |

#### Raspberry Pi 4B (4 threads)

| Model         |    Baseline   |       PR      | Change |
|---------------|:-------------:|:-------------:|:------:|
| QuickNet      | 15.45 +- 0.05 | 14.94 +- 0.07 | -3.3%  |
| QuickNetLarge | 23.00 +- 0.07 | 22.20 +- 0.06 | -3.5%  |
| QuickNetXL    | 36.92 +- 0.11 | 35.41 +- 0.08 | -4.1%  |

There is a strong performance improvement. It is more pronounced on the OnePlus 6 than the Raspberry Pi, but significant on both. It's more than I was expecting!

For reference, the speed-up seems to be fairly evenly split between the new optimised packing routine and the LHS caching.

## Related issue number

N/A.

### DIFF
--- a/larq_compute_engine/core/bgemm_impl_ruy.h
+++ b/larq_compute_engine/core/bgemm_impl_ruy.h
@@ -10,6 +10,7 @@
 #include "ruy/profiler/instrumentation.h"
 #include "tensorflow/lite/kernels/cpu_backend_context.h"
 #include "tensorflow/lite/kernels/cpu_backend_gemm_params.h"
+#include "tensorflow/lite/kernels/cpu_backend_gemm_ruy.h"
 
 using namespace tflite;
 using namespace tflite::cpu_backend_gemm;
@@ -38,17 +39,16 @@ struct BGemmImplUsingRuy {
 
     // Set up the matrix layouts and mul_params.
     ruy::Matrix<TBitpacked> lhs;
-    ruy::MakeSimpleLayout(lhs_params.rows, lhs_params.cols,
-                          ruy::Order::kRowMajor, lhs.mutable_layout());
     ruy::Matrix<TBitpacked> rhs;
-    ruy::MakeSimpleLayout(rhs_params.rows, rhs_params.cols,
-                          ruy::Order::kColMajor, rhs.mutable_layout());
     ruy::Matrix<DstScalar> dst;
-    ruy::MakeSimpleLayout(dst_params.rows, dst_params.cols,
-                          ruy::Order::kColMajor, dst.mutable_layout());
-    lhs.set_data(lhs_data);
-    rhs.set_data(rhs_data);
-    dst.set_data(dst_data);
+    // We allow these matrices to be cached. Note that this doesn't force them
+    // to be cached; it means that the `cache_policy` of the MatrixParams will
+    // be respected.
+    cpu_backend_gemm::detail::MakeRuyMatrix(lhs_params, lhs_data, &lhs,
+                                            /*use_caching=*/true);
+    cpu_backend_gemm::detail::MakeRuyMatrix(rhs_params, rhs_data, &rhs,
+                                            /*use_caching=*/true);
+    cpu_backend_gemm::detail::MakeRuyMatrix(dst_params, dst_data, &dst);
 
     // We have to make this a `const` matrix because otherwise gcc will try to
     // use the non-const versions of `matrix.data()`
@@ -63,13 +63,8 @@ struct BGemmImplUsingRuy {
 
     constexpr auto BGemmCompiledPaths = ruy::kAllPaths;
 
-    // avoid the reference path for production code
     ruy::Path bgemm_runtime_path = ruy_ctx->SelectPath(ruy::kAllPaths);
 
-    // fallback to standard cpp kernel for all architectures that are not
-    // supported yet.
-    // TODO: this needs to be modified as soon as architecture-specific
-    // optimized kernels are added.
 #if RUY_PLATFORM_ARM
     if (bgemm_runtime_path == ruy::Path::kNeonDotprod)
       bgemm_runtime_path = ruy::Path::kNeon;
@@ -79,12 +74,12 @@ struct BGemmImplUsingRuy {
       bgemm_runtime_path = ruy::Path::kStandardCpp;
 #endif
 
-    // For writing bitpacked output, fallback to the standard C++ kernel.
+    // For now, writing bitpacked output only has a C++ implementation.
     if (std::is_same<DstScalar, TBitpacked>::value) {
       bgemm_runtime_path = ruy::Path::kStandardCpp;
     }
 
-    // For now, int8 only has a C++ implementation.
+    // For now, int8 output only has a C++ implementation.
     if (std::is_same<DstScalar, std::int8_t>::value) {
       bgemm_runtime_path = ruy::Path::kStandardCpp;
     }
--- a/larq_compute_engine/core/bgemm_trmul_params.h
+++ b/larq_compute_engine/core/bgemm_trmul_params.h
@@ -1,10 +1,10 @@
 #ifndef COMPUTE_EGNINE_TFLITE_KERNELS_BGEMM_TRMUL_PARAMS_H_
 #define COMPUTE_EGNINE_TFLITE_KERNELS_BGEMM_TRMUL_PARAMS_H_
 
-#include "bgemm_kernels_ruy.h"
+#include "larq_compute_engine/core/bgemm_kernels_ruy.h"
+#include "larq_compute_engine/core/ruy_pack.h"
 #include "ruy/dispatch.h"
 #include "ruy/mul_params.h"
-#include "ruy/pack_common.h"
 #include "ruy/path.h"
 #include "ruy/trmul_params.h"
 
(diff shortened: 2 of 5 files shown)


AdamHillier:
Note that the baseline benchmark figures I have here for the Raspberry Pi are already a lot better than those reported on our website. I'm not sure why that is, it *might* be because the previous results - used on the website - might have been taken from a benchmark build with the Ruy profiler enabled.

larq_compute_engine/core/ruy_pack.h
```diff
@@ -0,0 +1,137 @@
...
+
+    RUY_DCHECK(IsColMajor(src_matrix.layout));
+    RUY_DCHECK(IsColMajor(packed_matrix->layout));
+    RUY_DCHECK_EQ(start_col % 4, 0);
+
+    TBitpacked zerobuf[4];
+    memset(zerobuf, 0, sizeof(zerobuf));
+
+    for (int block_col = start_col; block_col < end_col; block_col += 4) {
+      int src_stride = src_matrix.layout.stride;
```
lgeiger, line 61: I don't think this matters much, but I guess it's nicer to move the assignment out of the loop:
```suggestion
    int src_stride = src_matrix.layout.stride;
    for (int block_col = start_col; block_col < end_col; block_col += 4) {
```

larq_compute_engine/core/ruy_pack.h
```diff
@@ -0,0 +1,137 @@
...
+          src_inc1 = 0;
+        }
+        if (block_col >= src_matrix.layout.cols - 2) {
+          src_ptr2 = zerobuf;
+          src_inc2 = 0;
+        }
+        if (block_col >= src_matrix.layout.cols - 3) {
+          src_ptr3 = zerobuf;
+          src_inc3 = 0;
+        }
```
lgeiger, line 87: If I understand the code correctly, this condition should always be true.
```suggestion
        src_ptr3 = zerobuf;
        src_inc3 = 0;
```

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -652,6 +652,18 @@ void EvalOpt(TfLiteContext* context, TfLiteNode* node,
...
+  // cache the result of packing the weights matrix, which significantly reduces
+  // benchmark latency, at the cost of slightly increased memory usage. The
+  // increase will be approximately the size of the bitpacked model weights
+  // (plus a small amount of overhead if the kernel layout doesn't divide the
+  // shape of the weights). On our target devices we have many GB of memory
+  // available, and we expect the bitpacked model weights to be at most low
+  // double-digit MB, so this is very unlikely to cause any problems.
+  CpuBackendContext* cpu_backend_context =
+      CpuBackendContext::GetFromContext(context);
+  cpu_backend_context->SetUseCaching(true);
```
Tombana, line 665: I haven't looked into the `SetUseCaching` API, but are we sure this only enables it for the bconv op and not for everything? Since the `context` object is shared, so are we not setting a global flag here?
This is fine with me, I'm just curious.

AdamHillier: That is a good question, you might well be right about that. I'll investigate.

larq_compute_engine/core/bgemm_trmul_params.h
```diff
@@ -35,21 +35,20 @@ void PopulateBinaryTrMulParams(ruy::TrMulParams* params) {
...
   using RhsKernelLayout = typename Kernel::RhsLayout;
 
   params->path = ThePath;
 
-  CreatePackedMatrix<TBitpacked, PackedTBitpacked>(
+  CreatePackedMatrix<TBitpacked, TBitpacked>(
       Side::kLhs, ToKernelLayout<LhsKernelLayout>(), params);
-  CreatePackedMatrix<TBitpacked, PackedTBitpacked>(
+  CreatePackedMatrix<TBitpacked, TBitpacked>(
       Side::kRhs, ToKernelLayout<RhsKernelLayout>(), params);
```
Tombana, line 47: What does this change do? What is normally the 'packed' type of `float` or `int8`, and why is it `TBitpacked` in our case?

larq_compute_engine/core/bgemm_trmul_params.h
```diff
@@ -35,21 +35,20 @@ void PopulateBinaryTrMulParams(ruy::TrMulParams* params) {
...
   params->run_pack[Side::kLhs] =
-      &RunPack<ThePath, LhsKernelLayout, TBitpacked, PackedTBitpacked>;
+      &compute_engine::tflite::RunPack<ThePath, LhsKernelLayout>;
```
Tombana, line 49: From what I understand from the comments in `ruy_pack.h` , we only implement optimized 4x4 layouts (required for the RHS) and not 8x4 layouts for the LHS, since the LHS is cached anyway; makes sense :+1:. However here we call `compute_engine::tflite::RunPack` on the LHS, and I don't understand why that would not throw a compiler error because that `RunPack` only seems to have support for 4x4, and I don't see any code that makes it fallback to the generic Ruy packing code.

larq_compute_engine/core/bgemm_trmul_params.h
AdamHillier: It seems that in all cases, the 'packed type' is always just equal to the original type: https://github.com/google/ruy/blob/f876353d64a03e0e1b5d2262c0accbc5011bda3b/ruy/pack_common.h#L47-L53

I *think* that the distinction between a packed type and a normal type exists so that it's possible to take uint8 data and convert it to int8 as part of the packing process, for example. It's weird though because as far as I can tell that functionality doesn't actually get used anywhere in Ruy.

In any case, we definitely want to keep the `TBitpacked` type when we perform the packing.

larq_compute_engine/tflite/kernels/bconv2d.cc
AdamHillier: Yep you're definitely right, thanks for pointing that out. To be safe, let's cache the set value, set it to true, run our op eval, and reset it to the original value at the end of inference.

larq_compute_engine/core/bgemm_trmul_params.h
Tombana: Thanks :+1:

larq_compute_engine/core/bgemm_impl_ruy.h
```diff
@@ -38,17 +39,13 @@ struct BGemmImplUsingRuy {
...
-    ruy::MakeSimpleLayout(rhs_params.rows, rhs_params.cols,
-                          ruy::Order::kColMajor, rhs.mutable_layout());
     ruy::Matrix<DstScalar> dst;
-    ruy::MakeSimpleLayout(dst_params.rows, dst_params.cols,
-                          ruy::Order::kColMajor, dst.mutable_layout());
-    lhs.set_data(lhs_data);
-    rhs.set_data(rhs_data);
-    dst.set_data(dst_data);
+    cpu_backend_gemm::detail::MakeRuyMatrix(lhs_params, lhs_data, &lhs,
+                                            context->use_caching());
```
lgeiger, line 45: What about hard coding `use_caching = true` instead of infering the value from the context which is currently overwritten in the TFLite kernel?

AdamHillier: Yeah that's nicer, thanks :+1:

Tombana:
Did you verify that having the global cache option set to false, but the parameter to `MakeRuyMatrix` set to true, actually caches the matrix? (e.g. by benchmarking Quicknet for example)

AdamHillier:
> Did you verify that having the global cache option set to false, but the parameter to `MakeRuyMatrix` set to true, actually caches the matrix? (e.g. by benchmarking Quicknet for example)

Yep, it works nicely :+1:

AdamHillier:
Note that the benchmark numbers for the Raspberry Pi recorded here (both the baseline and the PR) are slightly questionable - we've had challenges reproducing them. See https://github.com/larq/compute-engine/pull/470#issuecomment-679322167.

PR #464, opened 2020-08-19
Change custom option enums to match flatbuffer schema

### LINKED ISSUE #458: LceBconv2d "padding" custom option enum
It looks like the `padding` custom option is encoded as a `TfLitePadding` enum value. While there is nothing that prevents using this in a custom option, it is slightly odd since the built-in TFL operators all use the `Padding` enum from the schema. The difference is that `TfLitePadding` can encode unknown padding, which is not permitted by the schema (only same and valid).

If there is no need for the custom options to encode the unknown case, would it be possible to change this? It seems desirable to make the serialized flatbuffer depend explicitly on the runtime as little as possible, and would also make our code simpler.

lgeiger:
## What do these changes do?
This changes activation and padding enums to match the enum used in the TFLite flatbuffer schema.

## How Has This Been Tested?
CI

## Related issue number
Closes #458

### DIFF
--- a/larq_compute_engine/tflite/kernels/utils.h
+++ b/larq_compute_engine/tflite/kernels/utils.h
@@ -1,0 +1,17 @@
+#ifndef COMPUTE_ENGINE_TFLITE_KERNEL_UTILS_H
+#define COMPUTE_ENGINE_TFLITE_KERNEL_UTILS_H
+
+#include "tensorflow/lite/c/builtin_op_data.h"
+#include "tensorflow/lite/schema/schema_generated.h"
+
+namespace tflite {
+
+// Converts the flatbuffer activation to what is used at runtime.
+TfLiteFusedActivation ConvertActivation(ActivationFunctionType activation);
+
+// Converts the flatbuffer padding enum to what is used at runtime.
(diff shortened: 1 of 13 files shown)


larq_compute_engine/tflite/kernels/utils.h
```diff
@@ -0,0 +1,17 @@
...
+#include "tensorflow/lite/c/builtin_op_data.h"
+#include "tensorflow/lite/schema/schema_generated.h"
+
+namespace tflite {
+
+// Converts the flatbuffer activation to what is used at runtime.
+TfLiteFusedActivation ConvertActivation(ActivationFunctionType activation);
+
+// Converts the flatbuffer padding enum to what is used at runtime.
+TfLitePadding ConvertPadding(Padding padding);
```
lgeiger, line 13: Not sure if we could reuse [these functions from TensorFlow](https://github.com/tensorflow/tensorflow/blob/22a73ff29265cb534d7f56a4fe7631c8f09e7813/tensorflow/lite/core/api/flatbuffer_conversions.cc#L104-L132) somehow.

Tombana (approved):

PR #466, opened 2020-08-19
Add Quantize and Dequantize ops

Tombana:
## What do these changes do?
This adds Quantize and Dequantize ops, to be merged into #457 , but made as a separate PR for easier reviewing.

## How Has This Been Tested?
Tests for these ops have been added, although only in a simple form: we run quantize + dequantize on random `{-1, +1}` input and check if this is identity. Since the dequantize op never sees the original input, I think there is really no way this can pass the test in any other way than actually doing the correct thing.


larq_compute_engine/core/bitpack.h
```diff
@@ -196,9 +196,10 @@ inline void bitpack_matrix(const TIn* input, const std::size_t input_num_rows,
 
 template <typename TUnpacked>
 void unpack_bitfield(const TBitpacked in, TUnpacked*& out,
-                     std::size_t num_elements) {
+                     std::size_t num_elements, const TUnpacked zero_bit_result,
+                     const TUnpacked one_bit_result) {
   for (size_t i = 0; i < num_elements; ++i) {
-    *out++ = (in & (1ULL << i)) ? TUnpacked(-1) : TUnpacked(1);
+    *out++ = (in & (1ULL << i)) ? one_bit_result : zero_bit_result;
```
AdamHillier, line 202:
```suggestion
    *out++ = (in & (TBitpacked(1) << i)) ? one_bit_result : zero_bit_result;
```

larq_compute_engine/tflite/kernels/quantization.cc
```diff
@@ -0,0 +1,138 @@
...
+    TF_LITE_ENSURE_EQ(context, output->dims->data[i], input->dims->data[i]);
+  }
+  // The last dimension is bitpacked
+  int packed_channels = input->dims->data[num_dims - 1];
+  int unpacked_channels = output->dims->data[num_dims - 1];
+  TF_LITE_ENSURE_EQ(context, packed_channels,
+                    ce::core::GetPackedSize(unpacked_channels));
+
+  // We don't support resizing here, because we can not know the number of
+  // output channels based on the number of input channels
```
AdamHillier, line 69: What impact does this have?

larq_compute_engine/tflite/tests/quantization_test.cc
```diff
@@ -0,0 +1,132 @@
...
+typedef TfLiteRegistration* (*register_function)(void);
+
+class QuantizationOpModel : public SingleOpModel {
+ public:
+  QuantizationOpModel(register_function registration, const TensorData& input,
+                      const TensorData& output) {
+    input_ = AddInput(input);
+    output_ = AddOutput(output);
+    // It doesn't matter if the name is not correct
+    SetCustomOp("LceQuantization", {}, registration);
```
AdamHillier, line 28: I don't quite understand the comment here, why doesn't it matter?

larq_compute_engine/tflite/kernels/quantization.cc
```diff
@@ -0,0 +1,138 @@
...
+  const TfLiteTensor* input = GetInput(context, node, 0);
+  TfLiteTensor* output = GetOutput(context, node, 0);
+
+  TF_LITE_ENSURE(context,
+                 input->type == kTfLiteFloat32 || input->type == kTfLiteInt8);
+  TF_LITE_ENSURE_EQ(context, output->type, kTfLiteInt32);
+
+  TF_LITE_ENSURE_EQ(context, NumDimensions(input), NumDimensions(output));
+
+  int num_dims = NumDimensions(input);
```
lgeiger, line 29:
```suggestion
  int num_dims = NumDimensions(input);

  TF_LITE_ENSURE_EQ(context, num_dims, NumDimensions(output));
```

larq_compute_engine/tflite/tests/quantization_test.cc
Tombana: Normally, when a model is loaded, the `OpResolver` maps the op names to the corresponding registration functions. In this testing setting, however, we directly build an interpreter object with the model 'already loaded' and op names already resolved. The only effect of the string here, is that when the op throws an error it'll print a message like `"Node 0 (LceQuantization) failed to prepare."`. And since I'm using this class for both the Quantize and Dequantize op, I thought I'd just give it a single name that is a bit more general.

larq_compute_engine/tflite/kernels/quantization.cc
```diff
@@ -0,0 +1,138 @@
...
+  const TfLiteTensor* input = GetInput(context, node, 0);
+  TfLiteTensor* output = GetOutput(context, node, 0);
+
+  TF_LITE_ENSURE_EQ(context, input->type, kTfLiteInt32);
+  TF_LITE_ENSURE(context,
+                 output->type == kTfLiteFloat32 || output->type == kTfLiteInt8);
+
+  TF_LITE_ENSURE_EQ(context, NumDimensions(input), NumDimensions(output));
+
+  int num_dims = NumDimensions(input);
```
lgeiger, line 56:
```suggestion
  int num_dims = NumDimensions(input);

  TF_LITE_ENSURE_EQ(context, num_dims, NumDimensions(output));
```

lgeiger (commented):
Very nice!

larq_compute_engine/tflite/kernels/quantization.cc
Tombana: I think we never use this resizing: basically you can resize the input tensor of your model, for example when you use a different camera resolution. The `Prepare` functions are then all called again (`Init` is not called again), and they are expected to resize their internal buffers (im2col buffer for example) and propagate a new output size, to be sent to the `Prepare` function of the next op.
This way you can use the same `.tflite` file on different input image sizes. Also, I think early toco converters would not store shape information of intermediate tensors in the `.tflite` file, so back then ops definitely had to compute their output size, although I'm not entirely sure about this.
Note that tf-micro does *not* support this resizing, and expects shapes to be present in the `.tflite` file.

For the Dequantize op, it can't compute the number of output channels based on the input shape, since it only knows the packed size. We could add it as a custom option (similar to `input_channels` in `LceBconv2d`), but I thought it'd be easier to just read it from the output tensor, and assume that no resizing will happen. Especially since we never expect to actually use the Dequantize op in a real model.

lgeiger (approved):

larq_compute_engine/tflite/kernels/quantization.cc
AdamHillier: Makes sense, thanks :+1: I guess also even if we somehow were in a scenario that needed resizing, we'd expect the next op to perform validation of it's input tensor size against what it expects, so it wouldn't fail silently.

AdamHillier (approved):

PR #467, opened 2020-08-20
Simplify bconv test

Tombana:
## What do these changes do?
This updates the unittests to only use bitpacked input, to be merged into #457.

It further simplifies the tests by making the builtin op always use float in/output instead of sometimes float and sometimes int8.

### DIFF
--- a/larq_compute_engine/tflite/tests/bconv2d_test.cc
+++ b/larq_compute_engine/tflite/tests/bconv2d_test.cc
@@ -726,8 +726,6 @@ INSTANTIATE_TEST_SUITE_P(
 
 // The BigTest suite will be skipped in the qemu CI runs as they take more than
 // an hour.
-// This test does not include zero-padding because it will be skipped in most
-// configurations
 INSTANTIATE_TEST_SUITE_P(
     BigTest, BConv2DOpTest,
     ::testing::Combine(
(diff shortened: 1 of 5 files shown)


larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -731,41 +674,15 @@ void runTest(const TestParam& param) {
...
-                     builtin_output, builtin_output_tensor.zero_point);
+                     builtin_output, output_tensor.zero_point,
+                     output_tensor.scale);
 }
 
-// Three input types: int8, TBitpacked (int32), float
-// Three output types: int8, TBitpacked (int32), float
-// Thats 9 combinations, but float -> int8 and int8 -> float are never used so
-// that leaves 7 combinations.
+TEST_P(BConv2DOpTest, Bitpacked) { runTest<TBitpacked>(TestParam(GetParam())); }
```
AdamHillier, line 681: How do you feel about being more explicit with the test name here (also `FloatOutput` and `Int8Output` for the cases below)?

```suggestion
TEST_P(BConv2DOpTest, BitpackedOutput) { runTest<TBitpacked>(TestParam(GetParam())); }
```

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -824,16 +743,44 @@ INSTANTIATE_TEST_SUITE_P(
...
+               std::array<int, 3>{2, 3, 64}),  // filter shape [HWO]
+        Values(std::array<int, 2>{1, 1},
+               std::array<int, 2>{2, 3}),  // strides height/width
+        Values(std::array<int, 2>{1, 1},
+               std::array<int, 2>{3, 2}),     // dilation height/width
+        Values(Padding_SAME),                 // padding
+        Values(ActivationFunctionType_NONE),  // activation function
+        Values(1, 2),                         // number of threads
+        Values(std::pair<std::string, register_function>{
+            "BConv2DOPT", compute_engine::tflite::Register_BCONV_2D_OPT})),
```
AdamHillier, line 779: Does this mean that we wouldn't test 'same-zero' padding with the reference kernel?

AdamHillier (commented):
Nice, it's good to see things getting simpler :)

The only thing I'm unsure about is splitting off the `same-zero` padding tests into a separate test group. Is the fact that so many test cases (with 'same-zero' padding) are skipped with the existing 'big' test group a problem?

larq_compute_engine/tflite/tests/bconv2d_test.cc
Tombana: Good idea :+1:

larq_compute_engine/tflite/tests/bconv2d_test.cc
Tombana: The reference kernel doesn't support it (although it should be easy to add), so the that test was already skipped.

Tombana:
> The only thing I'm unsure about is splitting off the `same-zero` padding tests into a separate test group. Is the fact that so many test cases (with 'same-zero' padding) are skipped with the existing 'big' test group a problem?

After thinking some more about it, I think you're right, I'll revert that commit.

PR #468, opened 2020-08-21
Remove int8/float input support from bconv

### LINKED ISSUE #110: fusing bitpacking with im2col operation for bitpacking along channel dimension
falls #52 is finished, we can fuse the bitpacking with im2col. Technically we bitpack the channels while  we are copying them in im2col.

### LINKED ISSUE #290: Benchmark when "bitpack after im2col" is faster

lgeiger:
## What do these changes do?
This PR removes support for floating point and integer input in the bconv op.

## How Has This Been Tested?
CI

## Related issue number
To be merged into #457

Closed #290 and closes #110 in favour of #447

### DIFF
--- a/larq_compute_engine/tflite/kernels/bconv2d.cc
+++ b/larq_compute_engine/tflite/kernels/bconv2d.cc
@@ -45,17 +45,6 @@ enum class KernelType {
   kRuyOptimized,
 };
 
-inline void decide_bitpack_before_im2col(KernelType kernel_type,
-                                         TfLiteBConv2DParams* conv_params) {
-  if (kernel_type == KernelType::kGenericRef ||
-      conv_params->read_bitpacked_input ||
-      conv_params->channels_in >= bitpacking_bitwidth / 4) {
-    conv_params->bitpack_before_im2col = true;
-  } else {
-    conv_params->bitpack_before_im2col = false;
-  }
-}
-
 #define LCE_ENSURE_PARAM(conv_params, context, a)                       \
   do {                                                                  \
     if (!(a)) {                                                         \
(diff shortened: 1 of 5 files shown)


larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -367,54 +316,12 @@ TfLiteStatus Prepare(TfLiteContext* context, TfLiteNode* node) {
         GetTemporary(context, node, conv_params->im2col_index);
 
     // Determine the type
```
Tombana, line 318:
```suggestion
```

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -507,26 +414,6 @@ void SetupQuantization(TfLiteContext* context, TfLiteNode* node,
...
-
-template <>
-struct GetUnpackType<std::int8_t> {
-  using type = std::int8_t;
-};
-
-template <class SrcScalar>
 void OneTimeSetup(TfLiteContext* context, TfLiteNode* node,
                   TfLiteBConv2DParams* params) {
   if (!params->is_filter_repacked || !params->is_padding_correction_cached) {
```
Tombana, line 419: I think we can remove the entire 'repacking' section of the code now, because all packing is now canonical.
EDIT: I see that we use the unpacked type for the zero-padding correction, but I think the zero-padding correction also supports packed filters.

PR #470, opened 2020-08-21
Update benchmark table in Readme

### LINKED ISSUE #469: Update benchmark table in Readme
We've made a lot of great performance improvements lately, before releasing v0.4 we need to update the [official benchmarks](https://github.com/larq/compute-engine#performance) in the README.

lgeiger:
Closes #469

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,12 +50,12 @@ and a Raspberry Pi 4 Model B ([BCM2711](https://www.raspberrypi.org/documentatio
 
 | Model                                                                                                                                                                      | Top-1 Accuracy | RPi 4 B, ms (1 thread) | Pixel 1, ms (1 thread) |
 | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | :------------: | :--------------------: | :--------------------: |
-| [QuickNet](https://docs.larq.dev/zoo/api/sota/#quicknet) ([.h5](https://github.com/larq/zoo/releases/download/quicknet-v0.2.0/quicknet_weights.h5))                        |     58.6 %     |          34.9          |          18.3          |
-| [QuickNet-Large](https://docs.larq.dev/zoo/api/sota/#quicknetlarge) ([.h5](https://github.com/larq/zoo/releases/download/quicknet_large-v0.2.0/quicknet_large_weights.h5)) |     62.7 %     |          52.7          |          27.6          |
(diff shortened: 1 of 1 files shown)


AdamHillier:
## Adventures in Raspberry Pi benchmarking

I thought I was going mad. The numbers you quote here for the Raspberry Pi 4B are substantially worse than even the baseline that I measured in #462. Unfortunately, after doing a Manjaro OS update my Pi developed a boot issue, which I can't fix because I don't have a USB keyboard to hand. Thus, I couldn't actually re-run the benchmarks.

However, I had another MicroSD card lying around with an old Manjaro ARM image on it, so I used that instead, and was able to run benchmarks again. I got similar numbers to what you have here, e.g. 31.4 ms for QuickNet single-threaded. This is way worse than the 30.5 ms that I measured with the exact same hardware only a week or so ago in #462.

After quite a long time trying to work out what was going on, I physically handled the device, and lots of parts of it were very hot, including not just the CPU but also the USB/networking stack. And then I remembered that it's possible to disable some parts of the Pi to save power through a boot config file: [`/boot/config.txt`](https://www.raspberrypi.org/documentation/configuration/config-txt/boot.md). In fact I had previously done just this - I had disabled WiFi and Bluetooth on the device (I used a LAN cable for networking). But swapping to the other SD card meant this change was no longer applied.

After making the same change with the new MicroSD card, I restarted it, and recorded some highly variable benchmark runs on single threaded QuickNet, including one with an average latency of 30.7 ms. This is not quite as good as the 30.5 ms from #462, but is pretty close. However, the results were really sporadic. Most runs were around 31.4/5/6 ms, and that 30.7 was the only run (out of maybe 10) with an average below 31 ms. As a result, I'm not sure at all that disabling the WiFi/Bluetooth had an effect.

I have no idea what the conclusion should be drawn here. As far as I can remember, when recording the benchmarks for #462 the results were very consistent, and I haven't been able to reproduce them now. The hardware is clearly capable of faster runs than 31.4 ms, but I'm not sure what to do to make that happen. None of this makes much sense to me, but my best guess is that there's some nuance and/or weirdness relating to the Pi's power management that's at play.

I think the safest option is probably to go with the numbers you've recorded here - at the end of the day there's a < 3% difference between them. They also are definitely achievable with a 'stock' Pi without fiddling about with configuration.

AdamHillier (approved):

PR #471, opened 2020-08-23
Add profiler scopes for bitpacking functions and quantization/maxpool ops.

AdamHillier:
## What do these changes do?

This is a non-functional PR that adds [Ruy profiler](https://github.com/google/ruy/blob/4452e73747d162f86aadc614f3296420d0d345a9/ruy/profiler/README.md) scopes to the bitpacking (and unpacking) functions as well as the `LceQuantize` and `LceDequantize` ops.

## How Has This Been Tested?

CI.

## Benchmark Results

N/A.

## Related issue number

N/A.

### DIFF
--- a/larq_compute_engine/core/bgemm_kernels_ruy.h
+++ b/larq_compute_engine/core/bgemm_kernels_ruy.h
@@ -111,7 +111,7 @@ struct BgemmKernel<ruy::Path::kStandardCpp, TBitpacked, Spec> {
 
     const int depth = lhs.layout.rows;
     const int dst_stride_bitpacked =
-        ce::core::GetPackedSize(dst->layout.stride);
+        ce::core::GetBitpackedSize(dst->layout.stride);
 
     // The destination is column major and we need to bitpack along the row
     // (channels) axis so we need to loop over column index then row index.
--- a/larq_compute_engine/core/bitpack.h
+++ b/larq_compute_engine/core/bitpack.h
@@ -11,25 +11,28 @@
 #include "larq_compute_engine/core/bitpack_aarch64.h"
 #endif
 
-// From @flatbuffers, used for the FLATBUFFERS_LITTLEENDIAN macro
-#include "flatbuffers/base.h"
+#include "flatbuffers/base.h"  // Used for the FLATBUFFERS_LITTLEENDIAN macro
+#include "ruy/profiler/instrumentation.h"
+#include "tensorflow/lite/kernels/internal/types.h"
 
 namespace compute_engine {
 namespace core {
 
 // Utility functions
 
-constexpr int GetPackedSize(int unpacked_elements) {
+constexpr int GetBitpackedSize(int unpacked_elements) {
   return (unpacked_elements + bitpacking_bitwidth - 1) / bitpacking_bitwidth;
 }
 
-constexpr int GetPackedMatrixSize(int rows, int cols) {
-  return rows * GetPackedSize(cols);
+constexpr int GetBitpackedMatrixSize(int rows, int cols) {
+  return rows * GetBitpackedSize(cols);
 }
 
 template <class TIn>
-inline void pack_quantized(const TIn* in, TBitpacked* out,
-                           const std::int32_t zero_point) {
+inline void bitpack_bitfield_quantized(const TIn* in, TBitpacked* out,
+                                       const std::int32_t zero_point) {
+  ruy::profiler::ScopeLabel label(
+      "Bitpack bitfield (quantized input, unoptimised)");
   *out = 0;
   for (size_t i = 0; i < bitpacking_bitwidth; ++i) {
     // Note: uint8 to int32 will set the top 24 bits to 0
@@ -40,7 +43,10 @@ inline void pack_quantized(const TIn* in, TBitpacked* out,
 }
 
 template <class T>
-inline void pack(const T* fptr, TBitpacked* buf) {
+inline void bitpack_bitfield(const T* fptr, TBitpacked* buf) {
+  ruy::profiler::ScopeLabel label(
+      "Bitpack bitfield (non-quantized input, unoptimised)");
+
   struct bf {
     unsigned int b0 : 1;
     unsigned int b1 : 1;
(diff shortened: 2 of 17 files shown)


larq_compute_engine/core/bgemm_kernels_ruy.h
```diff
@@ -111,7 +111,7 @@ struct BgemmKernel<ruy::Path::kStandardCpp, TBitpacked, Spec> {
 
     const int depth = lhs.layout.rows;
     const int dst_stride_bitpacked =
-        ce::core::GetPackedSize(dst->layout.stride);
+        ce::core::GetBitpackedSize(dst->layout.stride);
```
AdamHillier, line 114: There are few things like this that were missed in #463 that I'm updating here.

larq_compute_engine/tflite/kernels/quantization.cc
```diff
@@ -72,6 +73,8 @@ TfLiteStatus DequantizePrepare(TfLiteContext* context, TfLiteNode* node) {
 }
 
 TfLiteStatus QuantizeEval(TfLiteContext* context, TfLiteNode* node) {
+  ruy::profiler::ScopeLabel label("Binary quantize");
```
AdamHillier, line 76: I originally named the scope after the op name `LceQuantize`, but that's inconsistent with the scope for our bconv op, which has op name `LceBconv2d`, but scope name `BConv2D`. I'm open to alternative suggestions for names.

AdamHillier:
Here is an example profile from a single-threaded run of QuickNet on my Android phone:

```
Profile (1 threads):

Thread 0 (4168 samples)

* 64.11% BConv2D
  * 57.97% BGemm
    * 57.97% BGemm/GeneralBGEMM
      * 57.97% BGemmRuy
        * 57.51% TrMul (Path=0x4, max_num_threads=1, is_prepacked=(1,0))
          * 38.17% TrMulImpl, general case
            * 37.38% Binary Kernel (8x4) (kNeon, optimized for out-of-order cores)
            * 0.70% Pack (ColMajor, 4x4)
            * 0.10% [other]
          * 19.31% TrMulImpl, simple loop
            * 19.17% Binary Kernel (8x4) (kNeon, optimized for out-of-order cores)
            * 0.12% Pack (ColMajor, 4x4)
            * 0.02% [other]
          * 0.02% [other]
        * 0.38% [other]
        * 0.07% Pack (generic)
  * 6.02% Im2col
    * 5.42% [other]
    * 0.60% ExtractPatchIntoBufferColumn
  * 0.12% [other]
* 16.00% Conv
  * 9.48% Im2col
    * 5.81% [other]
    * 3.67% ExtractPatchIntoBufferColumn
  * 6.53% cpu_backend_gemm::Gemm
    * 6.48% cpu_backend_gemm::Gemm: general GEMM
      * 6.43% Mul
        * 6.43% matmul shape: 8x27x12544
          * 6.41% TrMul (Path=0x4, max_num_threads=1, is_prepacked=(0,0))
            * 6.41% TrMulImpl, general case
              * 3.12% Pack (kNeon, optimized for out-of-order cores)
              * 2.81% Kernel (kNeon, optimized for out-of-order cores)
              * 0.48% [other]
          * 0.02% [other]
      * 0.05% [other]
    * 0.05% [other]
* 7.44% DepthwiseConv
  * 7.44% DepthwiseConv/float/DepthwiseConvImpl
    * 6.14% void tflite::optimized_ops::FloatDepthwiseConvAccumRow(int, int, int, int, const float *, int, int, int, const float *, int, int, int, float *) [kAllowStrided = true, kFixedInputDepth = 0, kFixedDepthMultiplier = 8]
    * 1.30% [other]
* 3.93% FullyConnected
  * 3.93% cpu_backend_gemm::Gemm
    * 3.93% cpu_backend_gemm::Gemm: CustomGemv
* 3.55% Add
* 2.54% MaxPool
* 2.26% Binary quantize
  * 2.18% Bitpack 4x32 block (non-quantized input, optimised)
  * 0.07% [other]
* 0.14% Softmax
  * 0.14% Softmax/Impl
* 0.02% AveragePool
```

lgeiger (approved):
:+1: Thanks

larq_compute_engine/tflite/kernels/bmaxpool.cc
```diff
@@ -72,6 +73,8 @@ TfLiteStatus Prepare(TfLiteContext* context, TfLiteNode* node) {
 }
 
 TfLiteStatus Eval(TfLiteContext* context, TfLiteNode* node) {
+  ruy::profiler::ScopeLabel label("Binary MaxPool");
```
AdamHillier, line 76: Not sure if this should be like so or `"BMaxPool2D"`, like the existing `"BConv2D"` label for the binary conv op.

larq_compute_engine/core/bitpack.h
```diff
@@ -11,25 +11,28 @@
 #include "larq_compute_engine/core/bitpack_aarch64.h"
 #endif
 
-// From @flatbuffers, used for the FLATBUFFERS_LITTLEENDIAN macro
-#include "flatbuffers/base.h"
+#include "flatbuffers/base.h"  // Used for the FLATBUFFERS_LITTLEENDIAN macro
+#include "ruy/check_macros.h"
+#include "ruy/profiler/instrumentation.h"
```
Tombana, line 16: I'd rather not put this in our core files but in the tflite kernel files, because of lce-micro where we use the same core files but don't have the ruy profiler. (The 'optimized' label in the aarch64 header would still be fine).
However I think that i saw some ruy profiler files in recent tf-micro builds so maybe this is not a problem anymore, I'll investigate.

Tombana: Apparently the file `ruy/profiler/instrumentation.h` is now pulled into the tf-micro builds, maybe because they realized these tags were scattered all over the code so it was easier to just add that profiler file.
That is the *only* file, however, so the `check_macros.h` file can definitely not be used.
We could just move the tags into the tflite kernel files, except that we can then not see if the optimized or non-optimized path is used. I'm also fine with leaving the tags here (given that we can get rid of `check_macros.h`). What do you think @adamh ?

AdamHillier: Awesome, I've removed the `check_macros.h` import and replaced it with `tflite/kernels/internal/types.h` which I think should be fine for micro?

lgeiger (approved):

Tombana (approved):

PR #477, opened 2020-08-26
Remove lce_benchmark_all binary

lgeiger:
This PR removes the `lce_benchmark_all` build target. This build target is not documented anywhere and I don't think anyone is still using it. I think this can savely be removed.


AdamHillier (approved):

examples/lce_minimal.cc
```diff
@@ -7,7 +7,7 @@
 #include "tensorflow/lite/optional_debug_tools.h"
 
 // This file is based on the TF lite minimal example where the
-// "BuiltinObResolver" is modified to include the "Larq Compute Engine" custom
+// "BuiltinOpResolver" is modified to include the "Larq Compute Engine" custom
 // ops. Here we read a binary model form disk and perform inference by using the
```
Tombana, line 11:
```suggestion
// ops. Here we read a binary model from disk and perform inference by using the
```

Tombana (approved):

PR #478, opened 2020-08-26
Remove zero_point argument from im2col

lgeiger:
## What do these changes do?
Since we are always doing im2col on bitpacked output the zero point argument to `im2col` is not needed anymore.

## Related issue number
More cleanup continuing #457

### DIFF
--- a/larq_compute_engine/tflite/kernels/bconv2d_impl.h
+++ b/larq_compute_engine/tflite/kernels/bconv2d_impl.h
@@ -25,14 +25,13 @@ inline void im2col(const ConvParams& params, const RuntimeShape& input_shape,
                    const RuntimeShape& filter_shape,
                    const RuntimeShape& output_shape,
                    const RuntimeShape& im2col_shape, TBitpacked* im2col_data,
-                   RuntimeShape& result_shape, const TBitpacked** result_data,
-                   const std::int32_t zero_point) {
+                   RuntimeShape& result_shape, const TBitpacked** result_data) {
   const int stride_width = params.stride_width;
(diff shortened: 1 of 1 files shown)


AdamHillier (approved):

larq_compute_engine/tflite/kernels/bconv2d_impl.h
```diff
@@ -25,14 +25,13 @@ inline void im2col(const ConvParams& params, const RuntimeShape& input_shape,
...
-                   RuntimeShape& result_shape, const TBitpacked** result_data,
-                   const std::int32_t zero_point) {
+                   RuntimeShape& result_shape, const TBitpacked** result_data) {
   const int stride_width = params.stride_width;
   const int stride_height = params.stride_height;
   const int dilation_width_factor = params.dilation_width_factor;
   const int dilation_height_factor = params.dilation_height_factor;
 
-  const std::uint8_t zero_byte = (std::uint8_t)zero_point;
+  const std::uint8_t zero_byte = 0;
```
Tombana, line 34: Do you think that for clarity/readability, it makes sense add a comment here that this `0` means it'll be `+1` padding? (Conversely, writing `~0` or `0xff` here would result in `-1` padding.) I do like that the `zero_point` argument is removed from the function arguments itself.
```suggestion
  // On bitpacked data, padding with `0` effectively pads with `+1.0` values.
  const std::uint8_t zero_byte = 0;
```

lgeiger: Good point :+1:

PR #484, opened 2020-08-30
Simplify BConv2DParams by removing flags.

AdamHillier:
## What do these changes do?

This is a minor change that removes two flags from the `BConv2DParams` struct.

## How Has This Been Tested?

CI.

## Benchmark Results

N/A.

## Related issue number

N/A.

### DIFF
--- a/larq_compute_engine/tflite/kernels/bconv2d.cc
+++ b/larq_compute_engine/tflite/kernels/bconv2d.cc
@@ -30,13 +30,7 @@ enum class KernelType {
   // kGenericRef: the reference implementation without im2col
   kGenericRef,
 
-  // kGenericOptimized: the impl. path using reference BGEMM kernels in RUY.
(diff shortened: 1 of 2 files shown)


larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -30,13 +30,7 @@ enum class KernelType {
   // kGenericRef: the reference implementation without im2col
   kGenericRef,
 
-  // kGenericOptimized: the impl. path using reference BGEMM kernels in RUY.
-  // TODO(arashb): the generic optimized needs to be redirected to the
-  // reference impl. of RUY kernels.
-  kGenericOptimized,
```
AdamHillier, line 36: This kernel type isn't implemented; we can add this back if we ever need to in the future.

Tombana (approved):

PR #487, opened 2020-09-01
Update int8 reference bitpack code

Tombana:
## What do these changes do?
I noticed that on Cortex-M the generic int8 bitpacking code compiles to a `ldr`+`str` for every OR'd bit instead of first ORing 32 bits and then storing the result.

## How Has This Been Tested?
Manually verified the generated assembly.

### DIFF
--- a/larq_compute_engine/core/bitpack.h
+++ b/larq_compute_engine/core/bitpack.h
@@ -33,10 +33,80 @@ inline void bitpack_bitfield_quantized(const TIn* in, TBitpacked* out,
                                        const TIn zero_point) {
   ruy::profiler::ScopeLabel label(
       "Bitpack bitfield (quantized input, unoptimised)");
-  *out = 0;
-  for (size_t i = 0; i < bitpacking_bitwidth; ++i) {
-    if (in[i] < zero_point) *out |= (TBitpacked(1) << i);
-  }
+  struct bf {
+    unsigned int b0 : 1;
+    unsigned int b1 : 1;
+    unsigned int b2 : 1;
+    unsigned int b3 : 1;
+    unsigned int b4 : 1;
+    unsigned int b5 : 1;
+    unsigned int b6 : 1;
+    unsigned int b7 : 1;
+    unsigned int b8 : 1;
+    unsigned int b9 : 1;
+    unsigned int b10 : 1;
+    unsigned int b11 : 1;
+    unsigned int b12 : 1;
+    unsigned int b13 : 1;
+    unsigned int b14 : 1;
+    unsigned int b15 : 1;
+    unsigned int b16 : 1;
+    unsigned int b17 : 1;
+    unsigned int b18 : 1;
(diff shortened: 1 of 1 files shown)


AdamHillier (approved):

lgeiger (approved):

Tombana:
I just realized that there was another easy tweak that also removed a few cycles per element on Cortex-M, so I also pushed that on this branch.

larq_compute_engine/core/bitpack.h
```diff
@@ -33,13 +33,15 @@ inline void bitpack_bitfield_quantized(const TIn* in, TBitpacked* out,
...
-  *out = 0;
+  TBitpacked result = 0;
+  TBitpacked bit = TBitpacked(1);
   for (size_t i = 0; i < bitpacking_bitwidth; ++i) {
     // Note: uint8 to int32 will set the top 24 bits to 0
     //        int8 to int32 will set the top 24 bits to the int8 sign bit
-    if (static_cast<std::int32_t>(in[i]) < zero_point)
-      *out |= (TBitpacked(1) << i);
+    if (static_cast<std::int32_t>(in[i]) < zero_point) result |= bit;
+    bit <<= 1;
```
AdamHillier, line 42: I think if the loop is unrolled you can do shift + or in a single cycle can't you (because the shift is constant)?

AdamHillier: (Which is to say, the existing line `result |= (TBitpacked(1) << i)` could be done in a single cycle if `i` is constant?

Tombana: Good point, by unrolling it into 32 pieces the shift will become a constant and can then be incorporated into the `orr`.
I'll benchmark a few implementations soon.

Tombana: Benchmarking showed that the "bitfield method" was fastest, so I went with that.

larq_compute_engine/core/bitpack.h
```diff
@@ -33,10 +33,80 @@ inline void bitpack_bitfield_quantized(const TIn* in, TBitpacked* out,
...
+    unsigned int b31 : 1;
+  };
+
+  union bf_i32 {
+    bf t;
+    TBitpacked i32;
+  };
+
+  bf_i32 i;
+  i.t.b0 = in[0] < zero_point;
```
AdamHillier, line 77: Out of interest, did you try benchmarking with `i.t.b0 = (int[0] - zero_point) >> 7`?

AdamHillier: Or I guess this is actually generic over TIn, so using numeric-limits to get num digits.

Tombana: Good idea, I did not try. Will do now!

Tombana: That turned out to be roughly 5% slower.

PR #490, opened 2020-09-02
Fuse part of the back-transformation into the multiplier and bias.

AdamHillier:
## What do these changes do?

This is a minor performance optimisation that fuses the subtraction of the back-transformation into the existing output transform multiplier and bias.

## How Has This Been Tested?

CI, and the 'big' kernel tests pass locally for Aarch64 and Arm32.

## Benchmark Results

I benchmarked QuickNet on my Android phone (a OnePlus 6), with `num_runs=250`, and report the average latency and the standard deviation below.

| Number of threads | Baseline      | PR            |
|-------------------|---------------|---------------|
| 1                 | 14.71 +- 0.07 | 14.62 +- 0.06 |
| 4                 | 6.40 +- 0.20  | 6.36 +- 0.26  |

As expected, there is a very minor performance improvement that's pretty much within run variation.

I didn't have converted QuickNet Large/XL models to hand so I haven't benchmarked those yet, let me know if you would like to see those additional benchmarks too.

## Related issue number

N/A.

### DIFF
--- a/larq_compute_engine/core/bconv2d_output_transform.h
+++ b/larq_compute_engine/core/bconv2d_output_transform.h
@@ -21,90 +21,125 @@ enum class OutputTransformDetails {
   PreprocessedIntegerOnly
 };
 
-// The `OutputTransform` struct describes what needs to be done to get from the
-// int32 accumulator value to the final result that is written back to memory.
-//
-// This transform depends on the destination type `DstScalar` which can be:
-// - float
-// - TBitpacked (i.e. int32_t, meaning bitpacked output)
-// - int8
-template <typename AccumScalar, typename DstScalar,
+/*
+ * The `OutputTransform` struct describes what needs to be done to get from the
+ * int32 accumulator value to the final result that is written back to memory.
+ */
+template <typename DstScalar,
           OutputTransformDetails = OutputTransformDetails::Default>
 struct OutputTransform {};
 
-// A part of the transformation is common to all output types.
-// This part is described by `OutputTransformBase`.
-template <typename AccumScalar>
-struct OutputTransformBase {
-  AccumScalar backtransform_add = 0;
-  std::int32_t clamp_min = std::numeric_limits<AccumScalar>::lowest();
-  std::int32_t clamp_max = std::numeric_limits<AccumScalar>::max();
+/*
+ * ------------
+ * Float output
+ * ------------
+ *
+ * Conceptually, the float output transform is as follows:
+ *   0. We start with an XOR-popcount accumulator value `accum`. It lies in the
+ *      range {0, ..., K}, where K := `kernel_height` * `kernel_width` *
+ *      `channels_in` is the maximum number of bits that can be set.
+ *   1. Perform a 'back-transformation', which is a linear transformation that
+ *      maps {0, ..., K} onto {-K, ..., K}:
+ *                         K - 2 * accum = -2 * accum + K
+ *      This yields the 'true output' of the binary convolution operation, in
+ *      the sense that it's the same value that you would get by running a
+ *      normal floating-point convolution with +1/-1 input and +1/-1 weights.
+ *   2. Perform a clamp operation that represents a fused activation function.
+ *   3. Convert to float and perform a linear transformation with a multiplier
+ *      and bias that represents a fused batch normalisation layer.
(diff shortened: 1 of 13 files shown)


larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -298,22 +295,40 @@ void OneTimeSetup(TfLiteContext* context, TfLiteNode* node,
...
+  // the output transform multiplier/bias.
+  if (output->type == kTfLiteFloat32 || output->type == kTfLiteInt8) {
+    params->output_transform_multiplier.resize(params->channels_out);
+    params->output_transform_bias.resize(params->channels_out);
+
+    const auto filter_shape = GetTensorShape(GetInput(context, node, 1));
+    const std::int32_t backtransform_add =
+        filter_shape.Dims(1) * filter_shape.Dims(2) * params->channels_in;
+    const double output_scale =
+        output->type == kTfLiteInt8 ? output->params.scale : 1.0f;
```
Tombana, line 308: Very nicely combined :+1:

larq_compute_engine/tflite/kernels/bconv2d_output_transform_utils.h
```diff
@@ -14,59 +14,27 @@ namespace tflite {
...
-  auto filter_shape = GetTensorShape(GetInput(context, node, 1));
-  output_transform.backtransform_add =
-      filter_shape.Dims(1) * filter_shape.Dims(2) * params->channels_in;
-  output_transform.clamp_min = params->output_activation_min;
-  output_transform.clamp_max = params->output_activation_max;
+// Fill in the OutputTransform values for float and/or int8 outputs
+template <typename DstScalar>
+void GetOutputTransform(OutputTransform<DstScalar>& output_transform,
+                        TfLiteContext* context, TfLiteNode* node,
+                        TfLiteBConv2DParams* params) {
```
Tombana, line 22: Very nice, this is so much cleaner than the whole `BaseParams` thing :smile:

Tombana (approved):
Great!

larq_compute_engine/core/bconv2d_output_transform.h
```diff
@@ -21,93 +21,128 @@ enum class OutputTransformDetails {
...
-    x *= post_activation_multiplier[out_channel];
-    x += post_activation_bias[out_channel];
-    return x;
+/*
+ * -----------
+ * Int8 output
+ * -----------
+ *
+ * For int8 output, there are two additional conceptual steps:
+ *   4. Multiply by the reciprical of the int8 output scale, and add the int8
```
lgeiger, line 88:
```suggestion
 *   4. Multiply by the reciprocal of the int8 output scale, and add the int8
```

lgeiger (approved):
Very nice!

larq_compute_engine/core/bconv2d_output_transform.h
```diff
@@ -21,93 +21,128 @@ enum class OutputTransformDetails {
...
+  std::int32_t clamp_max = std::numeric_limits<std::int32_t>::max();
+  const float* multiplier = nullptr;
+  const float* bias = nullptr;
+
+  std::int8_t Run(const std::int32_t accum, int out_channel) const {
+    TFLITE_DCHECK(multiplier != nullptr);
+    TFLITE_DCHECK(bias != nullptr);
+    // Clamping is done in int32
+    std::int32_t x = accum << 1;
+    x = std::max<std::int32_t>(std::min<std::int32_t>(x, clamp_max), clamp_min);
```
andrewstanfordjason, line 110: What do the clamps do? I assume the compiler will eliminate them as they statically resolve to true.  If they weren't there you could absorb the shift left into the multiplier array.

Tombana: The `clamp_min` , `clamp_max` values depend on the `"fused_activation_function"` attribute. They effectively take care of a ReLu. We could write two versions of the `OutputTransform`, with and without activation function. The latter could then indeed absorb the shift into the multiplier, and skip the clamp.

andrewstanfordjason: I see, that makes sense, thanks.

PR #491, opened 2020-09-03
Fix bug in 32-bit accumulator kernel introduced by 346f72e.

AdamHillier:
In #490 I accidentally missed out removing the subtract instructions from the 32-bit accumulator optimised Aarch64 kernel; this PR fixes that error.

For some reason the 'overflow' tests that are meant to test the 32-bit accumulator kernels were incorrectly not being run on Aarch64. When I did get them running (on the commit before #490) there were test failures because the floats were out-of-tolerance for the comparison (I would imagine this is because of the large input size for those specific tests). Thus, I have also increased the float comparison tolerance for large input sizes only.

## How Has This Been Tested?

CI.

## Benchmark Results

N/A.

## Related issue number

N/A.


Tombana (approved):

Tombana:
Nice catch!

lgeiger (approved):

PR #492, opened 2020-09-04
Add Int8 output support to the optimised binary convolution kernels.

AdamHillier:
## What do these changes do?

This PR adds support for writing Int8 output to the Aarch64 and Arm32 optimised assembly binary convolution kernels.

## How Has This Been Tested?

CI. The 'big' kernel tests pass locally for Aarch64 and Arm32.

## Benchmark Results

It turns out that benchmarking this change is quite difficult. Usually we use the benchmarking tool and measure the overall latencies of the QuickNet models. However, it appears (verified through profiling) that an Int8-quantised version of QuickNet is substantially slower than a float version because the full-precision ops are slower, in particular the depthwise convolution at the start and the add ops.

Therefore, to benchmark this change I have used the benchmark tool with the Ruy profiler enabled and multiplied the % of time spent on our `BConv2D` op by the average overall latency (`num_runs=250`). This is not a nice way of measuring things - having the profiler enabled *severely* increases run variation - but it should be enough to get a general idea.

These benchmarks were taken on my Android phone (a OnePlus 6), running the QuickNet models single-threaded. The reported values (in ms) are the estimated time spent executing the bconv op.

| Model         | Baseline (float) | Baseline (int8) | PR (int8) |
|---------------|------------------|-----------------|-----------|
| QuickNet      | 11.06            | 190.23          | 10.92     |
| QuickNetLarge | 15.60            | 260.65          | 16.11     |
| QuickNetXL    | 31.73            | 540.93          | 32.86     |

## Related issue number

N/A.

### DIFF
--- a/larq_compute_engine/core/bgemm_kernels_arm32.h
+++ b/larq_compute_engine/core/bgemm_kernels_arm32.h
@@ -122,7 +122,9 @@ void CheckOffsetsInKernelParams(const Params&) {
 
 // clang-format on
 
-void BinaryKernelNeonOutOfOrder4x4(BinaryKernelParams<4, 4>& params) {
+template <typename DstScalar>
+void BinaryKernelNeonOutOfOrder4x4(
+    BinaryKernelParams<DstScalar, 4, 4>& params) {
   CheckOffsetsInKernelParams(params);
   ruy::profiler::ScopeLabel label(
       "Binary Kernel (4x4) (kNeon, optimized for out-of-order cores)");
@@ -136,13 +138,20 @@ void BinaryKernelNeonOutOfOrder4x4(BinaryKernelParams<4, 4>& params) {
   TBitpacked* lhs_ptr = lhs_col_ptr;
   TBitpacked* rhs_ptr = rhs_col_ptr;
 
-  float* dst_col_ptr = params.dst_base_ptr;
-  float* dst_ptr = dst_col_ptr;
+  DstScalar* dst_col_ptr = params.dst_base_ptr;
+  DstScalar* dst_ptr = dst_col_ptr;
   int row = params.start_row;
   int col = params.start_col;
 
   asm volatile(
 #define RUY_MAKE_ZERO(reg) "vmov.i32 " #reg ", #0\n"
+
+#define IF_FLOAT_OUTPUT(a) ".if %c[float_output]\n" a ".endif\n"
+
+#define IF_INT8_OUTPUT(a) ".if %c[int8_output]\n" a ".endif\n"
+
+#define IF_FLOAT_ELIF_INT8_OUTPUT(a, b) \
+  ".if %c[float_output]\n" a ".elseif %c[int8_output]\n" b ".endif\n"
 
       "sub sp, sp, #" RUY_STR(RUY_STACK_OFFSET_SIZE) "\n"
       // auto dst_col_ptr = params.dst_base_ptr;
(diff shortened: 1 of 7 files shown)


larq_compute_engine/core/bgemm_kernels_arm32.h
```diff
@@ -136,14 +138,21 @@ void BinaryKernelNeonOutOfOrder4x4(BinaryKernelParams<4, 4>& params) {
...
 
   asm volatile(
 #define RUY_MAKE_ZERO(reg) "vmov.i32 " #reg ", #0\n"
 
+#define IF_FLOAT_OUTPUT(a) ".if %c[float_output]\n" a ".endif\n"
+
+#define IF_INT8_OUTPUT(a) ".if %c[int8_output]\n" a ".endif\n"
+
+#define IF_FLOAT_ELIF_INT8_OUTPUT(a, b) \
+  ".if %c[float_output]\n" a ".elseif %c[int8_output]\n" b ".endif\n"
```
Tombana, line 154: Nice construction!

larq_compute_engine/core/bgemm_kernels_arm32.h
```diff
@@ -341,6 +350,34 @@ void BinaryKernelNeonOutOfOrder4x4(BinaryKernelParams<4, 4>& params) {
...
 
+IF_INT8_OUTPUT(
+      // Convert to signed integer, rounding to nearest.
+      //     Arm32 NEON has an instruction (vcvtr) to convert from float->int
+      // with rounding-to-nearest behaviour, but for some reason when used here
+      // truncation (rounding-to-zero) behaviour was observed instead. It's
+      // unclear what the issue is - perhaps QEMU doesn't correctly emulate the
+      // rounding behaviour?
+      //     The work-around is to first convert from float to fixed-point (24
+      // integer bits, 8 fractional bits)...
```
Tombana, line 361: Another workaround would be to add `+0.5` and then either floor or truncate.

AdamHillier: Yeah I thought about adding 0.5 but I think to get that to work properly it has to be +0.5 for a positive value and -0.5 for a negative value, which means it isn't possible to blindly add 0.5 to the output transform bias.

And you're right that there's a FPSCR register with a bit that you can set for the rounding mode, but according to the docs I found Neon ignores this value (at least on Arm32).

larq_compute_engine/core/bgemm_kernels_arm64.h
```diff
@@ -353,6 +362,26 @@ void BinaryKernelNeonOutOfOrder4x4(const BinaryKernelParams<4, 4>& params) {
...
+      "fcvtns v26.4s, v26.4s\n"
+      "fcvtns v27.4s, v27.4s\n"
+
+      // Narrow to int16 with "signed saturating extract narrow" instructions.
+      "sqxtn v24.4h, v24.4s\n"
+      "sqxtn v25.4h, v25.4s\n"
+      "sqxtn v26.4h, v26.4s\n"
+      "sqxtn v27.4h, v27.4s\n"
+
+      // Narrow from to int8.
```
Tombana, line 378:
```suggestion
      // Narrow from int16 to int8.
```

larq_compute_engine/core/bgemm_kernels_arm64.h
```diff
@@ -358,6 +362,26 @@ void BinaryKernelNeonOutOfOrder4x4(const BinaryKernelParams<4, 4>& params) {
...
+      "fcvtns v24.4s, v24.4s\n"
+      "fcvtns v25.4s, v25.4s\n"
+      "fcvtns v26.4s, v26.4s\n"
+      "fcvtns v27.4s, v27.4s\n"
+
+      // Narrow to int16 with "signed saturating extract narrow" instructions.
+      "sqxtn v24.4h, v24.4s\n"
+      "sqxtn v25.4h, v25.4s\n"
+      "sqxtn v26.4h, v26.4s\n"
+      "sqxtn v27.4h, v27.4s\n"
```
Tombana, line 376: Is it correct that these (together with the other 4 below) are only meant for clamping to int8 range? Or are they moving the values into the lower lanes?

AdamHillier: These are essentially equivalent to a saturating shift-right-and-narrow with a shift of 0 - in each register we start off with 4 int32s next to each other, and after this block we have 4 int16s next to each other in the lower half of the register. Then after the block below we end up with four bytes next to each other in the lower quarter of each register.

It's true that we could replace the block above with:
```cpp
      "sqxtn v24.4h, v24.4s\n"
      "sqxtn2 v24.8h, v25.4s\n"
      "sqxtn v25.4h, v26.4s\n"
      "sqxtn2 v25.8h, v27.4s\n"
```
and then the block below could become
```cpp
      "sqxtn v24.8b, v24.8h\n"
      "sqxtn v25.8b, v25.8h\n"
```
getting rid of two instructions. Then the values of interest are the first 8 bytes (lower half) of each of the two registers. But I don't think it's worth it because by using fewer registers we would have tighter dependency between nearby instructions which could cause some pipeline stalls (ideally we want > 3 cycles between writing to and reading from the same register).

AdamHillier: Ah and just to add, the clamping to int8 range is something that comes for free from using 'saturating' instructions - but we would still have to have these instructions even if not so as to gather the bytes correctly from the int32 values, unless we wanted to write one byte at a time to the destination.

larq_compute_engine/core/bgemm_kernels_arm32.h
Tombana: Ah you're right, I missed that you wrote "rounding-to-zero" instead of "rounding to negative infinity" (for which +0.5 would work). The current workaround seems totally fine to me :+1: , thanks for figuring that out.

AdamHillier: Yeah this actually doesn't use any additional instructions compared to having `vcvtr` convert-with-rounding anyway, because the step where we shift-and-narrow-with-rounding-and-saturation the fixed-point values does the narrowing to int16 and so replaces what would have to otherwise be a block of `sqxtn` instructions (as used in the Aarch64 kernels).

AdamHillier:
I've updated the PR description with benchmark results (such that they are). I think this is now ready from my end.

Tombana (approved):

lgeiger (approved):

PR #495, opened 2020-09-08
Remove Preprocessed from OutputTransformDetails

Tombana:
## What do these changes do?
Previously, we had
```cpp
enum class OutputTransformDetails {
  Default,
  Preprocessed,
  PreprocessedIntegerOnly
};
```
but now the `Default` implementation does the same thing that `Preprocessed` does, so this PR removes the `Preprocessed` option and renames `PreprocessedIntegerOnly` to `IntegerOnly`.
The only thing that was still different in the `Preprocessed` version, was the use of a Cortex-M `round()` function, so I've added that here as well.

## How Has This Been Tested?
CI

### DIFF
--- a/larq_compute_engine/core/bconv2d_output_transform.h
+++ b/larq_compute_engine/core/bconv2d_output_transform.h
@@ -19,7 +19,7 @@ using compute_engine::core::TBitpacked;
 std::int8_t saturate(std::int32_t x) {
 #ifdef __arm__
   std::int8_t y;
-  asm("ssat %[y], 8, %[x]\n" : [ y ] "=r"(y) : [ x ] "r"(x));
+  asm("ssat %[y], #8, %[x]\n" : [ y ] "=r"(y) : [ x ] "r"(x));
   return y;
 #else
   x = std::min<std::int32_t>(x, std::numeric_limits<std::int8_t>::max());
--- a/larq_compute_engine/core/bconv2d_output_transform.h
+++ b/larq_compute_engine/core/bconv2d_output_transform.h
@@ -15,10 +15,39 @@ namespace core {
 
 using compute_engine::core::TBitpacked;
 
+// Clamp an int32 value to int8 range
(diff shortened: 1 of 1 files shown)


larq_compute_engine/core/bconv2d_output_transform.h
```diff
@@ -15,10 +15,39 @@ namespace core {
 
 using compute_engine::core::TBitpacked;
 
+// Clamp an int32 value to int8 range
+std::int8_t saturate(std::int32_t x) {
+#ifdef __thumb__
+  std::int8_t y;
+  asm("ssat %[y], 8, %[x]\n" : [ y ] "=r"(y) : [ x ] "r"(x));
+  return y;
+#else
```
AdamHillier, line 24: This is valid non-thumb Arm code too, right?

```suggestion
#ifdef __arm__
  std::int8_t y;
  asm("ssat %[y], 8, %[x]\n" : [ y ] "=r"(y) : [ x ] "r"(x));
  return y;
#else
```

larq_compute_engine/core/bconv2d_output_transform.h
```diff
@@ -112,10 +141,8 @@ struct OutputTransform<std::int8_t, OutputTransformDetails::Default> {
     float y =
         static_cast<float>(x) * multiplier[out_channel] + bias[out_channel];
     // And then we round back to int32 and clamp to the int8 range
-    std::int32_t z = tflite::TfLiteRound(y);
-    z = std::min<std::int32_t>(z, std::numeric_limits<std::int8_t>::max());
-    z = std::max<std::int32_t>(z, std::numeric_limits<std::int8_t>::lowest());
-    return static_cast<std::int8_t>(z);
+    std::int32_t z = round(y);
+    return saturate(z);
```
AdamHillier, line 145:
```suggestion
    return saturate(round(y));
```

AdamHillier (approved):

AdamHillier (approved):

PR #496, opened 2020-09-09
Add bitpacked output support to the Aarch64 optimised int16 kernel.

AdamHillier:
## What do these changes do?

This PR adds support for writing bitpacked output to the optimised assembly Aarch64 kernel with 16-bit accumulators.

Bitpacked output remains unsupported by the Aarch64 fallback optimised kernel with 32-bit accumulators, and the Arm32 optimised kernel. The reason for this is that the Aarch64 int16 kernel has a 8x4 kernel layout that writes eight output channels in parallel, whereas the others have a 4x4 kernel layout that writes only four channels in parallel. With eight channels we can always pack the bits into a byte, whereas four channels would require something more complicated (especially to support multi-threading).

## How Has This Been Tested?

CI. The 'big' kernel tests for Aarch64 and Arm32 pass locally.

## Benchmark Results

To benchmark this change, I've taken the binary alexnet model from Larq Zoo, modified it to use one-padding instead of zero-padding, and run the LCE benchmark tool (with `num_runs=250`) on my Android phone, a OnePlus 6, reporting the average latency and standard deviation:

| Model                             | Baseline       | PR            |
|-----------------------------------|----------------|---------------|
| Binary AlexNet (with one-padding) | 111.42 +- 0.60 | 64.95 +- 0.14 |

Note that the binary alexnet model includes large binary dense layers at the end of the model which are not currently accelerated by LCE. Thus, the performance improvement for the bconv op alone will be greater than the overall latency suggests. 

## Related issue number

N/A.

### DIFF
--- a/larq_compute_engine/core/bgemm_kernels_arm.h
+++ b/larq_compute_engine/core/bgemm_kernels_arm.h
@@ -21,20 +21,6 @@ using compute_engine::core::TBitpacked;
 
 #if RUY_PLATFORM_NEON
 
-// Generic kNeon template when no types are specified
-template <typename DstScalar, typename MulParamsType>
-struct BgemmKernel<ruy::Path::kNeon, DstScalar, MulParamsType> {
-  ruy::Tuning tuning = Tuning::kAuto;
-  using LhsLayout = FixedKernelLayout<Order::kRowMajor, 1, 8>;
-  using RhsLayout = FixedKernelLayout<Order::kRowMajor, 1, 8>;
-  explicit BgemmKernel(ruy::Tuning tuning_) : tuning(tuning_) {}
-  void Run(const ruy::PMat<TBitpacked>& lhs, const ruy::PMat<TBitpacked>& rhs,
-           const MulParamsType& mul_params, int start_row, int start_col,
-           int end_row, int end_col, ruy::Mat<DstScalar>* dst) const {
-    TFLITE_DCHECK(false);
-  }
-};
-
 // Generic kNeonDotprod template
 template <typename DstScalar, typename MulParamsType>
 struct BgemmKernel<ruy::Path::kNeonDotprod, DstScalar, MulParamsType> {
@@ -53,40 +39,23 @@ struct BgemmKernel<ruy::Path::kNeonDotprod, DstScalar, MulParamsType> {
 // A BGEMM kernel for ARM32 Neon.
 #include "bgemm_kernels_arm32.h"
 
-// Optimised Arm32 kernel for float output
-template <>
-struct BgemmKernel<ruy::Path::kNeon, float,
-                   BinaryMulParams<std::int32_t, float>> {
+// Optimised Arm32 kernel. Supports float or int8 output.
+template <typename DstScalar>
+struct BgemmKernel<ruy::Path::kNeon, DstScalar,
+                   BinaryMulParams<std::int32_t, DstScalar>> {
   Tuning tuning = Tuning::kAuto;
   using LhsLayout = FixedKernelLayout<Order::kColMajor, 4, 4>;
   using RhsLayout = FixedKernelLayout<Order::kColMajor, 4, 4>;
   explicit BgemmKernel(Tuning tuning_) : tuning(tuning_) {}
   void Run(const ruy::PMat<TBitpacked>& lhs, const ruy::PMat<TBitpacked>& rhs,
-           const BinaryMulParams<std::int32_t, float>& mul_params,
+           const BinaryMulParams<std::int32_t, DstScalar>& mul_params,
            int start_row, int start_col, int end_row, int end_col,
-           ruy::Mat<float>* dst) const {
-    BinaryKernelParams<float, LhsLayout::kCols, RhsLayout::kCols> params;
-    MakeBinaryKernelParams(lhs, rhs, mul_params, start_row, start_col, end_row,
-                           end_col, dst, &params);
-    BinaryKernelNeonOutOfOrder4x4(params);
-  }
-};
-
-// Optimised Arm32 kernel for int8 output
-template <>
-struct BgemmKernel<ruy::Path::kNeon, std::int8_t,
-                   BinaryMulParams<std::int32_t, std::int8_t>> {
-  Tuning tuning = Tuning::kAuto;
-  using LhsLayout = FixedKernelLayout<Order::kColMajor, 4, 4>;
-  using RhsLayout = FixedKernelLayout<Order::kColMajor, 4, 4>;
-  explicit BgemmKernel(Tuning tuning_) : tuning(tuning_) {}
-  void Run(const ruy::PMat<TBitpacked>& lhs, const ruy::PMat<TBitpacked>& rhs,
-           const BinaryMulParams<std::int32_t, std::int8_t>& mul_params,
-           int start_row, int start_col, int end_row, int end_col,
-           ruy::Mat<std::int8_t>* dst) const {
-    BinaryKernelParams<std::int8_t, LhsLayout::kCols, RhsLayout::kCols> params;
-    MakeBinaryKernelParams(lhs, rhs, mul_params, start_row, start_col, end_row,
-                           end_col, dst, &params);
+           ruy::Mat<DstScalar>* dst) const {
+    TFLITE_DCHECK((std::is_same<DstScalar, float>::value ||
+                   std::is_same<DstScalar, std::int8_t>::value));
+    BinaryKernelParams<DstScalar, LhsLayout::kCols, RhsLayout::kCols> params;
+    MakeBinaryKernelParams(lhs, rhs, start_row, start_col, end_row, end_col,
+                           dst, mul_params, &params);
     BinaryKernelNeonOutOfOrder4x4(params);
   }
 };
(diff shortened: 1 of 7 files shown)


larq_compute_engine/core/bgemm_kernels_arm.h
```diff
@@ -21,20 +21,6 @@ using compute_engine::core::TBitpacked;
...
-  ruy::Tuning tuning = Tuning::kAuto;
-  using LhsLayout = FixedKernelLayout<Order::kRowMajor, 1, 8>;
-  using RhsLayout = FixedKernelLayout<Order::kRowMajor, 1, 8>;
-  explicit BgemmKernel(ruy::Tuning tuning_) : tuning(tuning_) {}
-  void Run(const ruy::PMat<TBitpacked>& lhs, const ruy::PMat<TBitpacked>& rhs,
-           const MulParamsType& mul_params, int start_row, int start_col,
-           int end_row, int end_col, ruy::Mat<DstScalar>* dst) const {
-    TFLITE_DCHECK(false);
-  }
-};
```
AdamHillier, line 36: We've been able to remove this because the Aarch64/Arm32 kernel declarations below are now generic over `DstScalar`, which means we don't need to declare a `kNeon` path kernel without types.

larq_compute_engine/core/bgemm_kernels_arm.h
```diff
@@ -97,80 +66,47 @@ struct BgemmKernel<ruy::Path::kNeon, std::int8_t,
...
-
-// Optimised Aarch64 kernel for int8 output with 16-bit accumulators.
-template <>
-struct BgemmKernel<ruy::Path::kNeon, std::int8_t,
-                   BinaryMulParams<std::int16_t, std::int8_t>> {
+// Optimised Aarch64 kernel with 16-bit accumulators. Supports float or int8 or
+// bitpacked output.
+template <typename DstScalar>
+struct BgemmKernel<ruy::Path::kNeon, DstScalar,
+                   BinaryMulParams<std::int16_t, DstScalar>> {
```
AdamHillier, line 73: The fact that this now supports all output types means it can be generic over `DstScalar`, and we can get rid of the ugliness of having to declare each kernel multiple times for each output type.

larq_compute_engine/core/bgemm_kernels_arm32.h
```diff
@@ -144,7 +165,7 @@ void BinaryKernelNeonOutOfOrder4x4(
   int col = params.start_col;
 
   asm volatile(
-#define RUY_MAKE_ZERO(reg) "vmov.i32 " #reg ", #0\n"
+#define RUY_MAKE_ZERO(reg) "veor.s8 " #reg ", " #reg ", " #reg "\n"
```
AdamHillier, line 168: XORing a register with itself is quicker than duplicating zero into each lane.

larq_compute_engine/core/bgemm_kernels_arm64.h
```diff
@@ -819,76 +836,192 @@ void BinaryKernelNeonOutOfOrder8x4(
...
+      "ld1 {v0.4s, v1.4s}, [%[lhs_ptr]], #32\n" // In parallel, load next LHS data.
+      "cmgt v20.4s, v20.4s, v29.4s\n"
+      "cmgt v21.4s, v21.4s, v30.4s\n"
+      "cmgt v22.4s, v22.4s, v29.4s\n"
+      "cmgt v23.4s, v23.4s, v30.4s\n"
+      "ld1 {v2.4s, v3.4s}, [%[lhs_ptr]], #32\n" // In parallel, load next LHS data.
+      "cmgt v24.4s, v24.4s, v29.4s\n"
+      "cmgt v25.4s, v25.4s, v30.4s\n"
+      "cmgt v26.4s, v26.4s, v29.4s\n"
+      "cmgt v27.4s, v27.4s, v30.4s\n"
```
AdamHillier, line 885: This is a bit ugly. Basically, the accumulators are *unsigned* uint16 (even though `AccumScalar` is signed int16†), but the thresholds have to be signed integers, and thus there's no way to correctly do this comparison at int16.

The thresholds have to be signed, because there has the be a threshold value that causes a 1-bit to *always* be written (e.g. because of a small multiplier and huge negative bias for a given channel. As we write a 1-bit if and only if `accum` > `threshold`, and accumulator values are >= 0, a threshold value that forces a 1-bit to be written in all circumstances must be strictly negative.

Thus, this code I'm introducing works by extending the int16 accumulators to int32, doing the comparison there, then narrowing the result againg and collecting the bits.

An alternative solution would be to make the accumulator type actually be int16, i.e. only use this faster kernel if the maximum accumulator value is <= 2^15 - 1. Then the comparison could correctly be done at int16 without extension, but at the cost of fewer input combinations being able to use this kernel as opposed to the fallback.

† It would probably make more sense to have the accumulator types be uint32 and uint16 respectively, since that's how we actually use them.

larq_compute_engine/core/bgemm_kernels_arm64.h
```diff
@@ -974,35 +1090,27 @@ IF_FLOAT_ELIF_INT8_OUTPUT(
...
-      "str d23, [x3]\n"
-      RUY_MAKE_ZERO(v24)
-      RUY_MAKE_ZERO(v25)
-      RUY_MAKE_ZERO(v26)
-      RUY_MAKE_ZERO(v27)
-)
+      IF_FLOAT_ELIF_INT8_OUTPUT(
+            "str q20, [x3]\n"
+            "str q21, [x3, #16]\n"
+            "add x3, x3, x4\n"
```
AdamHillier, line 1096: This block is indented because it's a macro inside a macro.

larq_compute_engine/core/bgemm_kernels_common.h
```diff
@@ -67,42 +63,57 @@ struct BinaryKernelParams {
   std::int32_t rhs_stride;
   std::int32_t dst_stride;
   std::int32_t depth;
-  std::int32_t clamp_min;
-  std::int32_t clamp_max;
-  DstScalar dst_tmp_buf[LhsCols * RhsCols];
+  OutputTransform<DstScalar> output_transform;
```
AdamHillier, line 66: I've modified the kernel params to include the output transform as a member, which makes dealing with different output transforms (with different output types) easier.

larq_compute_engine/core/bgemm_kernels_arm64.h
```diff
@@ -10,6 +13,94 @@ using compute_engine::core::TBitpacked;
...
+                    RUY_OFFSET_OUTPUT_TRANSFORM_BIAS,
+                "");
+
+  // For bitpacked output
+  static_assert(OT_OFFSET + offsetof(OutputTransform<TBitpacked>, thresholds) ==
+                    RUY_OFFSET_OUTPUT_TRANSFORM_THRESHOLDS,
+                "");
+}
+
+#define MAKE_ZERO(reg) "eor " #reg ".16b, " #reg ".16b, " #reg ".16b\n"
```
lgeiger, line 86: Is this faster than `"dup " #reg ".4s, wzr`?

lgeiger (commented):
very nice 🚀

larq_compute_engine/core/bgemm_kernels_arm64.h
AdamHillier: Yeah, because NEON isn't in general optimised for transferring from general registers into NEON registers.

From the [Cortex A72 software optimisation guide](https://developer.arm.com/documentation/uan0016/a/):

`dup` from a general register has a latency of 8 cycles and a throughput of 1 instruction per cycle.

image

`eor` has a latency of 3 cycles and a throughput of 2 instructions per cycle.

image

lgeiger: Nice find :+1:

larq_compute_engine/core/bgemm_kernels_arm64.h
```diff
@@ -819,76 +821,192 @@ void BinaryKernelNeonOutOfOrder8x4(
...
+      // Test if w1 == 8 && w2 == 4, i.e. if all of the 8x4 block fits.
+      "cmp w1, w3\n"
+      "ccmp w2, w4, 0, eq\n"
+      "beq 60f\n"
+
+      // Slow path: not all the 8x4 block fits
+      "mov w15, #1\n"
+      "lsl w15, w15, w1\n"
+      "sub w15, w15, #1\n"
+      "mov v30.b[0], w15\n"
```
Tombana, line 909: (Added a comment)
```suggestion
      // Fill `v30.b[0]` with a bitstring of the form 000...000111...111 with `w1` ones, for padding.
      "mov w15, #1\n"
      "lsl w15, w15, w1\n"
      "sub w15, w15, #1\n"
      "mov v30.b[0], w15\n"
```

Tombana (approved):
Awesome :rocket:

larq_compute_engine/core/bgemm_kernels_arm.h
```diff
@@ -97,80 +66,47 @@ struct BgemmKernel<ruy::Path::kNeon, std::int8_t,
...
-};
-
-// Fallback Aarch64 kernel for int8 output with 32-bit accumulators (when
-// there's a risk of overflowing 16-bit accumulators).
-template <>
-struct BgemmKernel<ruy::Path::kNeon, std::int8_t,
-                   BinaryMulParams<std::int32_t, std::int8_t>> {
+// Fallback Aarch64 kernel with 32-bit accumulators (when there's a risk of
+// overflowing 16-bit accumulators). Supports float or int8 output.
+template <typename DstScalar>
```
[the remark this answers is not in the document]
AdamHillier: Ah nice one, I did investigate using `enable_if` in a somewhat similar way at an earlier point when developing this PR but wasn't able to get it to work properly.

Unfortunately it doesn't quite work because at some point it tries to compile this template with `TBitpacked`, then type substitution fails, and so the empty base struct gets used instead: https://github.com/larq/compute-engine/blob/18af158b4447075f6968c4acd0d30f3a13aca2df/larq_compute_engine/core/bgemm_kernels_ruy.h#L18-L19

Then at this line a proper compile error is caused: https://github.com/larq/compute-engine/blob/18af158b4447075f6968c4acd0d30f3a13aca2df/larq_compute_engine/core/bgemm_trmul_params.h#L39

because the empty base struct doesn't define a type `LhsLayout`.

I think for now let's leave it, but I might see if there's a smarter thing that we can do to solve this in #497.

Tombana (approved):

lgeiger (approved):

PR #497, opened 2020-09-09
Simplify path selection and remove unnecessary kernel registrations.

AdamHillier:
## What do these changes do?

This PR removes the overly-complicated Ruy path-search implementation and instead performs path selection with a single constexpr dependent on `AccumScalar` and `DstScalar`.  This makes path selection much cleaner, and also allows us to strip out the no-op kernels that previously had to be registered for every possible Ruy path (including all the various x86 AVX varieties).

## How Has This Been Tested?

CI.

## Benchmark Results

N/A.

## Related issue number

N/A.

PR #501, opened 2020-09-10
Fix bug introduced by #497. Make sure OPT kernels run.

AdamHillier:
## What do these changes do?

#497 accidentally disabled OPT kernels from running. This restores them.

## How Has This Been Tested?

CI.

## Benchmark Results

I've verified that the optimised kernel is running on my Android phone.

## Related issue number

#497.


lgeiger (approved):
Thanks for the fast fix, and sorry for not catching it in the review :)

PR #505, opened 2020-09-16
Add padding to temporary arrays to ensure we don't read beyond bounds.

AdamHillier:
## What do these changes do?

@Tombana and I had a conversation last week about the fact that in our (optimised) kernels we sometimes read beyond the bounds of our temporary arrays. For example, in the int16 accumulator kernel we read eight output transform multiplier/biases, even if there's only a single output channel (and so only one value in each array). This currently doesn't seem to be causing a problem in practice, but could lead to undefined behaviour and/or segfaults.

This PR adds a `LCE_EXTRA_BYTES` value (inspired by this [definition in XNNPack](https://github.com/google/XNNPACK/blob/68447302abcfad0d4b6b19a1efe7d7eef8833f4a/include/xnnpack.h#L24-L25)), set to 16, and uses it to add padding to our intermediate tensors.

## How Has This Been Tested?

CI.

## Benchmark Results

N/A.

## Related issue number

N/A.

### DIFF
--- a/larq_compute_engine/tflite/kernels/bconv2d.cc
+++ b/larq_compute_engine/tflite/kernels/bconv2d.cc
@@ -298,8 +298,10 @@ void OneTimeSetup(TfLiteContext* context, TfLiteNode* node,
   // If applicable, fuse the back-transformation and int8 scale/zero-point into
   // the output transform multiplier/bias.
   if (output->type == kTfLiteFloat32 || output->type == kTfLiteInt8) {
-    params->output_transform_multiplier.resize(params->channels_out);
-    params->output_transform_bias.resize(params->channels_out);
+    params->output_transform_multiplier.resize(params->channels_out +
+                                               LCE_EXTRA_BYTES / sizeof(float));
+    params->output_transform_bias.resize(params->channels_out +
+                                         LCE_EXTRA_BYTES / sizeof(float));
 
     const auto filter_shape = GetTensorShape(GetInput(context, node, 1));
     const std::int32_t backtransform_add =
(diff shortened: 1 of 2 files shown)


larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -284,9 +285,11 @@ void OneTimeSetup(TfLiteContext* context, TfLiteNode* node,
...
   // For 'same-zero' padding, compute the padding-correction.
   if (params->padding_type == kTfLitePaddingSame && params->pad_value == 0) {
-    params->padding_buffer.resize(ce::core::PaddingFunctor::get_cache_size(
-        params->filter_height, params->filter_width, params->channels_out,
-        params->dilation_height_factor, params->dilation_width_factor));
+    params->padding_buffer.resize(
+        ce::core::PaddingFunctor::get_cache_size(
+            params->filter_height, params->filter_width, params->channels_out,
+            params->dilation_height_factor, params->dilation_width_factor) +
+        LCE_EXTRA_BYTES);
```
AdamHillier, line 292: I believe that padding is not required for this padding correction tensor or the im2col tensor above because we don't ever read from them beyond bounds, but I've added it to be safe.

I'm also happy to remove these.

Tombana: You're right, its not required for the padding part because this is never done in assembly, only in a separate C++ pass after finishing the bgemm. I'd have a slight preference for removing it to avoid potential confusion on why it's there.

Tombana (approved):

PR #506, opened 2020-09-17
Add Python wrapper for LCE interpreter

lgeiger:
## What do these changes do?
This PR exposes a python wrapper of the LCE interpreter at `lce.testing.Interpreter` which can be used for testing and debugging.

## How Has This Been Tested?
I added unittests and switched the end2end tests to use the new `Interpreter`.


larq_compute_engine/tests/end2end_test.py
```diff
@@ -154,14 +155,10 @@ def preprocess(data):
...
-        assert len(actual_outputs) > 1
-        np.testing.assert_allclose(actual_outputs[0], actual_outputs[1], rtol=1e-5)
-        for actual_output in actual_outputs:
-            np.testing.assert_allclose(
-                actual_output, output.flatten(), rtol=0.001, atol=0.25
-            )
+    interpreter = Interpreter(model_lce)
+    actual_outputs = interpreter.predict(inputs)
+    for actual_output, output in zip(actual_outputs, outputs):
+        np.testing.assert_allclose(actual_output[0], output, rtol=0.001, atol=0.25)
```
lgeiger, line 161: This trailing, batch dimension in the output can be quite confusing. @Tombana do you think we should change that?

lgeiger: I changed the behaviour to mimic the format of `keras.Model.predic` in a15db69fd1e0dd9ddfa613799b4d1453ef13a157

PR #507, opened 2020-09-17
Add Python wrapper for LCE interpreter

lgeiger:
## What do these changes do?
This PR exposes a python wrapper of the LCE interpreter at `lce.testing.Interpreter` which can be used for testing and debugging.

## How Has This Been Tested?
I added unittests and switched the end2end tests to use the new `Interpreter`.

This is a resubmision of #506, which I accidentally closed.


larq_compute_engine/tflite/python/interpreter_wrapper_utils.h
```diff
@@ -0,0 +1,201 @@
...
+template <typename InterpreterType>
+pybind11::list InterpreterWrapperBase<InterpreterType>::get_shapes(
+    const std::vector<int>& tensors) const {
+  pybind11::list result;
+
+  for (auto tensor_id : tensors) {
+    const TfLiteTensor* tensor = interpreter_->tensor(tensor_id);
+    pybind11::tuple shape(tensor->dims->size - 1);
+    for (int j = 0; j < tensor->dims->size - 1; ++j)
+      shape[j] = tensor->dims->data[j + 1];
```
lgeiger, line 158: This won't return the empty batch dimension now, but I'd be open to revert this to the previous behaviour.

larq_compute_engine/tflite/python/interpreter.py
```diff
@@ -0,0 +1,57 @@
...
+
+    !!! example
+        ```python
+        tflite_model = convert_keras_model(model)
+        interpreter = Interpreter(tflite_model)
+        interpreter.predict(input_data, verbose=1)
+        ```
+
+    # Arguments
+        flatbuffer_model: A serialized TFLite model in the flatbuffer format.
```
AdamHillier, line 22: I don't feel very strongly about this.

```suggestion
        lce_model = convert_keras_model(model)
        interpreter = Interpreter(lce_model)
        interpreter.predict(input_data, verbose=1)
        ```

    # Arguments
        flatbuffer_model: A serialized Larq Compute Engine model in the flatbuffer format.
```

larq_compute_engine/tflite/python/interpreter.py
```diff
@@ -0,0 +1,57 @@
...
+                multiple inputs.
+            verbose: Verbosity mode, 0 or 1.
+
+        # Returns
+            Numpy array(s) of output predictions.
+        """
+
+        if not isinstance(x, (list, np.ndarray)) or len(x) == 0:
+            raise ValueError(
+                "Expected either a non-empty list of inputs or a numpy array with "
```
AdamHillier, line 41:
```suggestion
                "Expected either a non-empty list of inputs or a Numpy array with "
```

AdamHillier (approved):
Looks great :+1:

lgeiger:
I reverted 3fc3ab7602d0f674326633b7111340fdf3bcf551 since it ended up breaking our docs generation.

larq_compute_engine/tflite/python/interpreter_wrapper_utils.h
Tombana: I'd rather leave this the way it was and have the C++ pybind be as generic as possible, and then 'fix' it on the python side. TFLite has full support for `batch != 1`, and with this change we'd not support that.

larq_compute_engine/tflite/python/interpreter_wrapper_utils.h
```diff
@@ -0,0 +1,201 @@
...
+
+#include "pybind11/numpy.h"
+#include "pybind11/pybind11.h"
+#include "pybind11/stl.h"
+#include "tensorflow/lite/c/common.h"
+
+#define PY_ERROR(x)                                                           \
+  {                                                                           \
+    std::stringstream ss;                                                     \
+    ss << "ERROR at " << __FILE__ << ":" << __LINE__ << " : " x << std::endl; \
```
Tombana, line 15: (Don't apply the suggestion through the Github UI because it'll fail linting)
```suggestion
    ss << "ERROR at " << __FILE__ << ":" << __LINE__ << " : " << x << std::endl; \
```
This worked before because in C/C++ there is automatic concatenation of string literals, i.e `"ab" "c"` is equivalent to `"abc"`, and this macro was only used with `x` being a string literal. But it would be better to change it I think.

larq_compute_engine/tflite/python/interpreter_wrapper_utils.h
lgeiger: > TFLite has full support for `batch != 1`, and with this change we'd not support that.

Ah didn't know that since I only ever saw models with `batch == 1`.  Should we keep the Python API then the same as well and always return `[batch, *tensor_shape]` which we get from TFLite?

Tombana (approved):
LGTM (if the empty batch dimension fix is applied on the Python side instead of C++ side)

larq_compute_engine/tflite/python/interpreter_wrapper_utils.h
lgeiger: Applied in 60b780fbb08a3902a77dcaaa3f85429bd771f405

larq_compute_engine/tflite/python/interpreter_wrapper_utils.h
Tombana: That is fine with me, I don't feel too strongly about that.

lgeiger: OK, I reverted it to the previous behaviour. I think this now also roughly matches what Keras does:
```python
In [9]: input_shape = (24, 24, 3)
   ...: x = tf.keras.Input(input_shape)
   ...: model = tf.keras.Model(x, tf.keras.layers.Flatten()(x))

In [10]: model.input_shape
Out[10]: (None, 24, 24, 3)

In [11]: model.output_shape
Out[11]: (None, 1728)
```

Tombana (approved):

AdamHillier (approved):

PR #516, opened 2020-09-23
Add optimised 'Indirect BGEMM' binary convolution kernels.

AdamHillier:
## What do these changes do?

This PR adds a new type of binary convolution kernel that uses an 'indirect' BGEMM algorithm that doesn't require im2col. This is an adaptation of the algorithm introduced in the paper [The Indirect Convolution Algorithm](https://arxiv.org/pdf/1907.02129.pdf) and used extensively in the [XNNPack](https://github.com/google/xnnpack) library.

Only one BGEMM-micro kernel is included: a portable 4x2 kernel written in C++. However, this PR lays the groundwork for adding additional micro-kernels -- including hand-optimised and architecture-specific variations -- in the future. As such, the focus of this PR is not performance; the new kernel will be substantially slower than our existing highly-optimised im2col + BGEMM kernel, which will remain the default.

## How Has This Been Tested?

CI. The non-CI 'big' kernel tests pass locally for Aarch64 and Arm32.

## Benchmark Results

Benchmarks aren't really relevant because this PR does not change the default optimised kernel that is run, but to give a rough idea I ran QuickNet on my Raspberry Pi 4B with our three different kernel types:

| Kernel                                         | Average latency over 250 runs (ms) |
|------------------------------------------------|:----------------------------------:|
| Optimised im2col + BGEMM (hand-tuned assembly) | 30.0                               |
| Optimised indirect BGEMM (C++)  - *this PR*      | 128.8                              |
| Reference (C++)                                | 269.5                              |

## Related issue number

N/A.

### DIFF
--- a/larq_compute_engine/core/indirect_bgemm/kernel_4x2_portable.h
+++ b/larq_compute_engine/core/indirect_bgemm/kernel_4x2_portable.h
@@ -1,0 +1,245 @@
+#ifndef COMPUTE_ENGINE_INDIRECT_BGEMM_KERNEL_4x2_PORTABLE_H_
+#define COMPUTE_ENGINE_INDIRECT_BGEMM_KERNEL_4x2_PORTABLE_H_
+
+#include <cstdint>
+#include <type_traits>
+
+#include "larq_compute_engine/core/bconv2d/output_transform.h"
+#include "larq_compute_engine/core/types.h"
+#include "ruy/profiler/instrumentation.h"
+#include "tensorflow/lite/c/builtin_op_data.h"
+#include "tensorflow/lite/kernels/internal/types.h"
+
+namespace compute_engine {
+namespace core {
+namespace indirect_bgemm {
+namespace kernel_4x2_portable {
+
+/**
+ * A 4x2 C++ micro-kernel for float or int8 output.
+ */
+template <typename DstScalar>
+void RunKernel(const std::int32_t block_num_pixels,
+               const std::int32_t conv_kernel_size,
+               const std::int32_t channels_in, const std::int32_t channels_out,
+               const bconv2d::OutputTransform<DstScalar>& output_transform,
+               const TBitpacked* weights_ptr,
+               const TBitpacked** indirection_buffer, DstScalar* output_ptr) {
+  static_assert(std::is_same<DstScalar, float>::value ||
+                    std::is_same<DstScalar, std::int8_t>::value,
+                "");
+
+  ruy::profiler::ScopeLabel label("Indirect BGEMM block (4x2, portable)");
+
+  TFLITE_DCHECK_GE(block_num_pixels, 1);
+  TFLITE_DCHECK_LE(block_num_pixels, 2);
+  TFLITE_DCHECK_GE(conv_kernel_size, 1);
+  TFLITE_DCHECK_GE(channels_in, 1);
+  TFLITE_DCHECK_GE(channels_out, 1);
+
+  DstScalar* output_ptr_0 = output_ptr;
+  DstScalar* output_ptr_1 = output_ptr + channels_out;
+
+  // At the end of the output array we might get a block where the number of
+  // pixels is less than 2, if the overall output size is not a multiple of 2.
+  // When this happens we set the 'leftover' output pointer equal to the first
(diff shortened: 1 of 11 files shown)


larq_compute_engine/core/indirect_bgemm/kernel_4x2_portable.h
```diff
@@ -0,0 +1,243 @@
...
+    // the last output column value.
+    if (c_out_index % bitpacking_bitwidth == 0 || c_out_index >= channels_out) {
+      // If this is a 'leftover output channel' block (because the number of
+      // output channels isn't a multiple of four) then zero-out the extra bits.
+      if (c_out_index % bitpacking_bitwidth != 0) {
+        output_col_0 &=
+            (TBitpacked(1) << (channels_out % bitpacking_bitwidth)) - 1;
+        output_col_1 &=
+            (TBitpacked(1) << (channels_out % bitpacking_bitwidth)) - 1;
+      }
```
Tombana, line 229: The extra bits should already be zero right? Because you set the entire word to zero at the start and at the end of every iteration (otherwise the `|= (1 << i);` wouldn't work).

AdamHillier: I don't think so necessarily, because the kernel computes four bits at once, and there might be only two output channels.

Tombana: Ah yes, makes sense :+1:

larq_compute_engine/core/indirect_bgemm/prepare.h
```diff
@@ -0,0 +1,145 @@
...
+    for (int32_t output_tile_offset = 0; output_tile_offset < block_size_pixels;
+         output_tile_offset++) {
+      const int32_t output_index =
+          std::min(output_tile_start + output_tile_offset, output_size - 1);
+      const int32_t output_x = output_index % output_width;
+      const int32_t output_y = output_index / output_width;
+      for (int32_t kernel_y = 0; kernel_y < kernel_height; kernel_y++) {
+        const int32_t input_y = output_y * stride_height +
+                                kernel_y * dilation_height - input_padding_top;
+        if (0 <= input_y && input_y < input_height) {
```
lgeiger, line 64: Not sure how performance critical this part is, so feel free to ignore this comment. But this should remove one comparison in the generated code:
```suggestion
        // (0 <= input_y && input_y < input_height)
        if ((unsigned)input_y < (unsigned)input_height) {
```

larq_compute_engine/core/indirect_bgemm/prepare.h
```diff
@@ -0,0 +1,145 @@
...
+          for (int32_t kernel_x = 0; kernel_x < kernel_width; kernel_x++) {
+            const int32_t input_x = output_x * stride_width +
+                                    kernel_x * dilation_width -
+                                    input_padding_left;
+            const int32_t kernel_index = kernel_y * kernel_width + kernel_x;
+            const int32_t index = output_tile_start * kernel_size +
+                                  kernel_index * block_size_pixels +
+                                  output_tile_offset;
+            if (0 <= input_x && input_x < input_width) {
```
lgeiger, line 73:
```suggestion
            // (0 <= input_x && input_x < input_width)
            if ((unsigned)input_x < (unsigned)input_width) {
```

larq_compute_engine/core/indirect_bgemm/prepare.h
AdamHillier: Wow, I've not seen that trick before. How does it work exactly?

The functions in `prepare.h` aren't performance critical because they only run once at the start, so I think I might leave it as-is because it's easier to understand.

lgeiger: I got this from https://github.com/python/cpython/pull/9784 which links to section 14.2 "Bounds Checking" in https://www.agner.org/optimize/optimizing_cpp.pdf for an in-depth explanation.

AdamHillier: Ah nice, that's cool, thanks for the links :)

Tombana (approved):
:rocket:

PR #524, opened 2020-10-04
Initialise batch norm stats for e2e models

AdamHillier:
## What do these changes do?

The default BatchNormalization moving-variance initialisation is `"ones"`, which makes sense for a full-precision convolution (because the weights are usually initialised with very small absolute value) but doesn't make sense for a binary convolution (where the absolute value of the weights is always 1). This causes instability in the end2end tests because it means that the magnitude of the convolution outputs is unrealistic. In places, we've attempted to correct for this by fiddling with the gamma/beta initialisation, but the real problem is the moving-variance statistics.

This PR adds ten steps of training with a low learning rate for each model in the e2e test (excluding QuickNet, which we load with trained weights). These few steps should result in more sensible values for the batch norm statistics, and regardless I think this makes the tests more relevent that using models with entirely random weights.

A nice benefit is that it appears that this allows us to reduce the absolute tolerance in the `np.allclose` comparison.

## How Has This Been Tested?

CI.

## Benchmark Results

N/A.

## Related issue number

N/A.

### DIFF
--- a/larq_compute_engine/tests/end2end_test.py
+++ b/larq_compute_engine/tests/end2end_test.py
@@ -116,26 +116,12 @@ def toy_model_int8(**kwargs):
     x = lq.layers.QuantConv2D(
         12, 3, strides=2, input_quantizer="ste_sign", kernel_quantizer="ste_sign"
     )(x)
-    # Make sure the typical output is in the (-3, 3) range
-    # by dividing by sqrt(filter_height * filter_width * input_channels)
-    x = tf.keras.layers.BatchNormalization(
-        gamma_initializer=tf.keras.initializers.RandomNormal(
-            1.0 / math.sqrt(3 * 3 * 3), stddev=0.1 / math.sqrt(3 * 3 * 3)
-        ),
-        beta_initializer="uniform",
-    )(x)
+    x = tf.keras.layers.BatchNormalization(momentum=0.7)(x)
     x = quant(x)
     x = lq.layers.QuantConv2D(
         12, 3, strides=2, input_quantizer="ste_sign", kernel_quantizer="ste_sign"
     )(x)
-    # Make sure the typical output is in the (-3, 3) range
-    # by dividing by sqrt(filter_height * filter_width * input_channels)
-    x = tf.keras.layers.BatchNormalization(
-        gamma_initializer=tf.keras.initializers.RandomNormal(
-            1.0 / math.sqrt(3 * 3 * 12), stddev=0.1 / math.sqrt(3 * 3 * 12)
-        ),
-        beta_initializer="uniform",
-    )(x)
+    x = tf.keras.layers.BatchNormalization(momentum=0.7)(x)
     x = quant(x)
     x = tf.keras.layers.GlobalAveragePooling2D()(x)
     x = quant(x)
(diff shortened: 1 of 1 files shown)


larq_compute_engine/tests/end2end_test.py
```diff
@@ -152,35 +145,45 @@ def toy_model_int8(**kwargs):
...
 def preprocess(data):
-    return lqz.preprocess_input(data["image"])
+    return lqz.preprocess_input(data["image"]), data["label"]
 
 
 def assert_model_output(model_lce, inputs, outputs):
     interpreter = Interpreter(model_lce, num_threads=min(os.cpu_count(), 4))
     actual_outputs = interpreter.predict(inputs)
-    np.testing.assert_allclose(actual_outputs, outputs, rtol=0.001, atol=0.25)
+    np.testing.assert_allclose(actual_outputs, outputs, rtol=0.05, atol=0.125)
```
AdamHillier, line 154: I don't think it really made sense to have such a small rtol but huge atol (with rtol/atol 0.001/0.25, the rtol would only really be relevant for values with absolute value > 100), so I've bumped up the rtol as well as reducing the atol.

Tombana: I agree that `0.25` is huge. However it's what we observed for the keras-tflite difference for large models like Quicknet.
I think it might be better to pass a tolerance (or `Default`) to each model in `model_cls` so that we have normal tolerances everywhere and make it higher exclusively for QuickNet. What do you think?

AdamHillier: Yeah I agree, that's a good idea :+1:

larq_compute_engine/tests/end2end_test.py
```diff
@@ -152,43 +131,67 @@ def toy_model_int8(**kwargs):
...
+        rtol = 0.05
+        atol = 0.2
+    elif model_cls == toy_model_int8:
+        # 0.025 atol is chosen because it allows an off-by-one error in the int8
+        # model (with +-3 scales) but not off-by-two.
+        rtol = 0.01
+        atol = 0.025
+    else:
+        rtol = 0.001
+        atol = 0.001
```
AdamHillier, line 183: @Tombana yeah you're right, turns out we can use *way* lower tolerances for the non-QuickNet models. And probably should do.

Tombana (approved):

larq_compute_engine/tests/end2end_test.py
```diff
@@ -152,43 +130,67 @@ def toy_model_int8(**kwargs):
...
     [toy_model, toy_model_sequential, toy_model_int8, lqz.sota.QuickNet],
 )
 def test_simple_model(model_cls):
-    model = model_cls(weights="imagenet")
-    model_lce = convert_keras_model(
-        model, experimental_enable_bitpacked_activations=True
-    )
-
-    # Test on the flowers dataset
+    # Test on the OxfordFlowers dataset
```
lgeiger, line 147: This is not the same as oxford flowers, see: https://www.tensorflow.org/datasets/catalog/tf_flowers
```suggestion
    # Test on the TF flowers dataset
```

lgeiger (approved):
Nice!

PR #526, opened 2020-10-07
Add support for unipolar bconv2d inputs.

AdamHillier:
*This is very much a draft PR.*

## What do these changes do?

This is a converter-only change that adds support for binary convolutions with unipolar inputs -- all credit is given to @jneeven who had the original idea for the trick that this PR takes advantage of and that allows this to be a converter-only change, i.e. that avoids the need to modify the kernels. I've tried to explain the method in the code comment above the pattern I've added in `prepare_patterns.td`.

Specifically, it adds support for models with the following input quantiser:

```python
class UnipolarSteSign(lq.quantizers.SteSign):
    def __init__(self, clip_value: float = 0.5, **kwargs):
        super().__init__(clip_value=clip_value, **kwargs)

    def call(self, inputs):
        return 0.5 * super().call(inputs - 0.5) + 0.5
```

The end2end test results seem to indicate that the method works.

I'm publishing this as an early draft because I am keen to solicit feedback on whether this is a sensible approach; if we want to progress further we will need to add the quantiser to Larq.

Still to do:

- [ ] Add file-check tests.
- [ ] Add a `UnipolarSteSign` quantiser to Larq.
- [ ] Do a validation run with a unipolar model and check that we get the expected accuracy.

## How Has This Been Tested?

The end2end tests have been modified to test unipolar inputs. These pass locally, and hopefully on CI too, with the much tighter ouput bounds introduced by #524.

## Benchmark Results

N/A.

## Related issue number

N/A.

### DIFF
--- a/larq_compute_engine/mlir/transforms/prepare_patterns.td
+++ b/larq_compute_engine/mlir/transforms/prepare_patterns.td
@@ -40,6 +40,8 @@ def GetNumChannels : NativeCodeCall<"GetNumChannels($_builder, $0)">;
 def IsDataFormatNHWC : ConstantAttr<TF_ConvnetDataFormatAttr, "NHWC">;
 def CreateNoneAttrValue : NativeCodeCall<"$_builder.getUnitAttr()">;
 
+// Match a binary convolution with bipolar (+1/-1) inputs, padded (if
+// applicable) with zeroes.
 def : Pat<(TF_Conv2DOp
               (LQ_DequantizeOp: $dequantized_input $input),
               (ConstantOp $filter),
@@ -77,6 +79,7 @@ def ConstFloatValueIsOne : Constraint<
 
 def SamePadding : Constraint<CPred<"IsSamePadding($0, $1, $2, $3)">>;
 
+// Match a binary convolution with bipolar (+1/-1) inputs, padded with ones.
 def : Pat<(TF_Conv2DOp:$output
               (TF_PadV2Op
                   (LQ_DequantizeOp: $dequantized_input $input),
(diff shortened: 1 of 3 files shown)


lgeiger:
That's really cool, I completely forgot about this 🎉 

In larq we already have the unipolar [`SteHeaviside`](https://docs.larq.dev/larq/api/quantizers/#steheaviside) quantizer (although I think the backward pass should actually be a [0, 1] clip instead of the [-1, 1] clipping which is currently used). The forward pass is implemented in [`lq.math.heaviside`](https://docs.larq.dev/larq/api/math/#heaviside) as `tf.sign(tf.nn.relu(x))` so we could reuse this implementation without needing to modify larq itself.

larq_compute_engine/mlir/transforms/prepare_patterns.td
```diff
@@ -110,3 +113,70 @@ def : Pat<(TF_Conv2DOp:$output
...
+// Note that the bias `0.5 * w` is a constant at conversion-time. The
+// generalisation to more than a single activation/weight element is then:
+//         `dotprod(a, w) = 0.5 * dotprod(b, w) + 0.5 * reduce_sum(w)`
+// which is the form we use below.
+//     There is one final thing to note here. Because the unipolar input 1 is
+// represented by bipolar input 1 and unipolar input 0 by bipolar input -1, TF
+// zero-padding is equivalent to padding with -1 in LCE; however, the bconv2d op
+// only supports padding with 1. The solution is to multiply the (bipolar) input
+// by -1 and the weights by -1 -- this means that TF zero-padding is equivalent
+// to LCE bconv2d one-padding.
```
Tombana, line 135: Instead of the "multiplication by -1" argument, can you not do this directly?
Let `b=1` if `a=0` and `b=-1` if `a=1` (i.e. flipped from what you write here). Then `a = -0.5 * b + 0.5` (just flips the signs here). The rest of your computation is similar then:
`dotprod(a,w) = -0.5 * dotprod(b, w) + 0.5 * reduce_sum(w)`.
Its entirely the same as what you have but it avoid the extra "multiplication by -1" argument for zero-padding so it might be simpler to understand for the reader of the comment.

Tombana:
I'm confused as to why the end2end test works with `ReLu`. Basically we have the following:
```
UnipolarDotProd = -0.5 * BinaryDotProd + constant
```
(per output-channel, and the `constant` depends on the output channel).
We can compute `BinaryDotProd` and then fuse the `-0.5` and `constant` into the `post_activation_{multiplier_bias}`. However if there's a ReLu then for the unipolar conv, Keras expects `ReLu(UnipolarDotProd)`. But
```
ReLu(UnipolarDotProd) != -0.5 * ReLu( BinaryDotProd ) + constant
```
where the right-hand-side is what the bconv op would compute.

If we don't write it as `ReLu` but as a clamp with arbitrary clamp values, it might work (I'm not 100% sure, it might become channel-dependent). So maybe a solution would be to let the converter compute clamp-threshold values instead of an enum with 'None, ReLu, ...'. Previously we argued against this because it'd expose implementation details to the API. However if it would solve this issue we might want to rethink that. We could also add a flag `unipolar = {False/True}` to the API and then the op itself could recompute those clamp threshold values.

larq_compute_engine/mlir/transforms/prepare_patterns.td
AdamHillier: I originally had something very similar to what you are suggesting here. It forces you to pattern match against a weird formulation of the Larq quantizer, e.g. `-0.5 * (threshold - input) + 0.5`, which is equivalent to having pushed through a -1, the only difference is that the -1 is 'baked in' to the quantiser definition.

AdamHillier:
> I'm confused as to why the end2end test works with `ReLu`. 

Ah yes, I should have mentioned this in the PR. Doing this transformation prevents ReLU fusing, because it adds a non-trivial bias, and we only fuse ReLUs when the bconv2d has a zero bias. So the reason the test passes is that the ReLU isn't fused.

jneeven:
> That's really cool, I completely forgot about this tada
> 
> In larq we already have the unipolar [`SteHeaviside`](https://docs.larq.dev/larq/api/quantizers/#steheaviside) quantizer (although I think the backward pass should actually be a [0, 1] clip instead of the [-1, 1] clipping which is currently used). The forward pass is implemented in [`lq.math.heaviside`](https://docs.larq.dev/larq/api/math/#heaviside) as `tf.sign(tf.nn.relu(x))` so we could reuse this implementation without needing to modify larq itself.

The `DoReFaQuantizer` can be used for unipolar activations, and I think we did so without any modifications at all (simply set the precision to 1). However, we empirically found LSQ works better, but haven't added it to Larq yet. Since LSQ only has a learnable scale parameter and no offset, there should be no difference between the two in terms of inference, and it probably make more sense to use either of these quantizers than to add a new `UnipolarSteSign`.

AdamHillier:
> That's really cool, I completely forgot about this 🎉
> 
> In larq we already have the unipolar [`SteHeaviside`](https://docs.larq.dev/larq/api/quantizers/#steheaviside) quantizer (although I think the backward pass should actually be a [0, 1] clip instead of the [-1, 1] clipping which is currently used). The forward pass is implemented in [`lq.math.heaviside`](https://docs.larq.dev/larq/api/math/#heaviside) as `tf.sign(tf.nn.relu(x))` so we could reuse this implementation without needing to modify larq itself.

Ah cool, I hadn't realised that, so that relies on the fact that `tf.sign(0) = 0`? Doesn't that make it a bit weird in the sense that without `lq.math.sign` it doesn't generate a quant/dequant pair. I'm definitely happy to investigate using a definition like that though.

larq_compute_engine/mlir/transforms/prepare_patterns.td
AdamHillier: Incidentally, the multiplication by -1 isn't needed if you do unipolar + one-padding rather than unipolar + zero-padding.

CNugteren:
Closing this one due to inactivity. We can re-open later if someone wants to work on this.

PR #530, opened 2020-10-14
Add macros to help GCC with branch prediction

lgeiger:
## What do these changes do?
This PR adds macros to help GCC with branch prediction in cases where we know which branch is more likely to be taken. Both [TensorFlow](https://github.com/tensorflow/tensorflow/blob/fccc8b6179a24daf3d9779d236a4143b618a0a38/tensorflow/core/platform/macros.h#L105-L116) and [XNNPack](https://github.com/google/XNNPACK/blob/31677ad790d003ecfd94bbaf8ce88a2bc6847896/src/xnnpack/common.h#L171-L177) use similar macros which rely on [`__builtin_expect`](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html).

This also adds a [`FastBoundsCheck`](https://github.com/tensorflow/tensorflow/blob/fccc8b6179a24daf3d9779d236a4143b618a0a38/tensorflow/core/framework/bounds_check.h#L26-L37) check function to do comparisons like `0 <= x && x < bound` in a single instruction.

The actual impact of these helpers here is a bit questionable, but I would like to try them for some of our internal kernels and adding them here seems like the logical place.

## How Has This Been Tested?
CI

## Benchmark Results
I ran a quick benchmark with the indirect bgemm and didn't se any measurable impact

### DIFF
--- a/larq_compute_engine/core/types.h
+++ b/larq_compute_engine/core/types.h
@@ -9,6 +9,14 @@
 namespace compute_engine {
 namespace core {
 
+#if defined(__GNUC__)
+#define LCE_LIKELY(condition) (__builtin_expect(!!(condition), 1))
+#define LCE_UNLIKELY(condition) (__builtin_expect(condition, 0))
+#else
+#define LCE_LIKELY(condition) (condition)
+#define LCE_UNLIKELY(condition) (condition)
+#endif
+
 // In our kernels we may occasionally read (but never write) beyond the end of
 // an array. This is the maximum number of extra bytes that will be read, and
 // should be added as padding to the end of tensor allocations.
--- a/larq_compute_engine/core/indirect_bgemm/kernel_4x2_portable.h
+++ b/larq_compute_engine/core/indirect_bgemm/kernel_4x2_portable.h
@@ -87,7 +87,7 @@ void RunKernel(const std::int32_t block_num_pixels,
       } while (--c_in_index > 0);
     } while (--k_size_index > 0);
 
-    if (channels_out - c_out_index >= 4) {
+    if LCE_LIKELY (channels_out - c_out_index >= 4) {
       output_ptr_1[0] = output_transform.Run(acc_01, c_out_index);
       output_ptr_1[1] = output_transform.Run(acc_11, c_out_index + 1);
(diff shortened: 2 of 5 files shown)


AdamHillier:
Ah yeah I didn't include these branch prediction intrinsic things originally because I didn't see any difference in the generated assembly code (at least for the example I tried). I don't think there's any issue including them though.

larq_compute_engine/core/types.h
```diff
@@ -9,6 +9,21 @@
...
+#define LCE_LIKELY(condition) (condition)
+#define LCE_UNLIKELY(condition) (condition)
+#endif
+
+// Check that 0 <= index < limit using a single comparison, assuming
+// that 0 <= limit if Index is signed.  Intended for use in performance
+// critical contexts where 0 <= index < limit is almost always true.
+inline bool FastBoundsCheck(const int index, const int limit) {
+  return LCE_LIKELY((unsigned)index < (unsigned)limit);
+}
```
AdamHillier, line 25: I think this might rely on undefined behaviour. Definitely prior to [C++20](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0907r4.html), there was no requirement for signed integer types to use twos complement, which I think this trick relies on?

If this code were being used in performance-critical code I think there would be a strong argument for it, but the two uses are in `prepare.h` in functions which only get called once at the very start, not per-inference.

lgeiger: Makes sense, I remove their usage.

AdamHillier (approved):

Tombana (approved):
> This also adds a FastBoundsCheck check function to do comparisons like 0 <= x && x < bound in a single instruction.

I don't see that in the PR?

lgeiger:
> This also adds a FastBoundsCheck check function to do comparisons like 0 <= x && x < bound in a single instruction.

I don't see that in the PR?


Ahh sorry, I removed it again after discussing it in https://github.com/larq/compute-engine/pull/530#discussion_r504605286

Do you think we should keep the function and just not use it in prepare?

Tombana:
> Do you think we should keep the function and just not use it in prepare?

No this is fine, I simply wanted to make sure you didn't forget to commit a file.

PR #529, opened 2020-10-12
Upgrade TensorFlow to latest version

lgeiger:
## What do these changes do?
This PR upgrades TensorFlow to the latest nightly version.

## TODO
- [x] Upgrade MLIR converter
- [x] Update ruy usage (mainly https://github.com/google/ruy/commit/b7649fa263e550ee83cf190cb4830382969ad0aa and https://github.com/google/ruy/commit/5b496e04b69b44edf88687970d9378ad8f6aca3b)
- [x] Benchmark to ensure we don't introduce any unwanted regressions

## How Has This Been Tested?
CI and manually verified conversion of QuickNet.

## Benchmark Results
I benchmarked QuickNet on Android single and multithreaded and didn't see any obvious performance regressions.

PR #532, opened 2020-10-14
Prebuild the benchmark binary with -O3 on Android

lgeiger:
## What do these changes do?
This PR makes sure that the Android benchmarking binary get's built with all optimizations enabled. Previously we also set `-O3` in the Raspberry PI configs, I moved this now the the release built in the hope to slightly speedup ARM CI.

## Benchmark Results
I didn't see any measurable difference on my system, but the generated binary has a slightly different size, so it is definitely doing something.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -23,11 +23,11 @@ build:windows --host_cxxopt=/std:c++14
 # These can be activated using --config=rpi3 and --config=aarch64
 build:rpi3 --crosstool_top=@local_config_arm_compiler//:toolchain
 build:rpi3 --cpu=armeabi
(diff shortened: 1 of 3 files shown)


AdamHillier (approved):

lgeiger:
Looks like this change enabled debug checks for the ARM unit tests which now fail at:
https://github.com/larq/compute-engine/blob/fb25c8f3511c0f0de662c2e4f8c6c20f75042779/larq_compute_engine/core/bgemm/ruy_pack.h#L48-L50

I think #529 touches the same code, I can try and debug what's going on after that has been merged.

AdamHillier:
Ah yes I had long suspected that debug checks were for some reason disabled, and this confirms it haha. That specific check by the way was added by me and I'm pretty sure it can be removed.

lgeiger:
> That specific check by the way was added by me and I'm pretty sure it can be removed.

I'm fine with removing the check.

The matrices are prepared in:
https://github.com/larq/compute-engine/blob/590f8efadb616bed7f237c291d1ead06143c8600/larq_compute_engine/core/bgemm/bgemm.h#L80

Which sets the sums [here](https://github.com/google/ruy/blob/cd4f776c833964909c771e3f31c37ffcdf66e720/ruy/prepare_packed_matrices.cc#L85). Do you think we should remove this instead, or would this break things?

AdamHillier:
> Do you think we should remove this instead, or would this break things?

I'm not sure what the impact of removing that call would be, so for now I would probably just remove the debug check.

AdamHillier (approved):

PR #533, opened 2020-10-14
Change qemu test scripts by using a new bazel rule

Tombana:
## What do these changes do?
The old qemu test scripts would glob (`*`) for all `*_test` files in a directory. This PR adds a `lce_qemu_test_suite` bazel rule that will create individual bazel test targets for all binaries so that the script can simply run one binary.

## How Has This Been Tested?
CI


lgeiger (approved):
Thanks for the cleanup 🚀

PR #543, opened 2020-10-21
Add __version__ property.

AdamHillier:
See larq/larq#596.

### DIFF
--- a/larq_compute_engine/mlir/python/converter_test.py
+++ b/larq_compute_engine/mlir/python/converter_test.py
@@ -5,6 +5,8 @@ from unittest import mock
 import larq_zoo as lqz
 from tensorflow.python.eager import context
 
+sys.modules["importlib.metadata"] = mock.MagicMock()
+sys.modules["importlib_metadata"] = mock.MagicMock()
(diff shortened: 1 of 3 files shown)


larq_compute_engine/mlir/python/converter_test.py
```diff
@@ -10,12 +10,20 @@
...
 from larq_compute_engine.mlir._graphdef_tfl_flatbuffer import (
     convert_graphdef_to_tflite_flatbuffer as mocked_converter,
 )
 
 
+def test_version():
+    assert (
+        hasattr(larq_compute_engine, "__version__")
+        and "." in larq_compute_engine.__version__
+    )
```
AdamHillier, line 24: @lgeiger is this test failing because we don't actually install the package, so setup.py isn't being run?

lgeiger: Yes we don't actually install LCE in the Python only tests since it would require us to build all the C++ dependencies.

AdamHillier: Makes sense, then I think I'll just remove the test because I don't think there's much extra value including it.

lgeiger (approved):

lgeiger (approved):

PR #544, opened 2020-10-21
Use test sharding to parallelise the tests.

AdamHillier:
## What do these changes do?

Our CI runs on machines with two cores. Bazel supports [test sharding](larq_compute_engine/mlir/python/converter_test.py) which when enabled parallelises the tests over multiple processes. All we need to do is add a `shard_count` attribute to the test definition.

Note that the 'small' kernel tests which run on CI don't test multithreaded convolutions, so there's no problem testing two in parallel.

### DIFF
--- a/larq_compute_engine/tests/qemu_test.bzl
+++ b/larq_compute_engine/tests/qemu_test.bzl
@@ -35,10 +35,11 @@ def lce_qemu_test_suite(
         # Finally create a new sh_test target
         native.sh_test(
             name = sh_name,
-            size = "medium",
+            size = "large",
             srcs = [src],
             args = [test_path],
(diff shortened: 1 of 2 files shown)


larq_compute_engine/tests/qemu_test.bzl
```diff
@@ -35,10 +35,11 @@ def lce_qemu_test_suite(
         # Finally create a new sh_test target
         native.sh_test(
             name = sh_name,
-            size = "medium",
+            size = "large",
```
AdamHillier, line 38: I've made this "large" because otherwise the 'big' kernel tests which don't run on CI timeout.

lgeiger (approved):
Awesome!

PR #545, opened 2020-10-22
:arrow_up: tensorflow@r2.4

lgeiger:
This PR updates TensorFlow to point to the [r2.4](https://github.com/tensorflow/tensorflow/tree/r2.4) branch which will be used to tag the 2.4 releases.

PR #546, opened 2020-10-22
Auto-generate the TFLite Python schema.

AdamHillier:
## What do these changes do?

This PR uses a Bazel genrule to generate the Python TFLite schema file and copy it into the correct place during compilation, which means we can remove the checked-in file.

I'm not 100% sure it works because the end2end test takes ages to compile on my local machine, but running `bazel build -c opt //larq_compute_engine/mlir:tflite_schema_py` locally does generate the correct file as `bazel-bin/larq_compute_engine/mlir/python/tflite_schema.py`, so hopefully it works! Let's see what CI thinks.

~~This is a draft, and is based on #545.~~

## How Has This Been Tested?

CI.

## Benchmark Results

N/A.

## Related issue number

N/A.


Tombana (approved):
:rocket:

lgeiger (commented):
Great, having to checkin this file has annoyed me for quite a while!
The `@org_tensorflow//tensorflow/lite/python:schema_py` used to generate broken imports, glad that this has been fixed in the latest version now.

Do we need to add `python/tflite_schema.py` to `.gitignore` or is it only included in the `bazel` exec_dir?

Does this work correctly when building the `pip` packages? I think it should, but not 100% sure.

AdamHillier:
> Do we need to add `python/tflite_schema.py` to `.gitignore` or is it only included in the `bazel` exec_dir?

I don't think this should be necessary, it's only in the bazel outputs.

> Does this work correctly when building the `pip` packages? I think it should, but not 100% sure.

Yes I agree we need to check that. Once CI passes here I'm going to trigger a nightly release build and see what happens.

lgeiger (approved):

AdamHillier:
Good and bad news:
* The test-build I did worked for Mac; I downloaded and installed the built Pip package and tested that it works correctly to convert a model.
* The Linux test-build failed, with the error
    ```
    020-10-22T19:08:34.4885188Z bazel-out/k8-opt/bin/external/local_config_python/python_include/pyconfig.h:3:12: fatal error: x86_64-linux-gnu/python3.6m/pyconfig.h: No such file or directory
    2020-10-22T19:08:34.4886028Z  #  include <x86_64-linux-gnu/python3.6m/pyconfig.h>
    2020-10-22T19:08:34.4886356Z             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    2020-10-22T19:08:34.4886606Z compilation terminated.
    2020-10-22T19:08:34.5977016Z Target //:build_pip_pkg failed to build
    2020-10-22T19:08:34.5985991Z Use --verbose_failures to see the command lines of failed build steps.
    2020-10-22T19:08:35.1872804Z INFO: Elapsed time: 13429.448s, Critical Path: 142.10s
    ```
    which doesn't sound like a problem that could be caused by this PR - a more likely culprit is the recent TF2.4 upgrade. [This comment](https://github.com/okfn/piati/issues/65#issuecomment-39722476) suggests that it means the Docker container doesn't include all the necessary dependencies.

lgeiger:
Sounds good, thanks for testing!

Maybe https://github.com/larq/compute-engine/commit/ac3afd2b255840e952a4093a8c8a22edecf1d743#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R158 broke the build or we need an newer Docker image. In any case let's merge, I can take a look at the failure next week. TF or TFA might run into a similar issue so we can take a lot at their build config.

Issue #406, opened 2020-06-15
Add MLIR verifier to BConv op


lgeiger:
Since #384 we can make use of the [IR verification](https://mlir.llvm.org/docs/OpDefinitions/#custom-verifier-code) features of MLIR.

Our BConv op has a few parameter combinations where [it can throw during `Init` or `Prepare`]( https://github.com/larq/compute-engine/blob/master/larq_compute_engine/tflite/kernels/bconv2d.cc). In the converter we try to make sure that none of those cases will ever occur, but it would be good to add verification code so that potential errors are already caught during conversion and not at runtime which makes for a much nicer user experience if something goes wrong (especially when relying one some of our experimental features).

This can be done very nicely by populating the following verifier boilerplate with the relevant constraints of our TFLite kernel:
```diff
diff --git a/larq_compute_engine/mlir/ir/lce_ops.cc b/larq_compute_engine/mlir/ir/lce_ops.cc
index 4283888..9ec481a 100644
--- a/larq_compute_engine/mlir/ir/lce_ops.cc
+++ b/larq_compute_engine/mlir/ir/lce_ops.cc
@@ -47,5 +47,11 @@ LarqDialect::LarqDialect(MLIRContext* context)
 #include "larq_compute_engine/mlir/ir/lce_ops.cc.inc"
       >();
 }
+
+LogicalResult Verify(Bconv2dOp op) {
+  return mlir::success();
+  return op.emitOpError("Something is not correct!");
+}
+
 }  // namespace TF
 }  // namespace mlir
diff --git a/larq_compute_engine/mlir/ir/lce_ops.td b/larq_compute_engine/mlir/ir/lce_ops.td
index d45ef14..a5c54ec 100644
--- a/larq_compute_engine/mlir/ir/lce_ops.td
+++ b/larq_compute_engine/mlir/ir/lce_ops.td
@@ -111,6 +111,8 @@ TODO
   let results = (outs
     TensorOf<[F32, I32, QI8]>:$output
   );
+
+  /* let verifier = [{ return Verify(*this); }]; */
 }

 def LQ_BMaxPool2dOp : LQ_Op<"BMaxPool2d", [NoSideEffect]> {
```

I am labeling this as a "good first issue" since it is a well scoped problem and we are happy to assist if anyone wants to tackle this.

PR #549, opened 2020-10-22
Add grouped binary convolution support (1/3): the converter.

AdamHillier:
## What do these changes do?

This is first of a group of PRs to add support for grouped binary convolutions. I've split the work into three PRs to make review easier.

First, this PR adds support to the converter, by adding an attribute `groups` to the op definition. The correct value for `groups` is infered by dividing the input channels by the `channels_in` dimension of the filter. An error is raised if the size of each group (the input channels divided by `groups`) is not a multiple of 32 -- I have chosen to add this constraint because it significantly simplifies the implementation in the kernels.

## How Has This Been Tested?

MLIR file-check tests have been added to check that the attribute is correctly set, and that the error is raised if the group-size is not a multiple of 32.

## Benchmark Results

N/A.

## Related issue number

#550, #551.

### DIFF
--- a/larq_compute_engine/mlir/transforms/prepare_tf.cc
+++ b/larq_compute_engine/mlir/transforms/prepare_tf.cc
@@ -1,3 +1,4 @@
+#include "larq_compute_engine/core/types.h"
 #include "larq_compute_engine/mlir/ir/lce_ops.h"
 #include "mlir/Dialect/StandardOps/IR/Ops.h"
 #include "mlir/IR/PatternMatch.h"
@@ -11,6 +12,8 @@ namespace mlir {
 namespace TFL {
 
 namespace {
+
+using compute_engine::core::bitpacking_bitwidth;
 
 // Prepare LCE operations in functions for subsequent legalization.
 struct PrepareLCE : public PassWrapper<PrepareLCE, FunctionPass> {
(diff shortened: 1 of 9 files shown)


larq_compute_engine/mlir/transforms/prepare_tf.cc
```diff
@@ -109,6 +112,38 @@ IntegerAttr GetNumChannels(Builder& b, Value output_val) {
...
+
+  auto num_groups = total_input_channels / filter_input_channels;
+  if (total_input_channels % filter_input_channels != 0) {
+    mlir::emitError(filter_val.getLoc())
+        << "Filter dimensions invalid: the number of filter input channels "
+        << filter_input_channels
+        << " does not divide the total number of input channels "
+        << total_input_channels << "\n";
+    num_groups = 1;
+  }
```
AdamHillier, line 135: Technically I don't think this is required, because the Conv2D op throws an error if this condition is violated, but it's probably better safe than sorry.

larq_compute_engine/mlir/transforms/prepare_tf.cc
```diff
@@ -109,6 +112,38 @@ IntegerAttr GetNumChannels(Builder& b, Value output_val) {
...
+  if (num_groups > 1 && filter_input_channels % bitpacking_bitwidth != 0) {
+    mlir::emitError(filter_val.getLoc())
+        << "Invalid binary group convolution: the number of input channels "
+           "per-group "
+           "must be a multiple of "
+        << bitpacking_bitwidth << ", but is " << filter_input_channels << "\n";
+    num_groups = 1;
+  }
```
AdamHillier, line 143: We could choose not to throw an error here, and simply not apply the transformation if this condition is violated, so that there will be a float op in the model. I'm not sure what the right approach is.

AdamHillier:
Upon reflection and discussions with @Tombana, there's actually technically no need to have a `groups` argument, but only because of the constraint that the input-depth-per-group (if the number of groups is greater than one) is a multiple of 32. If we know we'll never see a flatbuffer with a grouped convolution that has a non-multiple-of-32 group size, then we can infer the number of groups from the (bitpacked) filter channels in dimension and (bitpacked) input channels in dimension.

I'm not sure what the right thing to do here is. Conciseness is nice, but I worry that relying on being able to infer the number of groups by relying on that constraint being satisfied might lead to subtle bugs in the future.

Tombana:
Even if we can infer it in theory, I think it's good to have it as a separate parameter, even if only as an extra correctness checksum.

lgeiger:
> I'm not sure what the right thing to do here is. Conciseness is nice, but I worry that relying on being able to infer the number of groups by relying on that constraint being satisfied might lead to subtle bugs in the future.

I'd actually prefer not to have a dedicated `group` attribute in the op. TensorFlow does the same for their Conv2DOp (although in their case I probably would've prefered a dedicated groups argument for the user facing code for better validation).

I think we can do all the necessary validation in `Prepare` even without a group argument, since we already have a `channels_in` argument and the shapes of the bitpacked input and bitpacked filters, or am I missing something a case that we can't catch that could lead to subtle bugs?

In both cases we should always make sure that the number of filters per group is divisible by 32 in the converter anyway.

AdamHillier:
I think I'm onboard with that.

In case somebody tries to convert a grouped Larq bconv with non-multiple-of-32 group size, do you think we should throw a converter error (as this PR currently does) or fall back to an emulated bsign + float conv?

lgeiger:
> In case somebody tries to convert a grouped Larq bconv with non-multiple-of-32 group size, do you think we should throw a converter error (as this PR currently does) or fall back to an emulated bsign + float conv?

I think it's fair to throw an error for now since TFLite doesn't support group convolutions anyway so it would fail during runtime anyway.

AdamHillier:
@lgeiger I've updated the PR so that it only performs the check, rather than adding the groups attribute.

lgeiger (approved):
This is great!

Do you think it would make sense to move this check into an op verifier for the bconv op as mentioned in #406 instead? This way the check would be op specific instead of limited to this transformation. Although I am not sure if we still have all of the information there since the bconv op will already have bitpacked inputs and migth either have float or binary filters.

AdamHillier:
> Do you think it would make sense to move this check into an op verifier for the bconv op as mentioned in #406 instead? This way the check would be op specific instead of limited to this transformation. Although I am not sure if we still have all of the information there since the bconv op will already have bitpacked inputs and migth either have float or binary filters.

I really like this idea in theory, but you're right that it won't be possible once the filters have been bitpacked. E.g. the number of input channels could be 64, and the bitpacked channels in dimension could be 1, and we don't know if that's because it's 32 input channels to the filter (which is valid) or only 1 that's been padded by bitpacking (1 input channel as in a depthwise conv, which would not be valid).

lgeiger:
> I really like this idea in theory, but you're right that it won't be possible once the filters have been bitpacked.

That's true. In theory I think it woud still work since the filters will be bitpacked in a later pass, but this means that the verifier would be tied to this transformation anyway. Let's keep it like this for now.

PR #552, opened 2020-10-23
Speedup end2end tests

lgeiger:
## What do these changes do?
This PR improves the speed of the end2end tests by caching the dataset between runs. This can be particularly noticable on slower internet connections since by default we will try and read the dataset directly from GCS.

## How Has This Been Tested?
CI

## Benchmark Results
Locally this reduces the test time from ~100s to ~30s


Tombana (approved):

AdamHillier (approved):
Awesome :tada:

PR #556, opened 2020-10-29
Cancel workflows of outdated commits

lgeiger:
This PR adds a GitHub action that will cancel outdated workflows. E.g. when pushing a sceond commit to a PR, the workflow for the first commit should be canceled to save CI time.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -59,6 +59,10 @@ jobs:
     runs-on: ubuntu-latest
     if: "!contains(github.event.head_commit.message, 'ci-skip')"
     steps:
+      - name: Cancel Previous Runs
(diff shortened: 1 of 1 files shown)


.github/workflows/unittests.yml
```diff
@@ -59,6 +59,10 @@ jobs:
     runs-on: ubuntu-latest
     if: "!contains(github.event.head_commit.message, 'ci-skip')"
     steps:
+      - name: Cancel Outdated Runs
```
lgeiger, line 62: I added this as a normal step to one of the CI jobs, however we could also run this on a dedicated job (however then it would showup in the PR checks summary).

Tombana (approved):

PR #560, opened 2020-11-05
Create a lightweight `BConv2DParams` struct in `/core`.

AdamHillier:
## What do these changes do?

As @Tombana suggested in #550, this PR removes the `TfliteBConv2DParams` struct that was quite heavy -- containing e.g. `std::vector`s -- and in its place introduces a lightweight, `BConv2DParams` struct in `/core/bconv2d/params.h`, which contains the important fields that are always needed at runtime on every platform. An `OpData` struct is added to `bconv2d.cc`, which wraps `BConv2DParams` but also contains the 'extra' fields such as the vectors.

## How Has This Been Tested?

CI.

## Benchmark Results

N/A.

## Related issue number

#550.


Tombana (approved):
:rocket:

lgeiger (approved):

PR #566, opened 2020-11-14
Set --linkopt=-dead_strip when building mac wheels

lgeiger:
## What do these changes do?

This PR strips dead symbols from the built wheel which together with #565 reduces the wheel size from 48MB to 44MB.

See 
https://developer.apple.com/library/archive/documentation/Performance/Conceptual/CodeFootprint/Articles/CompilerOptions.html

## How Has This Been Tested?
Locally built pip package and tested the converter..

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -129,7 +129,7 @@ jobs:
 
           ./configure.sh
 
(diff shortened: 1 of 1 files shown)


Tombana (approved):

Tombana:
Can we somehow test that this does not strip out kernels that are needed for constant-folding? (Based on the file-size I'd say its fine)

lgeiger:
> Can we somehow test that this does not strip out kernels that are needed for constant-folding? (Based on the file-size I'd say its fine)

Yes, I double checked with quicknet.

AdamHillier (approved):

PR #571, opened 2020-11-19
Add zero-padding support to the reference kernel

Tombana:
## What do these changes do?
This PR adds zero-padding support to the reference kernel, for all output types, group sizes, and activation functions.
(Previously we only supported zero-padding for the regular float kernel with no activation functions and no groups.)

Furthermore the `LiteInterpreter` now has an extra argument to choose the reference kernel, in order to be able to test against zero-padding.

## How Has This Been Tested?
The bconv2d tests have been updated.

### DIFF
--- a/larq_compute_engine/tflite/kernels/bconv2d.cc
+++ b/larq_compute_engine/tflite/kernels/bconv2d.cc
@@ -187,6 +187,20 @@ TfLiteStatus Prepare(TfLiteContext* context, TfLiteNode* node) {
     bconv2d_params->groups = groups;
   }
 
+  if (bconv2d_params->padding_type == kTfLitePaddingSame &&
+      bconv2d_params->pad_value == 0) {
+    TF_LITE_ENSURE_MSG(
+        context,
+        (kernel_type == KernelType::kReference &&
+         bconv2d_params->channels_in % 2 == 0) ||
+            (kernel_type != KernelType::kReference &&
+             output->type == kTfLiteFloat32 &&
+             op_data->fused_activation_function == kTfLiteActNone),
+        "Zero-padding is only supported by the reference kernel with an even "
+        "number of input channels, or when using "
+        "float output with no fused activation function.");
+  }
+
   // Compute the padding and output values (height, width)
   int out_width, out_height;
   bconv2d_params->padding_values = ComputePaddingHeightWidth(
@@ -202,12 +216,6 @@ TfLiteStatus Prepare(TfLiteContext* context, TfLiteNode* node) {
     TF_LITE_ENSURE_EQ(context, thresholds->type, kTfLiteInt32);
     TF_LITE_ENSURE_EQ(context, SizeOfDimension(thresholds, 0),
                       bconv2d_params->channels_out);
-    TF_LITE_ENSURE_MSG(context,
-                       kernel_type == KernelType::kReference ||
-                           bconv2d_params->padding_type != kTfLitePaddingSame ||
-                           bconv2d_params->pad_value == 1,
-                       "Writing bitpacked output is only supported by the "
-                       "reference kernel, or with valid or one-padding.");
   } else {
     TF_LITE_ENSURE_EQ(context, post_activation_multiplier->type,
                       kTfLiteFloat32);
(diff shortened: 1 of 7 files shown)


larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -124,14 +124,6 @@ void* Init(TfLiteContext* context, const char* buffer, std::size_t length) {
...
   op_data->fused_activation_function = ConvertActivation(
       (ActivationFunctionType)m["fused_activation_function"].AsInt32());
-  if (bconv2d_params->padding_type == kTfLitePaddingSame &&
-      bconv2d_params->pad_value != 1 &&
-      op_data->fused_activation_function != kTfLiteActNone) {
-    TF_LITE_KERNEL_LOG(
-        context,
-        "Fused activations are only supported with valid or one-padding.");
-    return op_data;
-  }
```
Tombana, line 134: This had to move from `Init` to `Prepare` because it is now dependent on `kernel_type`.

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -230,25 +223,31 @@ TfLiteStatus Prepare(TfLiteContext* context, TfLiteNode* node) {
...
-            bconv2d_params->pad_value == 1,
-        "8-bit quantization is only supported with valid or one-padding");
+
+    TF_LITE_ENSURE_MSG(context,
+                       (kernel_type == KernelType::kReference &&
+                        bconv2d_params->channels_in % 2 == 0) ||
+                           bconv2d_params->padding_type != kTfLitePaddingSame ||
+                           bconv2d_params->pad_value == 1,
+                       "8-bit quantization is only supported by the reference "
+                       "kernel, or with valid or one-padding");
```
AdamHillier, line 233: I'm not quite sure about this check, also because it's inside an `if int8` statement.

Maybe it should become: `if zero_padding { assert(float_output or (ref_kernel and even_c_in)) }`?

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -496,18 +496,25 @@ void runTest(const TestParam& param) {
...
-    // We could use `EXPECT_DEATH` here but it is extremely slow. Therefore we
-    // have a separate test below, and here we just skip.
-    GTEST_SKIP();
-    return;
+  const bool reference_supports_zero_padding =
+      is_reference_registration && (input_depth % 2 == 0);
+
+  if (padding == Padding_SAME && !reference_supports_zero_padding) {
+    if (groups > 1 || activation == ActivationFunctionType_RELU ||
+        write_bitpacked_output || int8_output) {
```
AdamHillier, line 504: I think `groups > 1` is no longer a constraint?

How do you feel about breaking it down slightly more, to something like:
```
if ZERO_PADDING {
    if (is_reference) {
        if (!even_c_i || has_relu) SKIP()
    } else {
        if (!float_output || has_relu) SKIP()
    }
}
```

I think that might make it a bit clearer what's going on.

larq_compute_engine/tflite/kernels/bconv2d.cc
Tombana: I think the current assert is technically correct, but I agree that your way is much easier to read. I've added it in the latest commit, and it fused three asserts into one easier-to-read assert! Could you double-check that the assert logic is still correct?

larq_compute_engine/tflite/tests/bconv2d_test.cc
```diff
@@ -492,28 +493,20 @@ void runTest(const TestParam& param) {
...
-      // - Writing bitpacked output
-      // - Int8 output
-      // We could use `EXPECT_DEATH` here but it is
-      // extremely slow. Therefore we have a separate test below, and here we
-      // just skip.
-      GTEST_SKIP();
-      return;
+  if (padding == Padding_SAME) {
+    if (is_reference_registration) {
+      if (input_depth % 2 != 0) GTEST_SKIP();
```
AdamHillier, line 501:
```suggestion
      if (input_depth % 2 != 0) {
          GTEST_SKIP();
          return;
      }
```

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -210,11 +216,6 @@ TfLiteStatus Prepare(TfLiteContext* context, TfLiteNode* node) {
     TF_LITE_ENSURE_EQ(context, thresholds->type, kTfLiteInt32);
     TF_LITE_ENSURE_EQ(context, SizeOfDimension(thresholds, 0),
                       bconv2d_params->channels_out);
-    TF_LITE_ENSURE_MSG(context,
-                       bconv2d_params->padding_type != kTfLitePaddingSame ||
-                           bconv2d_params->pad_value == 1,
-                       "Writing bitpacked output is only supported with "
-                       "valid or one-padding.");
```
AdamHillier, line 217: I think we still need this check

AdamHillier (approved):
Two minor remaining comments but otherwise this looks great :)

larq_compute_engine/tflite/kernels/bconv2d.cc
AdamHillier: Oh wait never mind, I'm wrong about that! Yep, we don't need it.

Tombana: Isn't this caught by the general zero-padding check? That one will fail when writing bitpacked output.

Tombana:
@lkindrat-xmos This might be relevant to you for internal testing. You can now run `LiteInterpreter( use_reference_bconv = True )` which supports zero-padding.

PR #576, opened 2020-12-09
Use GitHub actions workflow dispatch to trigger nightly releases

lgeiger:
This PR adds the possibility to manually trigger nightly releases using the GitHub actions `workflow_dispatch` event. It also allows us to set a version for the wheel when requesting the a build without the need to manually edit `setup.py` and push to a specific branch. This workflow can be manually triggered from the GitHub actions tab.


AdamHillier (approved):
Nice :+1:

PR #585, opened 2021-01-22
Add preliminary support for the xcore target.

AdamHillier:
## What do these changes do?

This PR adds support for a 'target' flag to the LCE converter, with two values: "arm" (the default), and "xcore" (the XMOS xcore.ai board). For the time being, the only difference between the two is that the "arm" target supports 'same zero' padding, and the binary maxpool op, but the "xcore" target does not.

## How Has This Been Tested?

I added a small test case to the Python converter test, but it's not possible to use the end2end tests with the xcore target because the converted models are not compatible with the LCE interpreter (and indeed are not meant to be). Need to work out what to do here.

## Benchmark Results

N/A.

## Related issue number

N/A.

### DIFF
--- a/larq_compute_engine/mlir/transforms/optimize.cc
+++ b/larq_compute_engine/mlir/transforms/optimize.cc
@@ -2,6 +2,7 @@
 
 #include "larq_compute_engine/core/bitpacking/bitpack.h"
 #include "larq_compute_engine/mlir/ir/lce_ops.h"
+#include "larq_compute_engine/mlir/transforms/passes.h"
 #include "llvm/ADT/Optional.h"
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/StringSwitch.h"
@@ -18,17 +19,27 @@ namespace {
 
 // Optimize LCE operations in functions.
 struct OptimizeLCE : public PassWrapper<OptimizeLCE, FunctionPass> {
- public:
-  // The default value must be true so that we can run with the optimisation in
-  // the file-check tests.
-  explicit OptimizeLCE() : experimental_enable_bitpacked_activations_(true) {}
-  explicit OptimizeLCE(bool experimental_enable_bitpacked_activations)
-      : experimental_enable_bitpacked_activations_(
-            experimental_enable_bitpacked_activations) {}
+  OptimizeLCE() = default;
+  OptimizeLCE(const OptimizeLCE& pass) {}
+  OptimizeLCE(const LCETarget target,
+              const bool experimental_enable_bitpacked_activations) {
+    target_.setValue(target);
+    experimental_enable_bitpacked_activations_.setValue(
+        experimental_enable_bitpacked_activations);
+  }
+
   void runOnFunction() override;
 
  private:
-  bool experimental_enable_bitpacked_activations_;
+  Option<LCETarget> target_{
+      *this, "target", llvm::cl::desc("Platform to target."),
+      llvm::cl::values(clEnumValN(LCETarget::ARM, "arm", "ARM target"),
+                       clEnumValN(LCETarget::XCORE, "xcore", "XCORE target"))};
+  Option<bool> experimental_enable_bitpacked_activations_{
+      *this, "experimental-enable-bitpacked-activations",
+      llvm::cl::desc("Include patterns to allow binary convolutions to output "
+                     "bitpacked activations."),
+      llvm::cl::init(true)};
 };
 
 bool IsConstantValue(Attribute values, float expected_value) {
@@ -40,7 +51,13 @@ bool IsConstantValue(Attribute values, float expected_value) {
   return true;
 }
 
-#include "larq_compute_engine/mlir/transforms/generated_optimize.inc"
+namespace target_arm {
+#include "larq_compute_engine/mlir/transforms/generated_optimize_target_arm.inc"
+}
+
+namespace target_other {
+#include "larq_compute_engine/mlir/transforms/generated_optimize_target_other.inc"
+}
 
 /**
  * =================================================
@@ -301,10 +318,15 @@ void OptimizeLCE::runOnFunction() {
   auto* ctx = &getContext();
   auto func = getFunction();
 
-  TFL::populateWithGenerated(ctx, patterns);
+  if (target_ == LCETarget::ARM) {
+    target_arm::populateWithGenerated(ctx, patterns);
+  } else {
+    target_other::populateWithGenerated(ctx, patterns);
+  }
   if (experimental_enable_bitpacked_activations_) {
     patterns.insert<SetBitpackedActivations>(ctx);
   }
+
   applyPatternsAndFoldGreedily(func, patterns);
 }
 
@@ -312,9 +334,10 @@ void OptimizeLCE::runOnFunction() {
 
 // Creates an instance of the TensorFlow dialect OptimizeLCE pass.
 std::unique_ptr<OperationPass<FuncOp>> CreateOptimizeLCEPass(
-    bool experimental_enable_bitpacked_activations) {
+    const LCETarget target,
+    const bool experimental_enable_bitpacked_activations) {
   return std::make_unique<OptimizeLCE>(
-      experimental_enable_bitpacked_activations);
+      target, experimental_enable_bitpacked_activations);
 }
 
 static PassRegistration<OptimizeLCE> pass(
--- a/larq_compute_engine/mlir/transforms/optimize_patterns_target_arm.td
+++ b/larq_compute_engine/mlir/transforms/optimize_patterns_target_arm.td
@@ -1,0 +1,20 @@
+include "larq_compute_engine/mlir/transforms/optimize_patterns_common.td"
+
+// Insert a binary maxpool if a maxpool is followed by a sign op.
+def : Pat<(LQ_QuantizeOp
+              (TFL_MaxPool2DOp:$pool_output
+                  $input,
+                  $padding,
+                  $stride_w,
+                  $stride_h,
+                  $filter_width,
+                  $filter_height,
+                  $fused_activation_function)),
+          (LQ_BMaxPool2dOp
+              (LQ_QuantizeOp $input),
+              $padding,
+              $stride_w,
+              $stride_h,
+              $filter_width,
+              $filter_height),
+          [(HasOneUse $pool_output)]>;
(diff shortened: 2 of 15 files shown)


AdamHillier:
Ideally we'd be able to have separate MLIR tests for the different targets, but I'm not sure our setup currently supports doing that?

larq_compute_engine/mlir/tests/optimize_target_arm.mlir
```diff
@@ -1,4 +1,34 @@
-// RUN: lce-tf-opt %s -tfl-optimize-lce -verify-diagnostics | FileCheck %s
+// RUN: lce-tf-opt %s -tfl-optimize-lce-target-arm -verify-diagnostics | FileCheck %s
+
+///////////////////////////
+// Target-specific tests //
+///////////////////////////
+
+// CHECK-LABEL: @reorder_maxpool_2d_quantize
```
AdamHillier, line 7: This is quite ugly. Basically, there are two sets of `optimize` pass tests, one for the Arm pass and one for the xcore pass. Most of the desired behaviour is the same, so most of the test cases are the same. Unfortunately, there isn't a nice way to define the 'common' test cases in one place since there's no 'include' statement in these MLIR filecheck files. Hence, my current solution is to have these two separate files with almost identical content - I have at least highlighted where they differ with the headings "Target-specific tests" and "Common tests".

Any other ideas are very welcome!

larq_compute_engine/mlir/transforms/optimize.cc
```diff
@@ -40,7 +39,13 @@ bool IsConstantValue(Attribute values, float expected_value) {
...
 }
 
-#include "larq_compute_engine/mlir/transforms/generated_optimize.inc"
+namespace target_arm {
+#include "larq_compute_engine/mlir/transforms/generated_optimize_target_arm.inc"
+}
+
+namespace target_other {
+#include "larq_compute_engine/mlir/transforms/generated_optimize_target_other.inc"
+}
```
AdamHillier, line 48: These have to be included into distinct namespaces otherwise there are name conflicts.

larq_compute_engine/mlir/transforms/optimize.cc
```diff
@@ -312,13 +329,24 @@ void OptimizeLCE::runOnFunction() {
...
 
-static PassRegistration<OptimizeLCE> pass(
-    "tfl-optimize-lce", "Optimize within the TensorFlow Lite dialect");
+static PassRegistration<OptimizeLCE<LCETarget::ARM>> pass_target_arm(
+    "tfl-optimize-lce-target-arm",
+    "Optimize within the TensorFlow Lite dialect, for ARM targets");
+
+static PassRegistration<OptimizeLCE<LCETarget::XCORE>> pass_target_xcore(
+    "tfl-optimize-lce-target-xcore",
+    "Optimize within the TensorFlow Lite dialect, for XCORE targets");
```
AdamHillier, line 349: This is how two target-specific passes are registered for the purposes of the file-check tests.

The reason that `target` has to be a template argument of the `OptimizeLCE` struct is because when registering MLIR passes, the key that is used is derived from the type of the pass. This means it's not possible to register more than one pass of the same type! I discoved this after lots of trial-and-error. The relevant MLIR function is [here](https://github.com/llvm/llvm-project/blob/62ec4ac90738a5f2d209ed28c822223e58aaaeb7/mlir/lib/Pass/PassRegistry.cpp#L105-L111) - note that if the TODO there had been completed this wouldn't be an issue.

larq_compute_engine/mlir/transforms/optimize_patterns_target_arm.td
```diff
@@ -0,0 +1,20 @@
...
+                  $filter_height,
+                  $fused_activation_function)),
+          (LQ_BMaxPool2dOp
+              (LQ_QuantizeOp $input),
+              $padding,
+              $stride_w,
+              $stride_h,
+              $filter_width,
+              $filter_height),
+          [(HasOneUse $pool_output)]>;
```
AdamHillier, line 20: The format here is that there is one `optimize_patterns_common.td` file that defines the common re-write patterns, and is imported and extended as needed by target-specific table-gen files, as is being done here.

larq_compute_engine/mlir/tests/optimize_target_arm.mlir
Tombana: I think the current way is fine, the comments make it clear which part is duplicate.
Another option would be is to write something like ` // LCE_INCLUDE: filename` and then let the `run_lit.sh` script inject the file at that position. This should be easy to do from bash with `sed` or maybe even just `cat` to concatenate the files. (We could even replace it by a python script if that's easier, although I think it mainly calls subprocesses which arguably is what bash is made for.)

larq_compute_engine/mlir/transforms/prepare_patterns_target_arm.td
```diff
@@ -0,0 +1,4 @@
+include "larq_compute_engine/mlir/transforms/prepare_patterns_common.td"
+
+// On ARM we support 'same-zero' padding.
+def : PrepareBConvPadValue0Pat<"SAME">;
```
Tombana, line 4: Very clean! :rocket:

Tombana (approved):

larq_compute_engine/mlir/tests/optimize_target_arm.mlir
AdamHillier: Ah yeah, nice, well if @lgeiger thinks that would be nicer we can do that in this PR, otherwise that's an improvement we can do in a later PR.

AdamHillier: With help from @lgeiger I was able to get this working in a much nicer way, so resolving this now.

larq_compute_engine/mlir/transforms/optimize.cc
lgeiger: I don't think I fully understand why we need two dedicated passes. Why can't we handle this through a pass option similar to [this](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/mlir/lite/transforms/lower_static_tensor_list.cc#L94-L99) and [this](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/mlir/lite/tests/lower-static-tensor-list.mlir#L1) or [this](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/mlir/lite/quantization/import_quant_stats_pass.cc#L42-L45) and [this](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/mlir/lite/quantization/tests/import_quant_stats.mlir#L1)?

larq_compute_engine/mlir/transforms/optimize.cc
lgeiger: This approach sounds good to me. Alternatively we could not include `optimize_common.td` in `optimize_target_arm.td` and then always populate the patterns from `optimize_common.td` here and only conditionally add the target specific ones.

larq_compute_engine/mlir/transforms/prepare_tf.cc
```diff
@@ -155,18 +164,42 @@ void PrepareLCE::runOnFunction() {
...
+    return std::make_unique<PrepareLCE<LCETarget::XCORE>>();
+  }
 }
 
-static PassRegistration<PrepareLCE> pass("tfl-prepare-lce", "Inject LCE Ops");
+static PassRegistration<PrepareLCE<LCETarget::ARM>> pass_target_arm(
+    "tfl-prepare-lce-target-arm", "Inject LCE Ops, for ARM targets");
+
+static PassRegistration<PrepareLCE<LCETarget::XCORE>> pass_target_xcore(
+    "tfl-prepare-lce-target-xcore", "Inject LCE Ops, for XCORE targets");
```
lgeiger, line 203: Same comment as above, why do we need multiple passes instead of options to a single pass

larq_compute_engine/mlir/transforms/optimize.cc
AdamHillier: There is a very answer to this, and it's that I wasn't aware this was an option :)

Thanks for finding this, I guess we should have been using this already for the `experimental_enable_bitpacked_activations` option of the optimise pass.

lgeiger: Sounds good :)

> I guess we should have been using this already for the experimental_enable_bitpacked_activations option of the optimise pass.

Yes, although I don't think there it really mattered since we never really tested it with the flag disabled.

larq_compute_engine/mlir/transforms/optimize.cc
AdamHillier: The nice thing about the `include` statement is that it means we don't need to redefine the 'helper' functions like these at the top of the target-specific files:

https://github.com/larq/compute-engine/blob/95fd0b09a55647a021bed352fdacf4257c31767b/larq_compute_engine/mlir/transforms/optimize_patterns.td#L6-L10

larq_compute_engine/mlir/transforms/optimize.cc
AdamHillier: I've just pushed an updated commit using these flags.

larq_compute_engine/mlir/transforms/prepare_tf.cc
```diff
@@ -17,10 +18,21 @@ using compute_engine::core::bitpacking_bitwidth;
...
+
   void runOnFunction() override;
   void getDependentDialects(DialectRegistry& registry) const override {
     registry.insert<mlir::lq::LarqDialect>();
   }
+
+ private:
+  Option<bool> target_arm_{
+      *this, "target-arm",
+      llvm::cl::desc("Include patterns to target the ARM platform.")};
```
lgeiger, line 35: Great 👍 

How do you feel about making this an enum?
```cpp
  Option<LCETarget> target_{
      *this, "target",
      llvm::cl::desc("Platform to target."),
      llvm::cl::values(
          clEnumValN(LCETarget::ARM, "arm", "ARM target"),
          clEnumValN(LCETarget::ARM, "xmos", "XMOS target"))};
```
which can then be configured using `-tfl-prepare-lce=target=arm`.

But since we only have two targets, I am also happy to keep it as a boolean for now.

larq_compute_engine/mlir/transforms/optimize.cc
lgeiger: 👍 Makes sense.

lgeiger (commented):
This looks great!

I only have one small suggestion, but I am also happy to merge this PR as is.

lgeiger (approved):

PR #586, opened 2021-01-22
Bump package versions on CI.

AdamHillier:
## How Has This Been Tested?

CI.

## Benchmark Results

N/A.

## Related issue number

N/A.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -79,7 +79,7 @@ jobs:
         if: github.ref != 'refs/heads/master'
         shell: bash
       - name: Install pip dependencies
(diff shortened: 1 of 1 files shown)


Tombana:
Should we also update the submodule and bazel dependency to `2.4.1` ?

AdamHillier:
> Should we also update the submodule and bazel dependency to `2.4.1` ?

Ah nice, good idea, will do :+1:

AdamHillier:
> > Should we also update the submodule and bazel dependency to `2.4.1` ?
> 
> Ah nice, good idea, will do 👍

I opened #587 to do this.

lgeiger (approved):

PR #589, opened 2021-01-25
Use a Python script for Bazel configuration, like TensorFlow does.

AdamHillier:
## What do these changes do?

This PR replaces the `configure.sh` Bash script with a `configure.py` Python script, to match what TensorFlow have. This is useful for building LCE on a platform that doesn't have a bash shell - i.e. most Windows machines, where you haven't gone out of your way to install one. Python is required for building LCE, so requiring Python for generating the Bazel config should not be a problem.

The included `configure.py` is a significantly cut down version of the `configure.py` in the TensorFlow repo.

## How Has This Been Tested?

CI, and manual testing on a Windows machine using PowerShell.

## Benchmark Results

N/A.

## Related issue number

N/A.


Tombana (approved):
The bash script only writes a `manylinux2010` line to bazelrc. This python script seems to also write things like a python path and some compiler options. I assume this means it has to be run within the virtual env of a python version that you're building TF for? (I always have to run bazel with a python3.8 venv otherwise it'll try to build a 3.9 version). I think we should add a print statement stating something like
```
Bazel is now configured for python xx builds. If this is not intended, re-run the script from within a virtual environment with a different python version.
```
This way, we avoid someone accidentally running the script with the wrong python version and then wasting a 4 hour converter build.

configure.py
```diff
@@ -0,0 +1,408 @@
...
+    paths = []
+    for path in all_paths:
+        if os.path.isdir(path):
+            paths.append(path)
+    return paths
+
+
+def get_python_major_version(python_bin_path):
+    """Get the python major version."""
+    return run_shell([python_bin_path, "-c", "import sys; print(sys.version[0])"])
```
Tombana, line 229: The TF people probably have a reason for this, but since the script is already python, why could they not just `return sys.version[0]` here?

AdamHillier:
> The bash script only writes a `manylinux2010` line to bazelrc. This python script seems to also write things like a python path and some compiler options. I assume this means it has to be run within the virtual env of a python version that you're building TF for? (I always have to run bazel with a python3.8 venv otherwise it'll try to build a 3.9 version).

I guess technically `configure.py` doesn't have to be run in the environment you want to use since you could put in a path to whichever Python binary you like, but yeah it'd be by far easiest if you run it in the environment you want to use because that will fill in the default value correctly.

I think the biggest advantage of writing the Python paths to the bazelrc file is that after configuration you don't have to worry about Python environments again as all Bazel builds will use the path that's hard-coded in the config file, no matter whats in your PATH. For me at least, at the moment I always have to run `bazel build/test` inside the right Python environment otherwise it gets confused, which won't be necessary any more.

> I think we should add a print statement stating something like
>
> ```
> Bazel is now configured for python xx builds. If this is not intended, re-run the script from within a virtual environment with a different python version.
> ```
> 
> This way, we avoid someone accidentally running the script with the wrong python version and then wasting a 4 hour converter build.

Ah I hadn't thought of that, but yes I'll add something along those lines :+1:

configure.py
AdamHillier: I think it's because the version of Python that you're using to run this script is not necessarily the same version of Python that you're configuring for the build. You could have overriden the bin/lib paths to point to whatever random version of Python you like.

configure.py
```diff
@@ -0,0 +1,413 @@
...
+        "Eigen strong inline overridden.",
+        "Not overriding eigen strong inline, "
+        "some compilations could take more than 20 mins.",
+    ):
+        # Due to a known MSVC compiler issue
+        # https://github.com/tensorflow/tensorflow/issues/10521
+        # Overriding eigen strong inline speeds up the compiling of
+        # conv_grad_ops_3d.cc and conv_ops_3d.cc by 20 minutes,
+        # but this also hurts the performance. Let users decide what they want.
+        write_to_bazelrc("build --define=override_eigen_strong_inline=true")
```
lgeiger, line 386: Are there any cases in which we would like to disable this? If not, how do you feel about adding this directly to the globabl `.bazelrc` instead?

configure.py
```diff
@@ -0,0 +1,413 @@
...
+
+
+def main():
+    # Make a copy of os.environ to be clear when functions and getting and setting
+    # environment variables.
+    environ_cp = dict(os.environ)
+
+    reset_lce_configure_bazelrc()
+
+    setup_python(environ_cp)
```
lgeiger, line 401: Thanks for the detailed comments on why this is a good idea. I agree that hardcoding it makes sense especially when using multiple environments to build LCE.

.github/workflows/release.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           path: /tmp/lce_android
           key: ${{ runner.os }}-${{ hashFiles('**/third_party/install_android.sh') }}
       - name: Configure Bazel
-        run: ./configure.sh <<< $'n\n'
+        run: python configure.py <<< ""
```
lgeiger, line 32: Why do we need to explicitly pass "" when calling this script?

lgeiger (changes requested):
Thanks for moving this script to Python, I think that will make it easier to adapt it if we need to in the future.

I made a few suggestions below.

.github/workflows/release.yml
AdamHillier: It's interactive, so if you don't pass input it will prompt for input. It's possible that when you run it non-locally (i.e. not attached to a terminal) then it would work without this, I haven't actually tested it.

configure.py
```diff
@@ -0,0 +1,413 @@
...
+
+def set_windows_build_flags(environ_cp):
+    """Set Windows specific build options."""
+
+    # First available in VS 16.4. Speeds up Windows compile times by a lot. See
+    # https://groups.google.com/a/tensorflow.org/d/topic/build/SsW98Eo7l3o/discussion
+    # pylint: disable=line-too-long
+    write_to_bazelrc(
+        "build --copt=/d2ReducedOptimizeHugeFunctions --host_copt=/d2ReducedOptimizeHugeFunctions"
+    )
```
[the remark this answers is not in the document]
AdamHillier: Ah yes, I developed this PR in parallel with #590 which removes it from the `.bazelrc` (so moving it from there to the config file), which was simply because that's where TensorFlow has it. I think they probably did it this way because these flags are not supported prior to Visual Studio 2019 v16.4, and it's nicer to override the user bazelrc than the repo one? Idk though. Do we care about supporting Windows builds with older versions of Windows/older compilers?

configure.py
AdamHillier: I guess it's here because it's a compiler problem workaround, and (apparently?) hurts performance. I think the user bazelrc file makes sense for a flag like that, unless you feel otherwise?

configure.py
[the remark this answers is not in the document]
lgeiger: > Do we care about supporting Windows builds with older versions of Windows/older compilers?

I personally don't, especially considering the crazy build times with older compiler versions:
https://groups.google.com/u/1/a/tensorflow.org/g/build/c/SsW98Eo7l3o/discussion

TensorFlow enabled it by default in https://github.com/tensorflow/tensorflow/commit/9983bb3dee46da301b8294000984eb67fd2b995a (it was done conditionally before) so I think it would be save to move it to the global config. But feel free to keep it like this.

configure.py
lgeiger: Makes sense :+1:

.github/workflows/release.yml
lgeiger: I am pretty sure that GitHub actions will handle this correctly, so we should be able to safely change this to
```shell
./configure.py
```

configure.py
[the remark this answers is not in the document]
AdamHillier: Ah nice, yeah makes sense, in that case let's make it default, thanks :)

lgeiger (approved):

configure.py
Tombana: Oh that makes sense, if you've set `PYTHON_BIN_PATH` from your shell. :+1:

PR #590, opened 2021-01-25
Fix Windows release builds.

AdamHillier:
## What do these changes do?

This PR (hopefully) fixes the Windows release builds, by copying in the most recent Windows build config from TensorFlow, using some extra commands that reduce memory usage, and increasing the size of the pagefile (swapfile).

## How Has This Been Tested?

I will trigger a nightly release from this branch.

## Benchmark Results

N/A.

## Related issue number

Relies on #589.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -57,6 +57,15 @@ build:macos --config=noaws --config=nogcp --config=nohdfs --config=nonccl
 build:windows --copt=/D_USE_MATH_DEFINES
 build:windows --host_copt=/D_USE_MATH_DEFINES
 
+# Config to use a mostly-static build and disable modular op registration
+# support (this will revert to loading TensorFlow with RTLD_GLOBAL in Python).
+# By default, TensorFlow will build with a dependence on
+# //tensorflow:libtensorflow_framework.so.
+build:monolithic --define framework_shared_object=false
+
+# On windows, we still link everything into a single DLL.
+build:windows --config=monolithic
+
 # Make sure to include as little of windows.h as possible
 build:windows --copt=-DWIN32_LEAN_AND_MEAN
 build:windows --host_copt=-DWIN32_LEAN_AND_MEAN
@@ -75,15 +84,15 @@ build:windows --linkopt=/OPT:REF
 build:windows --host_linkopt=/OPT:REF
 build:windows --linkopt=/OPT:ICF
 build:windows --host_linkopt=/OPT:ICF
-
 build:windows --copt=/d2ReducedOptimizeHugeFunctions
 build:windows --host_copt=/d2ReducedOptimizeHugeFunctions
+build:windows --experimental_strict_action_env=true
 
 # Verbose failure logs when something goes wrong
 build:windows --verbose_failures
 
 # Suppress C++ compiler warnings, otherwise build logs become 10s of MBs.
-build:windows --copt=/w
+build:windows --copt=/W0
 
 # On windows, we never cross compile
 build:windows --distinct_host_configuration=false
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -161,7 +161,7 @@ jobs:
           docker run -e PYTHON_VERSION=${{ matrix.python-version }} \
             -e LCE_RELEASE_VERSION=${{ github.event.inputs.version }} \
             -v ${PWD}:/compute-engine -w /compute-engine \
-            tensorflow/tensorflow:custom-op-ubuntu16 \
+            tensorflow/tensorflow:2.2.0-custom-op-ubuntu16 \
             .github/tools/release_linux.sh
 
           sudo apt-get -y -qq install patchelf --no-install-recommends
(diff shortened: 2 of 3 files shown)


AdamHillier:
Test release run: https://github.com/larq/compute-engine/runs/1761598156

.bazelrc
```diff
@@ -57,6 +57,15 @@ build:macos --config=noaws --config=nogcp --config=nohdfs --config=nonccl
...
 
+# Config to use a mostly-static build and disable modular op registration
+# support (this will revert to loading TensorFlow with RTLD_GLOBAL in Python).
+# By default, TensorFlow will build with a dependence on
+# //tensorflow:libtensorflow_framework.so.
+build:monolithic --define framework_shared_object=false
+
+# On windows, we still link everything into a single DLL.
+build:windows --config=monolithic
+
```
lgeiger, line 68: I thought this defaults to false anyway, but let's keep it like this to be save.

AdamHillier:
> Test release run: https://github.com/larq/compute-engine/runs/1761598156

The Windows builds succeeded here, but the Linux ones failed. @lgeiger do you think this is related to this PR?

lgeiger:
> The Windows builds succeeded here, but the Linux ones failed. @lgeiger do you think this is related to this PR?

Great news on the Windows builds!

The Linux failures could either be related to #589 since we are now hardcoding the Python env variables or it could be due to a change in the Dockerfile.
It looks like `tensorflow/tensorflow:custom-op-ubuntu16` was updated recently, it might be worth to pin it to `tensorflow/tensorflow:2.2.0-custom-op-ubuntu16` instead and try again.
Our release setup is a bit hacky and needs to work around a few issues with the docker file, but I think moving to the `tensorflow/tensorflow:2.4.0-custom-op-ubuntu16` might require more work, so I'd rather postpone this until we really need it (e.g. for Python 3.9 support).

.bazelrc
```diff
@@ -75,15 +84,13 @@ build:windows --linkopt=/OPT:REF
 build:windows --host_linkopt=/OPT:REF
 build:windows --linkopt=/OPT:ICF
 build:windows --host_linkopt=/OPT:ICF
-
-build:windows --copt=/d2ReducedOptimizeHugeFunctions
-build:windows --host_copt=/d2ReducedOptimizeHugeFunctions
```
lgeiger, line 80: Keep after rebase (see https://github.com/larq/compute-engine/pull/589/commits/f2f6c73f8ba2fccd9dacd4ee7098e6bc020ee792)

.github/workflows/release.yml
```diff
@@ -184,27 +184,33 @@ jobs:
...
         python-version: [3.6, 3.7, 3.8]
     steps:
+      - name: Configure Pagefile
+        uses: al-cheb/configure-pagefile-action@v1.2
+        with:
+          minimum-size: 4GB
+      - name: Pagefile size
+        run: |
+          (Get-CimInstance Win32_PageFileUsage).AllocatedBaseSize
+        shell: pwsh
```
lgeiger, line 194: Awesome find 👍

.github/workflows/release.yml
```diff
@@ -184,27 +184,33 @@ jobs:
...
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
       - name: Build Windows wheels
         run: |
-          export LCE_RELEASE_VERSION=${{ github.event.inputs.version }}
```
lgeiger, line 193: I think we still need this to set the correct version.

.github/workflows/release.yml
```diff
@@ -184,27 +184,33 @@ jobs:
...
-          export PYTHON_BIN_PATH=$(which python)
-          export BAZEL_VC="C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/"
-
-          mkdir C:/bazel_cache
+          $Env:PYTHON_BIN_PATH = Get-Command python | Select-Object -ExpandProperty Definition
```
lgeiger, line 201: Not sure if we still need this since we now set it already in #589

.github/workflows/release.yml
```diff
@@ -184,27 +184,33 @@ jobs:
...
+          $Env:BAZEL_VC = "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC"
+          set PreferredToolArchitecture=x64
```
lgeiger, line 203: Nice 👍

lgeiger (changes requested):
Thank you so much for debugging this 🙏

.github/workflows/release.yml
AdamHillier: Ah oops, didn't mean to remove that, thanks :)

.github/workflows/release.yml
AdamHillier: Ah yeah you're right, they must resolve to the same thing :+1:

lgeiger (approved):

AdamHillier:
I ran a build with the updated code last night and everything is now green, including the Linux builds, so I think this is good to go: https://github.com/larq/compute-engine/actions/runs/510137328

lgeiger (approved):
Great! Thanks for fixing the build 🚀

PR #595, opened 2021-01-29
Fix release version number for non-manual releases.

AdamHillier:
## What do these changes do?

The v0.5.0 release action silently failed to generate valid wheel files: specifically, for macOS and Windows the version number was set as v0.0.0 - this can be seen in the logs: https://github.com/larq/compute-engine/actions/runs/520610384.

I manually cancelled the run just before the PyPI upload step - had I not, I think we would have ended up accidentally publishing v0.0.0 wheels for Mac and Windows.

For some reason (and I didn't realise this was possible) this `os.getenv` call was returning the empty string: https://github.com/larq/compute-engine/blob/aa7ea1b1d4051f4a8b10702f7273d18552df0bbe/setup.py#L24

This PR makes a small change to the way the version number is calculated, and adds a sanity check that throws if the empty string is returned.

## How Has This Been Tested?

Manually generating the package.

## Benchmark Results

N/A.

## Related issue number

N/A.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -129,8 +129,6 @@ jobs:
           python --version
           python -m pip install delocate wheel setuptools numpy six --no-cache-dir
 
-          export LCE_RELEASE_VERSION=${{ github.event.inputs.version }}
-
           ./configure.py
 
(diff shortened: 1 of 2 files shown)


AdamHillier:
Once this issue is resolved I'll retrigger the v0.5.0 release.

lgeiger (approved):
Thanks for catching this. No idea why this was failing. Great that it is fixed now.

PR #596, opened 2021-01-30
Update release.yml to actually push Windows wheels.

AdamHillier:
🤦 https://github.com/larq/compute-engine/actions/runs/522141342 successfully ran and build Linux/Mac/Windows wheels, but only published the Linux/Mac ones. This should fix it. For the v0.5.0 release I will manually upload the built Windows wheels.


lgeiger (approved):
Nice catch!

PR #597, opened 2021-02-01
Enable bitpacked activation optimisations by default.

AdamHillier:
## What do these changes do?

This PR moves the bitpacked-activations converter rewrite pattern from C++ to TableGen, and enables it by default.

## How Has This Been Tested?

CI.

## Benchmark Results

N/A.

## Related issue number

N/A.


AdamHillier:
For the question of enabling it by default - this will work fine on Aarch64 since we support bitpacked output in the optimised kernels. On Aarch32 however I don't think it's currently supported. Do we think that is an issue @lgeiger @Tombana?

lgeiger:
> For the question of enabling it by default - this will work fine on Aarch64 since we support bitpacked output in the optimised kernels. On Aarch32 however I don't think it's currently supported. Do we think that is an issue?

For Aarch32 would this correctly fallback to the reference implementation? If this is the case I actually would be slightly in favour of enabling bitpacked activations by default and open an issue to add support for bitpacked activations to the Aarch32 optimized kernels. @AdamHillier @Tombana what do you think?

AdamHillier:
> For Aarch32 would this correctly fallback to the reference implementation? If this is the case I actually would be slightly in favour of enabling bitpacked activations by default and open an issue to add support for bitpacked activations to the Aarch32 optimized kernels. @AdamHillier @Tombana what do you think?

I'm fairly sure it will fall back to reference kernels but will check. If so, I agree that we should do that. I also think that adding Aarch32 support for optimised bitpacked output shouldn't be too hard.

CNugteren:
I have merged in the latest changes and updated this branch accordingly. @lgeiger perhaps you could take another look at the code and make a final decision on whether we can merge this in. Thanks!

Tombana (approved):

lgeiger (approved):
This looks great, from my side! Thanks for updating the PR @CNugteren

PR #607, opened 2021-02-22
Raise an error if mixed-precision models are being converted

CNugteren:
Mixed precision models are not supported by the TFLite conversion step yet (https://github.com/tensorflow/tensorflow/issues/46380), this could in some cases lead to segfaults. This guard prevents that and raises a proper error message before trying to convert such a model.

### DIFF
--- a/larq_compute_engine/mlir/python/converter.py
+++ b/larq_compute_engine/mlir/python/converter.py
@@ -113,6 +113,12 @@ def convert_keras_model(
             "Using `experimental_default_int8_range` as fallback quantization stats. "
             "This should only be used for latency tests."
         )
+    if hasattr(model, "dtype_policy") and model.dtype_policy.name == "mixed_float16":
+        raise RuntimeError(
+            "Mixed precision float16 models are not supported by the TFLite converted, "
+            "please convert them to float32 first. See also: "
+            "https://github.com/tensorflow/tensorflow/issues/46380"
+        )
     func = concrete_function_from_keras_model(model)
     if version.parse(tf.__version__) >= version.parse("1.15"):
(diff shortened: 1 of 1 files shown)


larq_compute_engine/mlir/python/converter.py
```diff
@@ -113,6 +113,12 @@ def convert_keras_model(
             "Using `experimental_default_int8_range` as fallback quantization stats. "
             "This should only be used for latency tests."
         )
+    if hasattr(model, "dtype_policy") and model.dtype_policy.name == "mixed_float16":
+        raise RuntimeError(
+            "Mixed precision float16 models are not supported by the TFLite converted, "
```
lgeiger, line 118:
```suggestion
            "Mixed precision float16 models are not supported by the TFLite converter, "
```

lgeiger (approved):
:+1:

larq_compute_engine/mlir/python/converter.py
```diff
@@ -113,6 +113,12 @@ def convert_keras_model(
             "Using `experimental_default_int8_range` as fallback quantization stats. "
             "This should only be used for latency tests."
         )
+    if hasattr(model, "dtype_policy") and model.dtype_policy.name == "mixed_float16":
```
lgeiger, line 116: How do you feel about changing this to:
```suggestion
    if hasattr(model, "dtype_policy") and model.dtype_policy.name != "float32":
```
Since `dtype_policy` can also be `bfloat16` in rare cases which I guess is also not supported.

lgeiger (approved):

AdamHillier (approved):

Issue #599, opened 2021-02-02
QuantConv2D binarized activations with bitpacked output as tf.int32


simonmaurer:
@lgeiger @AdamHillier given our latest mail conversation I'm trying to state one of the questions here so others can profit too.
this is somehow related to the #445 

given the hints from @lgeiger:
> If you use experimental_enable_bitpacked_activations=True during model conversion and have one-padding enabled in Larq, the converter will automatically fuse a convolution followed by a sign function into a single op where the binary convolution returns bitpacked binary feature maps directly.

I modelled a multitask network with binary layers and a sign operation (larq.math.sign respectively larq.quantizers.SteSign()) afterwards:
`output = larq.layers.QuantConv2D(256, kernel_size = (3,3), strides=(1,1), padding="same", pad_values=1, input_quantizer="ste_sign", kernel_quantizer="ste_sign",
                           kernel_constraint="weight_clip", use_bias=False)(output)
output = larq.quantizers.SteSign()(output)`

However there is a LceDequantize op and the output is a tensor with tf.float32 (identity).

<img width="891" alt="bitpacked_output" src="https://user-images.githubusercontent.com/5655311/106635521-3b006400-6581-11eb-971c-dfd0c94b41a9.png">

Can you give me some hint how I could strip that last dequantization layer so I get the actual tf.int32 tensor ?

I get it that you marked this as experimental but from a research perspective accessing the output layers (bitpacked output) directly is interesting (especially since you're thinking about making this default behavior as discussed here: [https://github.com/larq/compute-engine/pull/597](https://github.com/larq/compute-engine/pull/597)

I suppose the answer must lie in `modify_integer_quantized_model_io_type` as part of [https://github.com/larq/compute-engine/blob/master/larq_compute_engine/mlir/python/util.py](https://github.com/larq/compute-engine/blob/master/larq_compute_engine/mlir/python/util.py)

Tombana:
Hi, thanks for your interest in LCE!
The converter will indeed leave that `LceDequantize` in the graph, as this bitpacked datatype is kind of an LCE internal type that is only ever read by other LCE ops, so it won't be exposed to the outside.

I think that your best bet is to write a separate python script that modifies the generated flatbuffer, inspired by that function `modify_integer_quantized_model_io_type` that you linked to.

You could take that `util.py` file as example and then slightly modify it as follows.

- The following line will have to be changed to `opcode.customCode == "LceDequantize"`
[...]

PR #611, opened 2021-03-04
QuantConv2D binarized activations with tf.int32 bitpacked output

simonmaurer:
- `strip_lcedequantize_ops`: strips the output LceDequantize operators of a model such that the output is a bitpacked tf.int32 tensor

## What do these changes do?
Usually the lce_converter dequantizes the bitpacked output back to `tf.float32/tf.int8` resulting in an identity tensor. This is intended for training. However for inference or postprocessing one could use the bitpacked `tf.int32` tensors directly.
By using `strip_lce_dequantize_ops` one can strip the output **LceDequantize** operators of a model to get access to the bitpacked tf.int32 tensors.
Use cases: larq.layers.QuantConv2D followed by a sign operation (ie. larq.math.sign or larq.quantizers.SteSign())

## Update
Works for tf.float32 as well as for **tf.int8** quantized models.
Strips **Dequantize** ops too if any, ie. due to dequantization when using inference_output_type=_tf.int8_

## How Has This Been Tested?
```python
import tensorflow as tf
import larq as lq
from larq_compute_engine import convert_keras_model
from larq_compute_engine.mlir.python.util import strip_lcedequantize_ops

def toy_model_sign():
    img = tf.keras.layers.Input(shape=(224, 224, 3))
    x = lq.layers.QuantConv2D(256, kernel_size = 3, strides = 1, padding = "same", pad_values = 1, input_quantizer = "ste_sign", kernel_quantizer = "ste_sign",
                           kernel_constraint = "weight_clip")(img)
    x = lq.quantizers.SteSign()(x)
    return tf.keras.Model(inputs = img, outputs = x)

def quant(x):
    x = tf.quantization.fake_quant_with_min_max_vars(x, -3, 3)

def toy_model_int8_sign():
    img = tf.keras.layers.Input(shape=(224, 224, 3))
    x = quant(img)
    x = lq.layers.QuantConv2D(256, kernel_size = 3, strides = 1, padding = "same", pad_values = 1, input_quantizer = "ste_sign", kernel_quantizer = "ste_sign",
                           kernel_constraint = "weight_clip")(x)
    x = lq.quantizers.SteSign()(x)
    x = quant(x)
    return tf.keras.Model(inputs = img, outputs = x)

model = toy_model_sign()    # or toy_model_int8_sign()
tflite_model = convert_keras_model(model, inference_input_type = tf.float32, inference_output_type = tf.float32, experimental_enable_bitpacked_activations = True)
tflite_model = strip_lcedequantize_ops(tflite_model)
```

## Related issue number
#599


lgeiger:
@simonmaurer Thanks for the PR 🎉 

Could you also add a small unittest for this functionality?
Probably something similar to [this test](https://github.com/larq/compute-engine/blob/9297ff48c7bec562dc25aaad4853e65a351b710c/larq_compute_engine/tests/end2end_test.py#L203-L219) and a test that checks that the errors are raise correctly should do.

simonmaurer:
so the conversion works as seen on the following picturs:
Toy model (bitpacked)             |  Toy model (bitpacked _tf.int32_)
:-------------------------:|:-------------------------:
  |  


I realized that the error check conditions are never met if the model has TF quantized inputs/outputs (`tf.int8`) because these are different ops anyway (TF handles it to do fake quantization using `Quantize`, so there will never be a `LceDequantize` op as seen below):

toymodel_bitpacked_int8

simonmaurer:
sidenote: now I also got what these [lines](https://github.com/larq/compute-engine/blob/master/larq_compute_engine/mlir/python/util.py#L184-L198) together with [_find_int8_quantized_inputs_outputs](https://github.com/larq/compute-engine/blob/master/larq_compute_engine/mlir/python/util.py#L98) do:
they only remove input `Quantize` and output `DeQuantize` ops if in the model there were manually placed `tf.quantization.fake_quant_with_min_max_args` ops, but only if `inference_input_type = tf.int8` AND `inference_output_type = tf.int8`

lgeiger:
Sorry for the late reply.

> In that case these checks are not necessary, or am I missing something here?

To be honest, I am not 100% sure about this. This code was originally adapted from https://github.com/tensorflow/tensorflow/commit/ec94fabb838b0934eeaf46f1fd22c36b2b3ec333 but it looks like there've been [additional changes in the latest version](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/python/util.py) which would be great to apply here as well in another PR.
I would need to take a closer look if these errors would ever be thrown here, but if you think they won't, fell free to remove the checks for them in `strip_lcedequantize_ops`.

Could you still add a unittest similar to [this one](https://github.com/larq/compute-engine/blob/9297ff48c7bec562dc25aaad4853e65a351b710c/larq_compute_engine/tests/end2end_test.py#L203-L219) that tests that the functionality works as expected?
I do understand that you manually verified that the code works, but it is important for us to have these checks also run on CI so we can ensure that it also works with older versions of TensorFlow and crucially that the code will keep working and we don't accidentally break it in a future PR.

simonmaurer:
> Sorry for the late reply.
> 
> > In that case these checks are not necessary, or am I missing something here?
> 
> To be honest, I am not 100% sure about this. This code was originally adapted from [tensorflow/tensorflow@ec94fab](https://github.com/tensorflow/tensorflow/commit/ec94fabb838b0934eeaf46f1fd22c36b2b3ec333) but it looks like there've been [additional changes in the latest version](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/python/util.py) which would be great to apply here as well in another PR.
> I would need to take a closer look if these errors would ever be thrown here, but if you think they won't, fell free to remove the checks for them in `strip_lcedequantize_ops`.
> 
> Could you still add a unittest similar to [this one](https://github.com/larq/compute-engine/blob/9297ff48c7bec562dc25aaad4853e65a351b710c/larq_compute_engine/tests/end2end_test.py#L203-L219) that tests that the functionality works as expected?
> I do understand that you manually verified that the code works, but it is important for us to have these checks also run on CI so we can ensure that it also works with older versions of TensorFlow and crucially that the code will keep working and we don't accidentally break it in a future PR.

@lgeiger sorry for the delay.
I've added a small unit test similar to `end2end_test` to verify the `strip_lcedequantize_ops`

lgeiger:
Thanks for adding the test. Could you check the linting errors on CI? It looks like `pyflakes` complains about some imports in the test file.

simonmaurer:
@lgeiger fixed all import errors, reformatted code using black code style.
I've decided to put the unit test for the utility function `strip_lcedequantize_ops` in a separate `strip_lcedequantize_ops.py`module to keep `end2end_test.py` untouched, do you agree on this ?

lgeiger (approved):
@simonmaurer Sorry for the delay in reviewing this. I just rebased your PR onto master and added the unittests to our CI infrastructure in da432fd24a88c5163ae292c870e67063f2a9dbeb

Let's get this merged once CI passes 🚀

lgeiger:
I just rebased ontop of `main` to pull in #644, let's see what CI thinks

lgeiger (changes requested):
It looks like the unittests you added fail on CI. @simonmaurer could you take a look at what's going on there?

You should be able to reproduce the failure by running `bazelisk test larq_compute_engine/tests:strip_lcedequantize_test --test_output=all`

simonmaurer:
> It looks like the unittests you added fail on CI. @simonmaurer could you take a look at what's going on there?
> 
> You should be able to reproduce the failure by running `bazelisk test larq_compute_engine/tests:strip_lcedequantize_test --test_output=all`

yes, will have a look at the Unit test

larq_compute_engine/tests/strip_lcedequantize_test.py
```diff
@@ -0,0 +1,78 @@
...
+    interpreter = tf.lite.Interpreter(model_content=model_lce)
+    input_details = interpreter.get_input_details()
+    assert len(input_details) == 1
+    assert input_details[0]["dtype"] == inference_input_type.as_numpy_dtype
+    output_details = interpreter.get_output_details()
+    assert len(output_details) == 1
+    if inference_output_type == tf.float32:
+        assert output_details[0]["dtype"] == tf.int32.as_numpy_dtype
+    else:
+        assert output_details[0]["dtype"] == inference_output_type.as_numpy_dtype
```
lgeiger, line 74: Does this mean if output inference type is `tf.int8` the dequantize op is not removed?
@simonmaurer Is this intended behaviour?

lgeiger (commented):
Thanks for sticking with this PR, I just have an additional question below about the intended behaviour the unittest is verifying.

larq_compute_engine/tests/strip_lcedequantize_test.py
simonmaurer: yes, this is intended behavior.

as discussed here:
> Sorry for the late reply.
> 
> > In that case these checks are not necessary, or am I missing something here?
> 
> To be honest, I am not 100% sure about this. This code was originally adapted from [tensorflow/tensorflow@ec94fab](https://github.com/tensorflow/tensorflow/commit/ec94fabb838b0934eeaf46f1fd22c36b2b3ec333) but it looks like there've been [additional changes in the latest version](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/python/util.py) which would be great to apply here as well in another PR.
> I would need to take a closer look if these errors would ever be thrown here, but if you think they won't, fell free to remove the checks for them in `strip_lcedequantize_ops`.
> 
> Could you still add a unittest similar to [this one](https://github.com/larq/compute-engine/blob/9297ff48c7bec562dc25aaad4853e65a351b710c/larq_compute_engine/tests/end2end_test.py#L203-L219) that tests that the functionality works as expected?
> I do understand that you manually verified that the code works, but it is important for us to have these checks also run on CI so we can ensure that it also works with older versions of TensorFlow and crucially that the code will keep working and we don't accidentally break it in a future PR.

and when looking at the graph of `tf.float32/tf.int8` models there are only `LceDequantize` ops when the model has `tf.float32` outputs.
in other words the `LceDequantize` ops are removed (resulting in the desired tf.int32 outputs) iff the model has `tf.float32` outputs

simonmaurer: actually it shouldn't differ..as long as there are `LceDequantize` ops as outputs they should be removed to get `tf.int32` outputs, no matter what `tf.float32/tf.int8` ops are inside the model

simonmaurer: MLIR failed again for the assertion tests

lgeiger: You might also want to apply similar changes as done in #635 to this PR, but I don't think this will fix CI either. Looks like the dequantize node is not correctly removed in your example.

simonmaurer (commented):
@lgeiger after inspecting the unittest code again, I deactivated the default int8 ranges for `tf.float32` models. as you said the function `strip_lcedequantize_ops` would never remove the `LceDequantize` ops since any model that was generated via the parametrized pytest functions resulted in `tf.int8` outputs

lgeiger (changes requested):
Still looks like CI is failing. Are you able to reproduce the failure locally?

simonmaurer:
> Still looks like CI is failing. Are you able to reproduce the failure locally?

sorry if this didn't make it into the v0.6 release..didn't know your time schedule. it's been taking a while since I always thought I had to recompile everything.
gonna use your hint:
`bazelisk test larq_compute_engine/tests:strip_lcedequantize_test --test_output=all`

the code works for `tf.float32` models, it is related to the unittests only. working on it...

lgeiger:
> sorry if this didn't make it into the v0.6 release..didn't know your time schedule.

No worries, we can always push a patch release later.

> it's been taking a while since I always thought I had to recompile everything.

This should now be a bit better since we are now relying on a stable TF version so bazel should be able to properly cache the build artifacts.

larq_compute_engine/tests/strip_lcedequantize_test.py
```diff
@@ -6,7 +6,10 @@
 
 from larq_compute_engine.mlir.python.converter import convert_keras_model
 from larq_compute_engine.mlir.python.util import strip_lcedequantize_ops
+<<<<<<< HEAD
 from larq_compute_engine.tflite.python.interpreter import Interpreter
+=======
+>>>>>>> 7cf93f8b5fc3ab18d0de9897e67f9598e547ccb1
```
AdamHillier, line 12: @simonmaurer I think you accidentally commited merge indicators

simonmaurer:
@lgeiger testing `tf.float32` model outputs (w/ and w/o quantize ops inside model). after some more testing I do think the Unit tests are failing for `tf.int8` model outputs because there is a `Dequantize` op after model conversion to TFLite (before strip_lcedequantize_ops).

simonmaurer:
> You might also want to apply similar changes as done in #635 to this PR, but I don't think this will fix CI either. Looks like the dequantize node is not correctly removed in your example.
@lgeiger I assume this is exactly what you meant here and not the Larq dequantize op ;)

simonmaurer:
@lgeiger @AdamHillier finally the Unit tests work for both tf.float32 and tf.int8-quantized models. additionally I had to adapt the code to the following cases.
1. int8-quantized models (inference_output_type=tf.float32) ie when the output is dequantized back to tf.float32 (meant that there is the additional Dequantize op of TFLite: LceConv2D -> LceDequantize -> Dequantize)
2. int8-quantized models (inference_output_type=tf.int8) ie when the iutput is tf.int8 (LceConv2D -> LceDequantize)

simonmaurer:
@lgeiger @Tombana request a review given the latest commit as the PR is now ready including the updates on the unit tests for tf.float32/tf.int8 models

Tombana (approved):

lgeiger (approved):
Great! Sorry for the long wait, let's get this merged now 🎉

PR #627, opened 2021-04-09
Configure the Android SDK/NDK interactively in configure.py.

AdamHillier:
This means we don't have to hard-code paths, and Android only needs to be configured when necessary (it won't be configured by default). Configuration code taken from TF.

Hopefully closes #598, which is a prerequisite for upgrading to Bazel 3.7, which is a prerequisite for upgrading to TF 2.5.

PR #628, opened 2021-04-10
Upgrade TF dependency to v2.5.

AdamHillier:
## What do these changes do?
This PR updates the TensorFlow dependency to recently released ~~v2.5-rc0~~ v2.5-rc2.

## How Has This Been Tested?
CI.

## Benchmark Results
N/A, though we should collect new benchmark numbers before making a new release.

## Related issue number
~~Depends on #627, will be a draft until then.~~


.bazelrc
```diff
@@ -1,81 +1,123 @@
+################################################################
+# Config from TensorFlow's .bazelrc.                           #
+#                                                              #
+# This is copied directly, with some irrelevant parts removed. #
+################################################################
```
AdamHillier, line 5: I've re-arranged the bazelrc to make it more maintainable in the future. There are now two sections: the first (the bulk of it) is copied directly from TF's bazelrc, but with some sections (e.g. relating to GPUs) removed. The much smaller second section at the bottom adds LCE-specific config.

WORKSPACE
```diff
@@ -47,89 +11,25 @@ arm_compiler_configure(
...
+load("@org_tensorflow//tensorflow:workspace1.bzl", "tf_workspace1")
 
-bazel_version_repository(name = "bazel_version")
+tf_workspace1()
 
-load("@org_tensorflow//third_party/googleapis:repository_rules.bzl", "config_googleapis")
+load("@org_tensorflow//tensorflow:workspace0.bzl", "tf_workspace0")
 
-config_googleapis()
+tf_workspace0()
```
AdamHillier, line 35: The workspace file is *much* simplified, which is great. TF made some changes upstream to allow downstream projects to easily import the entire TF workspace config, which is what we're doing here. This means we no longer need to register our own versions of dependency repos, for example. See e.g. [IREE](https://github.com/google/iree/blob/29610cebc4174225b5d6785703c63e6053409b68/integrations/tensorflow/WORKSPACE) for another example of this.

larq_compute_engine/mlir/BUILD
```diff
@@ -16,6 +16,7 @@ gentbl(
     ],
     tblgen = "@llvm-project//mlir:mlir-tblgen",
     td_file = "ir/lce_ops.td",
+    td_includes = ["external/org_tensorflow"],
```
AdamHillier, line 19: I spent hours debugging why the `gentbl` rules weren't working (they complained about not being able to find the TF tablegen files). It turns out that the implementation of `gentbl` was [recently re-written](https://github.com/tensorflow/tensorflow/commit/76e6fd8ad9f4fd683c830118b941c42f90e1ad04#diff-1a9f4c347a8d3612bfaa2b92a030402ffb15a6fe9e7bf78fd0f13f006824cfed) and in the previous version there was a blanket hard-coded include of the TF workspace, which no longer survives. This ugliness is the only way I could work out how to get these working again; the `td_srcs` argument is only used for building the action graph and doesn't get fed into the `tblgen` binary, and the `deps` argument doesn't work either if a filegroup is passed in.

They've also implemented a new Bazel rule [`td-library`](https://github.com/tensorflow/tensorflow/commit/a275a1467fbce174e5bbd3b941733154186c563e#diff-1a9f4c347a8d3612bfaa2b92a030402ffb15a6fe9e7bf78fd0f13f006824cfed) but it's not actually used yet; when it is, I *think* that it will solve our issues as e.g. `tensorflow_ops_td_files` will be implemented as a `td-library` rather than a filegroup, which can then be passed in to `deps`.

third_party/arm_compiler.BUILD
```diff
@@ -1,75 +0,0 @@
-package(default_visibility = ["//visibility:public"])
-
-filegroup(
-    name = "gcc",
-    srcs = glob(["bin/*-gcc"]),
-)
```
AdamHillier, line 6: The arm compiler definition and indeed all of `third_party/toolchains` already exists in the TF upstream; now that we import the TF workspace directly, we get those for free and no longer need to define them in LCE ourselves, so they can be deleted.

AdamHillier: Hmm or perhaps not, since CI isn't happy with it. It seemed to work locally in a docker container, but I'll investigate.

larq_compute_engine/mlir/BUILD
lgeiger: Thanks for figuring this out! 🚀 
I briefly looked into upgrading as well but didn't make it past this road block.

.bazelrc
```diff
@@ -1,81 +1,123 @@
...
+build:macos --cxxopt=-std=c++14
+build:macos --host_cxxopt=-std=c++14
+build:windows --cxxopt=/std:c++14
+build:windows --host_cxxopt=/std:c++14
 
 # On windows, we still link everything into a single DLL.
 build:windows --config=monolithic
 
+# On linux, we dynamically link small amount of kernels
+build:linux --config=dynamic_kernels
```
lgeiger, line 109: Does this work our wheel builds?

.bazelrc
```diff
@@ -1,81 +1,123 @@
...
+# Default paths for TF_SYSTEM_LIBS
+build:linux --define=PREFIX=/usr
+build:linux --define=LIBDIR=$(PREFIX)/lib
+build:linux --define=INCLUDEDIR=$(PREFIX)/include
+build:linux --define=PROTOBUF_INCLUDE_PATH=$(PREFIX)/include
+build:macos --define=PREFIX=/usr
+build:macos --define=LIBDIR=$(PREFIX)/lib
+build:macos --define=INCLUDEDIR=$(PREFIX)/include
+build:macos --define=PROTOBUF_INCLUDE_PATH=$(PREFIX)/include
+# TF_SYSTEM_LIBS do not work on windows.
```
lgeiger, line 93: Do we need these include paths for our builds, I thought bazel would handle that correctly?

.bazelrc
```diff
@@ -84,55 +126,103 @@ build:windows --linkopt=/OPT:REF
 build:windows --host_linkopt=/OPT:REF
 build:windows --linkopt=/OPT:ICF
 build:windows --host_linkopt=/OPT:ICF
-build:windows --copt=/d2ReducedOptimizeHugeFunctions
-build:windows --host_copt=/d2ReducedOptimizeHugeFunctions
-build:windows --experimental_strict_action_env=true
```
lgeiger, line 89: Technically these things are also inherited from TensorFlow, but they set it in `configure.py`.

lgeiger (approved):
This is very cool, thanks for updating. Not sure what the CI failures are about, but we can take a look at it together later.

.bazelrc
AdamHillier: I haven't checked, but once the tests pass I'll do a test release.

third_party/arm_compiler.BUILD
lgeiger: Could it be that they are using a different compiler version?

AdamHillier: Yeah so it turns out that Aarch64 works but Arm32 doesn't.

It's because their definition of [compiler pieces](https://github.com/tensorflow/tensorflow/blob/22ba168c4b843c3f3a5c397f2bd45552d8477fbb/arm_compiler.BUILD#L43-L51) differs from [what we had before](https://github.com/larq/compute-engine/blob/f7f9352ba804f30e2fe1c6a9742c97699807ddf6/third_party/arm_compiler.BUILD#L43-L51) - I'm trying to work out how to work around this.

AdamHillier: I've opened https://github.com/tensorflow/tensorflow/pull/48479 to fix the `compiler_pieces` upstream; if they merge it quickly then we can ask for a cherry-pick onto the `r2.5` branch and hopefully use that directly. Otherwise, we'll just have to revert to registering our own cross-compiler like we previously did.

lgeiger: Nice, thanks for the fix. Alternatively, we could also try their [embedded toolchains](https://github.com/tensorflow/tensorflow/tree/master/third_party/toolchains/embedded/arm-linux) which seem to do a very similar thing.

AdamHillier: Yeah that's a good point, those are also a newer GCC version. There must be some reason that they still have the older toolchain around, maybe to support older RPi boards? There's a comment in [rpi-newer-crosstools](https://github.com/rvagg/rpi-newer-crosstools/tree/eb68350c5c8ec1663b7fe52c742ac4271e3217c5) about OS support, so maybe it's that.

lgeiger: Yes, I think the `rpi-newer-crosstools` predate the embedded toolchains which have been added in https://github.com/tensorflow/tensorflow/commit/4961f18733ca3967198393abf419e14476b4a85c to enable TFLite `pip` wheels for RPI in https://github.com/tensorflow/tensorflow/commit/ff997783433bd0975c4daab84facd22d139d555a

.bazelrc
```diff
@@ -84,46 +126,103 @@ build:windows --linkopt=/OPT:REF
...
-build:android_arm64 --config=android
-build:android_arm64 --cpu=arm64-v8a
-build:android_arm64 --fat_apk_cpu=arm64-v8a
-build:android_x86 --config=android
-build:android_x86 --cpu=x86
-build:android_x86 --fat_apk_cpu=x86
-build:android_x86_64 --config=android
-build:android_x86_64 --cpu=x86_64
-build:android_x86_64 --fat_apk_cpu=x86_64
+# Address sanitizer
```
CNugteren, line 143: Very nice that this is now included by TensorFlow itself (right?), I'm a big fan of the [Address Sanitizer](https://clang.llvm.org/docs/AddressSanitizer.html). We could later consider running some of our CI tests with the `asan` target such that they benefit from all kinds of checks (at a ~2x runtime overhead).

AdamHillier: Yeah I was pleasantly surprised to see this too, and I agree we should definitely at some point set up some CI using it :)

third_party/arm_compiler.BUILD
AdamHillier: I hadn't realised, but this is indeed what they recommend for RPi boards: https://www.tensorflow.org/lite/guide/build_arm#cross-compilation_for_arm_with_bazel

.bazelrc
AdamHillier: Ah yes, added the reduce huge functions thing to configure.py

.bazelrc
AdamHillier: I copied this just because it was in TF's `.bazelrc`. I *think* it's so that you can have protobufs pre-compiled and not have to compile it as part of the build process, but I'm not sure. I agree that whatever it is though, we aren't using it, so it should be fine to remove it.

third_party/arm_compiler.BUILD
AdamHillier: And indeed using the embedded linux toolchains fixed the ARM tests :tada:

.bazelrc
```diff
@@ -84,46 +126,107 @@ build:windows --linkopt=/OPT:REF
...
+
+# Disable unnecesary features
+build --config=noaws --config=nogcp --config=nohdfs --config=nonccl
+
+# Extra build options we need for windows.
+build:windows --experimental_strict_action_env=true
+
+# Disable certain warnings that come from TF code and pollute logs
+build:android --copt=-Wno-deprecated-declarations
+build:linux --copt=-Wno-deprecated-declarationsq
```
lgeiger, line 222: Simple typo fix
```suggestion
build:linux --copt=-Wno-deprecated-declarations
```

.bazelrc
AdamHillier: Yeah it doesn't seem to do anything for us, so I've removed it.

lgeiger:
Nice, looks like we are getting closed:
```
larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc:7:10: fatal error: mlir/IR/Module.h: No such file or directory
    7 | #include "mlir/IR/Module.h"
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.
```

AdamHillier:
> Nice, looks like we are getting closed:

Yeah, just building locally so that I can debug these properly, but fingers crossed not too far left to go :)

.bazelrc
```diff
@@ -1,71 +1,102 @@
...
+################################################################
+
+# Default build options. These are applied first and unconditionally.
+
+# The following line opts in to modular op registration support by default.
+build --define framework_shared_object=true
+
+# For workaround https://github.com/bazelbuild/bazel/issues/8772 with Bazel >= 0.29.1
+build --java_toolchain=@tf_toolchains//toolchains/java:tf_java_toolchain
+build --host_java_toolchain=@tf_toolchains//toolchains/java:tf_java_toolchain
```
lgeiger, line 14: Is this a workaround we actually need, or is it just included for completeness and ease of future upgrades?

.bazelrc
lgeiger: Do you think it would make sense to move these configs into a different `.tensorflow.bazelrc` file and include it at the top with `try-import %workspace%/.tensorflow.bazelrc` here?

larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
```diff
@@ -95,25 +106,25 @@ pybind11::bytes ConvertGraphDefToTFLiteFlatBuffer(
...
-  }
-
-  TruncateOpOrArgLocNameMapper op_or_arg_name_mapper;
-  // Write MLIR TFLite dialect into FlatBuffer
   std::string result;
-  if (tflite::MlirToFlatBufferTranslateFunction(
-          *module.ValueOrDie(), &result, /*emit_builtin_tflite_ops=*/true,
-          /*emit_select_tf_ops=*/false, /*emit_custom_ops=*/true,
-          &op_or_arg_name_mapper)) {
+  auto status = ConvertTFExecutorToTFLOrFlatbuffer(
```
lgeiger, line 121: I previously used the lower lever `MlirToFlatBufferTranslateFunction` API  so we do not need rely on `@org_tensorflow//tensorflow/compiler/mlir/lite:tf_to_tfl_flatbuffer`. Last time I checked this target included many for us unneeded dependencies which I I opted to not use it. However, I am not sure if this is relevant anymore since our builds take ages anyway.
Do we gain an advantage in using `ConvertTFExecutorToTFLOrFlatbuffer` over `MlirToFlatBufferTranslateFunction` here?

larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
```diff
@@ -95,25 +106,25 @@ pybind11::bytes ConvertGraphDefToTFLiteFlatBuffer(
           default_ranges.cast<std::pair<double, double>>();
     }
   }
-  mlir::PassManager pm(&context);
+
+  mlir::PassManager pm(&context, mlir::OpPassManager::Nesting::Implicit);
+  tensorflow::SetCrashReproducer(pm);
```
lgeiger, line 111: Nice :+1:

lgeiger (commented):
This looks great! I only have a few minor comments and questions.

.bazelrc
AdamHillier: I just included it for completeness but happy to remove it if you prefer

lgeiger: Nah that's fine we can keep them, I was just wondering.

larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
AdamHillier: Yeah I was getting a bunch of build errors from the invokation of `MlirToFlatBufferTranslateFunction` which is why I went 'up a level' here in order to get it working. Also I guess I thought mirroring what TF do more closely would make the upgrade easier, which indeed it did :p

I didn't realise though that the extra dependency would bring in a lot of transitive deps that weren't already having to build anyway - do you know which bits those are and why we're able to avoid them but TF aren't? For that matter, what does `ConvertTFExecutorToTFLOrFlatbuffer` actually do beyond what we need it to? To be clear I'm super happy to try and revert to the previous method, this is just for my understanding :)

.bazelrc
AdamHillier: Yeah that's a very good idea actually :+1:

larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
lgeiger: > do you know which bits those are and why we're able to avoid them but TF aren't

I can check if this is actually still the case in TF 2.5, but last time I checked it included some of the flatbuffer import logic that we don't need. I will look into this and see if this is still relevant since we are including most of TF already anyway.

lgeiger: I checked using `bazelisk query "deps(:build_pip_pkg)" --notool_deps` and compared the diff. It looks like `tf_to_tfl_flatbuffer` adds to following targets:
```
@org_tensorflow//tensorflow/compiler/mlir/lite:common
@org_tensorflow//tensorflow/compiler/mlir/lite:common/tfl_pass_config.h
@org_tensorflow//tensorflow/compiler/mlir/lite:flatbuffer_import
@org_tensorflow//tensorflow/compiler/mlir/lite:flatbuffer_import.cc
@org_tensorflow//tensorflow/compiler/mlir/lite:flatbuffer_import.h
@org_tensorflow//tensorflow/compiler/mlir/lite:flatbuffer_translate_lib
@org_tensorflow//tensorflow/compiler/mlir/lite:tf_to_tfl_flatbuffer
@org_tensorflow//tensorflow/compiler/mlir/lite:tf_to_tfl_flatbuffer.cc
@org_tensorflow//tensorflow/compiler/mlir/lite:tf_to_tfl_flatbuffer.h
@org_tensorflow//tensorflow/compiler/mlir/tensorflow:tf_dialect_lib
@org_tensorflow//tensorflow/compiler/mlir/tensorflow:translate/tf_mlir_translate.cc
@org_tensorflow//tensorflow/compiler/mlir/tensorflow:translate/tf_mlir_translate.h
@org_tensorflow//tensorflow/compiler/mlir/tensorflow:translate_lib
@org_tensorflow//tensorflow/lite/tools/optimize:quantize_weights
@org_tensorflow//tensorflow/lite/tools/optimize:quantize_weights.cc
@org_tensorflow//tensorflow/lite/tools/optimize:quantize_weights.h
```
TBH this doesn't seem to be as bad as I remember. Maybe this doesn't matter anymore since we now include the fallback kernel implementations anyway.

larq_compute_engine/tflite/benchmark/BUILD
```diff
@@ -7,12 +8,12 @@ package(
     licenses = ["notice"],  # Apache 2.0
 )
 
-cc_binary(
```
Tombana, line 10: What does `tf_cc_binary` do? Should we use it for every tflite related bazel target?

AdamHillier: Basically it adds a bunch of linkopts and other dependencies in order to make builds/tests work when using `framework_shared_object`, which is the reason I originally added them. We're not actually using `framework_shared_object`, but I figured that it doesn't hurt to add them. Also for some reason the benchmark binary is currently failing to build on `android_arm64` so I thought adding this might fix that (alas it did not).

AdamHillier:
So, now CI is green, I've run a few test release builds: https://github.com/larq/compute-engine/actions/runs/751501689

* MacOS builds pass (Python 3.6-3.9) ✅✅✅✅
* One Windows build passes (Python 3.7), but the other three time-out. Once https://github.com/tensorflow/tensorflow/pull/48525 is merged we will be able to disable GCP and AWS builds on Windows as we do on Linux/Mac which will hopefully help with compile times. ✅❌❌❌ 
* One Linux build currently fails with a `internal compiler error: Segmentation fault`, which then stops the other builds. Python 3.9 is currently unsupported in the `custom-op` Docker image that we use (even the TF-nightly version). 😱❓❓❌
* The Android `aar` build passes ✅
* Arm32 and Aarch64 benchmark binaries build successfully, but the Android benchmark binary fails (should be fixed by https://github.com/tensorflow/tensorflow/pull/48546) ✅✅❌

lgeiger:
Thanks for tackling the update and the extended testing!

> Python 3.9 is currently unsupported in the custom-op Docker image that we us

We currently still rely on an old TensorFlow image to do our builds. Therefore we install our custom Python version in the linux build skript and have a bunch of workarounds: https://github.com/larq/compute-engine/blob/657ef7984db5865c2469573b8e5f0d3553507c96/.github/tools/release_linux.sh#L9-L10
Maybe we can try adding a similar workaround for Python 3.9 as we did here:
https://github.com/larq/compute-engine/blob/657ef7984db5865c2469573b8e5f0d3553507c96/.github/tools/release_linux.sh#L18-L21
I tired updating the docker image at some point but ran into other errors and had to revert it in #558. We could look at what [TF-Addons is doing](https://github.com/tensorflow/addons/blob/master/tools/docker/build_wheel.Dockerfile), but that probably will require us rewrite our Linux build skript to remove all the workarounds that are tied to the ancient image although that might fix the compile errors.

> internal compiler error: Segmentation fault

Do you think this could be caused by `bazel` simply running out of memory similar to how it sometimes did on Windows in the past?

AdamHillier:
> Maybe we can try adding a similar workaround for Python 3.9

Ah nice, yeah that's a good idea. I only very briefly tried 3.9 on Linux, just by updating the image tag to `nightly-custom-op`. But I'll try again with edits to `release_linux.sh` as you suggest :)

> > internal compiler error: Segmentation fault
>
> Do you think this could be caused by bazel simply running out of memory similar to how it sometimes did on Windows in the past?

Ah yes could well be, definitely something to investigate. There are a few Bazel options that I enabled for Windows to reduce memory usage, it's probably worth trying those with Linux too. Maybe also configuring the swap space.

lgeiger:
I pushed a fix to reenable Python 3.9 builds on Linux. Note that this was tested on a large cloud VM, so the compiler segfaults did not reproduce.

I think for the remaining Linux build failures on CI our best shot is probably to enable the memory saving flags similar to our Windows builds or try and see if the targets added in https://github.com/larq/compute-engine/pull/628#discussion_r613301169 cause the problems.

lgeiger:
[Linux builds](https://github.com/larq/compute-engine/actions/runs/786913354) are now working for all Python versions, however the Windows builds still timeout after 6h (which is the maximum allowed runtime for GitHub actions).

AdamHillier:
Ah great, thanks for pushing this further Lukas :)

https://github.com/tensorflow/tensorflow/pull/48525 just got merged, which is nice, so I'll open a PR to cherry-pick that onto the 2.5 branch, and if that goes smoothly it might help with the Windows builds.

AdamHillier:
> [tensorflow/tensorflow#48525](https://github.com/tensorflow/tensorflow/pull/48525) just got merged, which is nice, so I'll open a PR to cherry-pick that onto the 2.5 branch, and if that goes smoothly it might help with the Windows builds.

And the cherry-pick is now merged, so I've updated the TF dependency and am running the test release again: https://github.com/larq/compute-engine/runs/2498846429?check_suite_focus=true. Fingers crossed the Windows builds now pass.

AdamHillier:
Hmm sadly the Windows builds still don't pass, they get roughly 90% of the way there: `9,017 / 9,710 targets` within the six hour limit.

I'll try and investigate other options for speeding things up.

.github/workflows/release.yml
```diff
@@ -221,7 +221,7 @@ jobs:
 
           "" | python configure.py
 
-          bazelisk --output_base=C:\build_output build :build_pip_pkg --enable_runfiles --discard_analysis_cache --notrack_incremental_state --nokeep_state_after_build --nouse_action_cache --local_ram_resources=4096
+          bazelisk --output_base=C:\build_output build :build_pip_pkg --compilation_mode=fastbuild --enable_runfiles --discard_analysis_cache --notrack_incremental_state --nokeep_state_after_build --nouse_action_cache --local_ram_resources=4096
```
lgeiger, line 224: Do you think it is worth trying to remove the `--nouse_action_cache` flag? Not sure if it also slows down things for non-incremental builds here.

Tombana:
> Hmm sadly the Windows builds still don't pass, they get roughly 90% of the way there: `9,017 / 9,710 targets` within the six hour limit.
> 
> I'll try and investigate other options for speeding things up.

What if we add another bazel cache for these builds? Then they might fail a first time but should succeed a second time, and it would also just be faster in general. It might have be a separate cache from the normal CI tests since the build configuration is completely different and we don't want to invalidate the normal cache (although maybe bazel already takes care of this properly by itself).

AdamHillier:
I spent a while this evening running a Windows build on my home machine, and didn't learn much, except that I saw lots of MKL targets - turns out that MKL is built on Windows indescriminately, even if you try and disable it (also on Linux, for that matter): https://github.com/tensorflow/tensorflow/blob/5dcfc51118817f27fad5246812d83e5dccdc5f72/third_party/mkl/build_defs.bzl#L37. This might explain why MacOS builds are relatively a lot quicker.

lgeiger:
Thanks for investigating! These logs also show up on CI and unfortunately 9922add4c15109398ac4798878fb000a64378b8b  makes the build crash :(

Do you think we can use the Windows equivalent of `-O2` for the builds?

AdamHillier:
> Thanks for investigating! These logs also show up on CI and unfortunately [9922add](https://github.com/larq/compute-engine/commit/9922add4c15109398ac4798878fb000a64378b8b) makes the build crash :(
> 
> Do you think we can use the Windows equivalent of `-O2` for the builds?

Ah yeah I know, it turns out that TF is just completely incompatible with `--config=fastbuild`. I tried adding `-DNDEBUG` but that didn't help, tested on my local machine.

I think `/O2` is probably what's already being used by Bazel with `--config=opt`, and in MSVC there is no `/O3`: https://docs.microsoft.com/en-us/cpp/build/reference/compiler-options-listed-by-category?view=msvc-160#optimization

.github/workflows/release.yml
lgeiger: Unfortunately this suggestion doesn't help, it also doesn't make it through in time.

WORKSPACE
```diff
@@ -11,10 +11,13 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 #    reliable downloads.
 http_archive(
     name = "org_tensorflow",
-    sha256 = "a17951020295c8bbf129787415efdc926128efaff5914b6c66fb1b5c763885a9",
-    strip_prefix = "tensorflow-7a29693084fe074aaea588e0de46479404006146",
+    patch_args = ["-p1"],
+    patch_tool = "patch",
+    patches = ["//third_party/tensorflow_patches:tf_pr_48546.patch"],
```
lgeiger, line 16: Nice!

AdamHillier: Yeah it's nice isn't it :)

The only problem is what to do about the makefile build. I discussed with @Tombana whether we even need the makefile build any more and we decided that we probably still do, since it is much easier to use than Bazel on e.g. a Raspberry Pi compiling natively.

But then on the other hand this is just a patch to Bazel build files, so I guess we don't need to apply it for the makefile build.

.github/workflows/release.yml
```diff
@@ -221,7 +224,7 @@ jobs:
...
```
lgeiger:
```suggestion
          bazelisk --output_base=C:\build_output build :build_pip_pkg --enable_runfiles --discard_analysis_cache --notrack_incremental_state --nokeep_state_after_build --nouse_action_cache --local_ram_resources=4096
```

larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
```diff
@@ -58,12 +58,23 @@ pybind11::bytes ConvertGraphDefToTFLiteFlatBuffer(
     throw std::runtime_error("Invalid target.");
   }
 
+  // `ParseInputArrayInfo` requires a type that isn't pybind compatible, so
+  // translate here.
+  std::vector<llvm::Optional<std::vector<int>>> translated_input_shapes;
```
lgeiger, line 63: Won't pybind recognise this type if we change the signature of `ConvertGraphDefToTFLiteFlatBuffer` to accept `std::vector<llvm::Optional<std::vector<int>>>`?

lgeiger (approved):
Great! I just have one minor comment, other than that this looks good.

larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
AdamHillier: I don't think so, no. I tried that previously and got an error. I think that makes sense though because pybind has no way of knowing how to construct an element with type `llvm::Optional`.

AdamHillier:
The test release looks good: https://github.com/larq/compute-engine/actions/runs/857230232

Actually only one build timed out, 3/4 of the Windows ones succeeded. But as discussed above let's resolve the Windows build time issues later in a future PR.

larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
```diff
@@ -95,25 +106,25 @@ pybind11::bytes ConvertGraphDefToTFLiteFlatBuffer(
...
 
   // Convert back to outlined while format for export back to flatbuffer.
   pm.addPass(mlir::TFL::CreateWhileOutlinePass());
   pm.addPass(mlir::TFL::CreateRuntimeVerifyPass());
 
-  if (failed(pm.run(*module.ValueOrDie()))) {
-    throw std::runtime_error("Could not complete conversion passes.");
-  }
-
-  TruncateOpOrArgLocNameMapper op_or_arg_name_mapper;
```
lgeiger, line 110: @AdamHillier Do we not need this truncation anymore, or is it now handled within the conversion function?

We added it in https://github.com/larq/compute-engine/pull/349/commits/df24c2b7f9b1f01631acab8c89f72160106afb42, not sure if we still want to keep this.

Sorry for not spotting this earlier.

AdamHillier: Ah sorry, I missed that too :p Thanks for the PR :)

PR #632, opened 2021-04-13
:arrow_up: tensorflow@2.5.0rc1

lgeiger:
This PR updates to RC1 and is built ontop of #628. We should probably only merge after CI passes in #628


AdamHillier (approved):
Nice, thanks

PR #644, opened 2021-05-20
Fuse `tfl.Quantize(lq.Dequantize(x))` -> `lq.Dequantize(x)`

lgeiger:
## What do these changes do?
`lq.Dequantize` supports `int8` output, so we can fuse `tfl.Quantize(lq.Dequantize(x))` -> `lq.Dequantize(x)` to remove an unnecessary `tfl.Quantize` op.

## How Has This Been Tested?
Added a MLIR filecheck test.

## Related issue number
I think this will be required to eventually make the unittests of #611 pass. /cc @simonmaurer


AdamHillier (approved):
🚀

PR #649, opened 2021-05-21
Fork `ConvertTFExecutorToTFLOrFlatbuffer` to re-enable argname truncation

lgeiger:
## What do these changes do?
This PR forks `ConvertTFExecutorToTFLOrFlatbuffer` to allow us to re-enable arg name truncation.

## How Has This Been Tested?
CI

## Related issue number
Fixes https://github.com/larq/compute-engine/pull/628#discussion_r636213211

### DIFF
--- a/larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
+++ b/larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
@@ -1,6 +1,7 @@
 #include <exception>
 
 #include "larq_compute_engine/mlir/tf_tfl_passes.h"
+#include "larq_compute_engine/mlir/tf_to_tfl_flatbuffer.h"
 #include "llvm/Support/MemoryBuffer.h"
 #include "llvm/Support/ToolOutputFile.h"
 #include "mlir/IR/MLIRContext.h"
@@ -10,7 +11,6 @@
 #include "pybind11/pytypes.h"
 #include "pybind11/stl.h"
 #include "tensorflow/compiler/mlir/lite/quantization/quantization_config.h"
-#include "tensorflow/compiler/mlir/lite/tf_to_tfl_flatbuffer.h"
 #include "tensorflow/compiler/mlir/lite/transforms/passes.h"
 #include "tensorflow/compiler/mlir/op_or_arg_name_mapper.h"
 #include "tensorflow/compiler/mlir/tensorflow/translate/import_model.h"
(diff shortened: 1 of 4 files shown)


larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
```diff
@@ -118,11 +107,8 @@ pybind11::bytes ConvertGraphDefToTFLiteFlatBuffer(
   pm.addPass(mlir::TFL::CreateRuntimeVerifyPass());
 
   std::string result;
-  auto status = ConvertTFExecutorToTFLOrFlatbuffer(
-      module->get(), /*export_to_mlir=*/false, /*emit_builtin_tflite_ops=*/true,
-      /*emit_select_tf_ops=*/false, /*emit_custom_ops=*/true,
-      /*select_user_tf_ops=*/{}, quant_specs, /*saved_model_tags=*/{}, &result,
-      &pm);
+  auto status = ConvertTFExecutorToFlatbuffer(
+      module->get(), /*export_to_mlir=*/false, &result, &pm);
```
lgeiger, line 111: I kept the `export_to_mlir` option since it might come in handy for debugging later on.

larq_compute_engine/mlir/tf_to_tfl_flatbuffer.cc
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  if (export_to_mlir) {
+    llvm::raw_string_ostream os(*result);
+    module.print(os);
+    return Status::OK();
+  }
+
+  // Write MLIR TFLite dialect into FlatBuffer
+  TruncateOpOrArgLocNameMapper op_or_arg_name_mapper;
```
lgeiger, line 85: This is the only functional modification compared to the upstream code.

AdamHillier (approved):

larq_compute_engine/mlir/python/graphdef_tfl_flatbuffer.cc
Tombana: That's amazing, I'm always searching for that option when I want to write new Filecheck tests.

larq_compute_engine/mlir/tf_to_tfl_flatbuffer.cc
Tombana: Should we maybe add that in a comment?
```cpp
// This is the only modification compared to the upstream tensorflow file
```
When we have to update this file in the next tf 2.6 upgrade we might have forgotten this by then.

Furthermore, in our private repo we have `TruncateOpOrArgLocNameMapper` in a separate file because its also used by the saved-model converter. Maybe this PR might be a nice place to also apply the change here to make merging easier.

lgeiger: > Furthermore, in our private repo we have TruncateOpOrArgLocNameMapper in a separate file because its also used by the saved-model converter. Maybe this PR might be a nice place to also apply the change here to make merging easier.

Yes, so `TruncateOpOrArgLocNameMapper` will not be needed anymore by the saved model converter after this change since it can just rely on `ConvertTFExecutorToFlatbuffer` directly. So merging will be as simple as removing the `TruncateOpOrArgLocNameMapper` file in our fork.

Tombana: Oh ofcourse, that makes sense. Thanks.

Tombana (approved):

PR #651, opened 2021-05-28
Add pytype linting to CI

lgeiger:
This PR add PyType linting to CI to verify our type signatures. We currently ignore some errors since TF and our pybind modules do not provide type stubs and might not exists during linting.


.github/workflows/lint.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           python-version: 3.8
       - name: Install Lint Dependencies
         run: |
-          pip install pyflakes==2.3.0 black==20.8b1 --no-cache-dir
+          pip install pyflakes==2.3.0 black==20.8b1 pytype --no-cache-dir
```
CNugteren, line 21: Wouldn't it be good to pin the version of `pytype`? Otherwise a new version might suddenly break some unrelated PR (because of a bug or a new feature).

lgeiger: Yeah, sorry I forgot to commit that change.

How do you feel about moving these to a `test_requirements.txt` file, so we can keep them up to date with dependabot?
I'm happy to do this in a future PR.

PR #654, opened 2021-06-08
Add constant folding for `lq.quantize` and `lq.dequantize`

lgeiger:
## What do these changes do?
This PR adds constant folders for `lq.quantize` and `lq.dequantize` ops.

This PR also includes a few minor changes from our internal fork to bring them inline with each other.

## How Has This Been Tested?
Added MLIR Filecheck

Co-Authored-By: Tom Bannink <Tombana@users.noreply.github.com>

### DIFF
--- a/larq_compute_engine/core/types.h
+++ b/larq_compute_engine/core/types.h
@@ -6,6 +6,8 @@
 #include <limits>
 #include <type_traits>
 
+#include "tensorflow/lite/kernels/internal/cppmath.h"
+
 namespace compute_engine {
 namespace core {
 
@@ -15,6 +17,12 @@ namespace core {
 #else
 #define LCE_LIKELY(condition) (condition)
 #define LCE_UNLIKELY(condition) (condition)
+#endif
+
+#if defined(__GNUC__)
+#define FORCE_INLINE __attribute__((always_inline)) inline
+#else
+#define FORCE_INLINE inline
 #endif
 
 // Check that 0 <= index < limit using a single comparison, assuming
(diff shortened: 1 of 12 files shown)


Tombana (approved):
Looks good to me, but maybe somebody else should also review it.

image

larq_compute_engine/core/types.h
```diff
@@ -17,6 +19,12 @@ namespace core {
 #define LCE_UNLIKELY(condition) (condition)
 #endif
 
+#if defined(__GNUC__)
+#define FORCE_INLINE __attribute__((always_inline)) inline
+#else
+#define FORCE_INLINE inline
+#endif
```
CNugteren, line 26: Maybe it is my search failing, but I don't think you use `FORCE_INLINE` somewhere?

CNugteren (approved):
Just a small comment, otherwise I approve. The name of the PR was a bit confusing to me though, I was going through the code to look for the place where you added a folder or moved something from one folder to another, but then it turns out it was about folding instead of folders (as in paths) :confused: :wink:

lgeiger:
> The name of the PR was a bit confusing to me though, I was going through the code to look for the place where you added a folder or moved something from one folder to another, but then it turns out it was about folding instead of folders (as in paths)

Good point, changed the title.

larq_compute_engine/core/types.h
lgeiger: Indeed it is not used anywhere at the moment. I mainly added it to sync `larq_compute_engine/core/types.h` with our internal fork to reduce confusing for us in the future. But I am happy to remove it, if you prefer.

Tombana: Yea i'd opt to keep public and private fork as much in sync as possible.

PR #655, opened 2021-06-08
Add support for saved model conversion

### LINKED ISSUE #407: Support conversion from saved model
Currently Larq Compute engine exposes [`convert_keras_model`](https://nightly.docs.larq.dev/compute-engine/api/converter/#convert_keras_model-function) as the only way to convert models. This is useful when converting models within the training script, but can be cumbersome when sharing networks since Keras `.h5` files might require additional code to properly load models  with custom layers.

The [default TFLiteConverter](https://www.tensorflow.org/api_docs/python/tf/lite/TFLiteConverter#from_saved_model) supports converting serialized models in the saved_model format, so I think it would be useful to support this in Larq Compute Engine as well.

Sidenote: Since TF 2.2 the Python side of the TFLite converter [had some major refactors](https://github.com/tensorflow/tensorflow/commits/master/tensorflow/lite/python/lite.py) so it would be good to check if this would make it easier for us to support saved models loading.

lgeiger:
## What do these changes do?
This PR enables support for directly converting saved models in TensorFlow.

## How Has This Been Tested?
We've been using this code internally for a while now and added some test cases to the end2end tests

## Related issue number
Closes #407

Co-Authored-By: Tom Bannink <Tombana@users.noreply.github.com>

### DIFF
--- a/larq_compute_engine/mlir/python/converter.py
+++ b/larq_compute_engine/mlir/python/converter.py
@@ -117,7 +117,8 @@ def convert_saved_model(
             must be either `tf.float32` or `tf.int8`.
         inference_output_type: Data type of the output layer. Defaults to `tf.float32`,
             must be either `tf.float32` or `tf.int8`.
-        target: Target hardware platform. Must be "arm", or "xcore".
+        target: Target hardware platform. Defaults to "arm", must be either "arm"
+            or "xcore".
         experimental_default_int8_range: Tuple of integers representing `(min, max)`
             range values for all arrays without a specified range. Intended for
             experimenting with quantization via "dummy quantization". (default None)
@@ -198,7 +199,8 @@ def convert_keras_model(
             must be either `tf.float32` or `tf.int8`.
         inference_output_type: Data type of the output layer. Defaults to `tf.float32`,
             must be either `tf.float32` or `tf.int8`.
-        target: Target hardware platform. Must be "arm" or "xcore".
+        target: Target hardware platform. Defaults to "arm", must be either "arm"
+            or "xcore".
         experimental_default_int8_range: Tuple of integers representing `(min, max)`
             range values for all arrays without a specified range. Intended for
             experimenting with quantization via "dummy quantization". (default None)
--- a/larq_compute_engine/mlir/python/converter.py
+++ b/larq_compute_engine/mlir/python/converter.py
@@ -1,11 +1,15 @@
+import os
 from packaging import version
 import warnings
-from typing import Optional, Tuple
+from typing import Optional, Tuple, Union
 import tensorflow as tf
+import tempfile
 
-from larq_compute_engine.mlir._graphdef_tfl_flatbuffer import (
+from larq_compute_engine.mlir._tf_tfl_flatbuffer import (
     convert_graphdef_to_tflite_flatbuffer,
+    convert_saved_model_to_tflite_flatbuffer,
 )
+
 from larq_compute_engine.mlir.python.util import modify_integer_quantized_model_io_type
 
 from tensorflow.core.framework.types_pb2 import DataType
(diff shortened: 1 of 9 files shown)


larq_compute_engine/mlir/python/converter.py
```diff
@@ -121,11 +200,30 @@ def convert_keras_model(
...
+                inference_input_type=inference_input_type,
+                inference_output_type=inference_output_type,
+                experimental_default_int8_range=experimental_default_int8_range,
+                experimental_enable_bitpacked_activations=experimental_enable_bitpacked_activations,
+                target=target,
+            )
+    except Exception:
+        warnings.warn(
+            "Saved-model conversion failed, falling back to graphdef-based conversion."
+        )
```
lgeiger, line 225: This is now enabled by default. Let me know if you prefer to keep this behind a flag instead.

larq_compute_engine/mlir/python/saved_model_tfl_flatbuffer.cc
```diff
@@ -0,0 +1,162 @@
...
+  if (!status.ok()) {
+    throw std::runtime_error(status.error_message());
+  }
+
+  mlir::TFL::QuantizationSpecs quant_specs;
+  // Normally we'd only set `inference_type` to QINT8 when there are fake_quant
+  // nodes in the graph. However this did not work reliably, and even for float
+  // models it is fine to set the inference type to QINT8, so we do that by
+  // default.
+  quant_specs.inference_type = tensorflow::DT_QINT8;
```
lgeiger, line 130: This is not ideal, but I don't think this causes any issues.

Tombana (approved):

larq_compute_engine/mlir/python/converter.py
```diff
@@ -57,6 +61,119 @@ def _contains_training_quant_op(graph_def):
...
+def convert_saved_model(
+    saved_model_dir: Union[str, os.PathLike],
+    *,  # Require remaining arguments to be keyword-only.
+    inference_input_type: tf.DType = tf.float32,
+    inference_output_type: tf.DType = tf.float32,
+    target: str = "arm",
+    experimental_default_int8_range: Optional[Tuple[float, float]] = None,
+    experimental_enable_bitpacked_activations: bool = False,
+) -> bytes:
+    """Converts a SavedModel to TFLite flatbuffer.
```
AdamHillier, line 105: I'm never sure how we should refer to things as being TensorFlow versus being Larq/LCE.

On the one hand, it's accurate because it _is_ a TFLite flatbuffer, just with extra custom ops. On the other hand, I feel like it's weird for a user to use the `larq_compute_engine` package and have method docstrings that refer to TensorFlow but not Larq or LCE.

I'm probably overthinking this though - how do you feel about it?

larq_compute_engine/mlir/python/converter.py
```diff
@@ -57,6 +61,119 @@ def _contains_training_quant_op(graph_def):
...
+            f.write(tflite_model)
+        ```
+
+    # Arguments
+        saved_model_dir: SavedModel directory to convert.
+        inference_input_type: Data type of the input layer. Defaults to `tf.float32`,
+            must be either `tf.float32` or `tf.int8`.
+        inference_output_type: Data type of the output layer. Defaults to `tf.float32`,
+            must be either `tf.float32` or `tf.int8`.
+        target: Target hardware platform. Must be "arm", or "xcore".
```
AdamHillier, line 120:
```suggestion
        target: Target hardware platform. Defaults to "arm", must be either "arm"
            or "xcore".
```

larq_compute_engine/mlir/python/saved_model_tfl_flatbuffer.cc
```diff
@@ -0,0 +1,162 @@
+#include <exception>
```
AdamHillier, line 1: Since this file is directly adapted from a TF file, I think we need to include [the license comment](https://github.com/tensorflow/tensorflow/blob/4ae0971abe3ad5c3a43c5c79283c98cfb08f4bdc/tensorflow/compiler/mlir/lite/python/saved_model_to_tfl_flatbuffer.cc#L1-L14).

AdamHillier (approved):
Looks great, thanks :)

Just a couple of minor comments.

larq_compute_engine/mlir/python/converter.py
lgeiger: This is pretty much copied from the `convert_keras_model` function, but I'm happy to rephrase it. Not sure how we refer to it in the guides though.

larq_compute_engine/mlir/python/saved_model_tfl_flatbuffer.cc
lgeiger: I wasn't aware that this was directly adapted from TF. I added a license comment.

larq_compute_engine/mlir/python/converter.py
AdamHillier: Yeah me neither, I think that's the real issue. Since we are going to be going through and updating our docs for the 0.6 release anyhow, we can maybe have a think about it then and not worry about it now.

PR #660, opened 2021-06-11
Upload release wheels even if some jobs fail.

AdamHillier:


Tombana (approved):

AdamHillier:
I triggered (then cancelled) a test release action just to validate that the yml is well formed: https://github.com/larq/compute-engine/actions/runs/928178209

PR #666, opened 2021-06-17
Disable mkl for windows builds

AdamHillier:
This PR adds a Bazel patch to the TF repo that disables MKL for Windows and Linux (unless the `build_with_mkl`/`enable_mkl` Bazel flags are provided).

It also removes the `/arch:AVX` build flag, which hopefully makes codegen a bit faster.

The goal is to get our Windows release builds to successfully complete within the 6 hour time limit of GitHubActions.

A test release from this branch passes (with Windows build durations of 5h40, 5h28, 5h12, 4h50): https://github.com/larq/compute-engine/actions/runs/943927532.

The variance on these Windows build times does seem to be pretty high, and those times are still quite close to the 6 hour limit, so I'm not convinced that even after doing this we could rely on these builds passing.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -219,6 +219,9 @@ jobs:
           $Env:BAZEL_VC = "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC"
           set PreferredToolArchitecture=x64
 
+          # The default is "/arch:AVX", which we don't want.
+          $Env:CC_OPT_FLAGS = "/O2"
+
           python --version
           python -m pip install wheel setuptools numpy six --no-cache-dir
 
--- a/third_party/tensorflow_patches/disable_forced_mkl.patch
+++ b/third_party/tensorflow_patches/disable_forced_mkl.patch
@@ -1,0 +1,27 @@
+diff --git a/third_party/mkl/build_defs.bzl b/third_party/mkl/build_defs.bzl
+index 806b157bad6..6c9f0172b97 100644
+--- a/third_party/mkl/build_defs.bzl
++++ b/third_party/mkl/build_defs.bzl
+@@ -33,8 +33,9 @@ def if_mkl(if_true, if_false = []):
+     """
+     return select({
+         "@org_tensorflow//third_party/mkl:build_with_mkl_aarch64": if_true,
+-        "@org_tensorflow//tensorflow:linux_x86_64": if_true,
+-        "@org_tensorflow//tensorflow:windows": if_true,
++        "@org_tensorflow//third_party/mkl:build_with_mkl_lnx_x64": if_true,
++        "@org_tensorflow//third_party/mkl:build_with_mkl_lnx_openmp": if_true,
++        "@org_tensorflow//third_party/mkl:build_with_mkl_windows_openmp": if_true,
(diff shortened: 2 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -219,6 +219,9 @@ jobs:
           $Env:BAZEL_VC = "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC"
           set PreferredToolArchitecture=x64
 
+          # The default is "/arch:AVX", which we don't want.
+          $Env:CC_OPT_FLAGS = "/O2"
```
AdamHillier, line 223: Before adding this change, I did a test release that _didn't_ pass: https://github.com/larq/compute-engine/actions/runs/943850235

But I don't know if that's because this change actually made a difference that makes the builds pass, or if I'm just observing high build-time noise (which anecdotally seems to be the case).

AdamHillier:
@lgeiger we could also consider ressurecting https://github.com/larq/compute-engine/pull/638?

Tombana (approved):

third_party/tensorflow_patches/disable_forced_mkl.patch
```diff
@@ -0,0 +1,27 @@
+diff --git a/third_party/mkl/build_defs.bzl b/third_party/mkl/build_defs.bzl
+index 806b157bad6..6c9f0172b97 100644
+--- a/third_party/mkl/build_defs.bzl
++++ b/third_party/mkl/build_defs.bzl
+@@ -33,8 +33,9 @@ def if_mkl(if_true, if_false = []):
+     """
+     return select({
+         "@org_tensorflow//third_party/mkl:build_with_mkl_aarch64": if_true,
+-        "@org_tensorflow//tensorflow:linux_x86_64": if_true,
+-        "@org_tensorflow//tensorflow:windows": if_true,
```
Tombana, line 10: This patch looks like its also disabling `mkl` for linux, not only for windows. Is that correct? (Anything that speeds up these builds sounds good to me)

AdamHillier: Yes, it disables it for both unless you request it through the relevant Bazel flags.

AdamHillier: Or at least, that's the intent. I haven't actually haven't tested on Linux whether this succeeds — if you run
```
bazel cquery "somepath(:build_pip_pkg, @mkl_dnn_v1//:mkl_dnn)" --define=build_with_mkl=false --define=enable_mkl=false --define=tensorflow_mkldnn_contraction_kernel=0
```
and get something that says "Empty query results" then MKL is successfully disabled.

.github/workflows/release.yml
AdamHillier: I guess this _could_ affect performance of the LCE interpreter on Windows, but:
a) We don't have optimised binary kernels on Windows anyway.
b) AFAIK even without this flag AVX intrinsics will insert AVX instructions, so TFL kernels with optimised Windows kernels should still be okay.

lgeiger: Yeah I think this is a very reasonable thing to do.

lgeiger (approved):

PR #668, opened 2021-06-21
Extend quantiser support so as to accelerate more binary models.

AdamHillier:
## What do these changes do?

This PR extends the LCE converter with patterns that support converting `tf.where`-style binary quantisers. It also adds support for binary input/output of `LceQuantize`/`LceDequantize`.

Note that in https://github.com/larq/larq/pull/677 we are moving towards having the Larq quantisers implemented with `tf.where` instead of the current `tf.sign` implementation. The main change is that this PR adds support for converting `tf.where`-style quantisers.

The second change to add support for boolean input, meaning that a wider variety of binary quantisers will be accelerated by LCE. For example, the following wacky quantiser will now convert successfully:
```python
lambda x: tf.where(tf.logical_or(tf.abs(x) < 0.5, tf.abs(x) > 1.0), 1.0, -1.0)
```
image

With these changes _any_ binary quantiser that can be implemented with `tf.where(boolean_condition, 1, -1)` can be converted into an `LceQuantize` op (and consequently, a subsequent convolution can be converted to a `BConv2D` and thus accelerated). The `boolean_condition` will be implemented with TFL ops, as in the example above, but since the quantisation in general is so quick compared to the binary convolution this doesn't present much of a performance issue.

## How Has This Been Tested?

MLIR test cases have been added; the end2end tests have been extended.

## Benchmark Results

N/A.

## Related issue number

https://github.com/larq/larq/pull/677

### DIFF
--- a/larq_compute_engine/mlir/ir/lce_ops.td
+++ b/larq_compute_engine/mlir/ir/lce_ops.td
@@ -70,11 +70,11 @@ def LQ_QuantizeOp : LQ_Op<"Quantize", [NoSideEffect]> {
   let summary = "Binary quantize operator";
 
   let description = [{
-Converts floating point or integer tensors to binarized bitpacked tensors.
+Converts floating point, integer, or boolean tensors to binarized bitpacked tensors.
   }];
 
   let arguments = (ins
-    TensorOf<[BF16, F16, F32, F64, I32, I64, QI8, QI16]>:$x
+    TensorOf<[BF16, F16, F32, F64, I32, I64, QI8, QI16, I1]>:$x
   );
 
   let results = (outs
@@ -90,7 +90,7 @@ def LQ_DequantizeOp : LQ_Op<"Dequantize", [NoSideEffect]> {
   let summary = "Binary dequantize operator";
 
   let description = [{
-Converts binarized bitpacked tensors to floating point or integer tensors.
+Converts binarized bitpacked tensors to floating point, integer, or boolean tensors.
   }];
 
   let arguments = (ins
@@ -98,7 +98,7 @@ Converts binarized bitpacked tensors to floating point or integer tensors.
   );
 
   let results = (outs
-    TensorOf<[BF16, F16, F32, F64, I32, I64, QI8, QI16]>:$y
+    TensorOf<[BF16, F16, F32, F64, I32, I64, QI8, QI16, I1]>:$y
   );
 
   let hasFolder = 1;
--- a/larq_compute_engine/mlir/transforms/optimize_patterns_common.td
+++ b/larq_compute_engine/mlir/transforms/optimize_patterns_common.td
@@ -11,6 +11,30 @@ def HasOneUse : Constraint<CPred<"$0.hasOneUse()">>;
 
 class ConstantValue<string val> : AttrConstraint<CPred<"IsConstantValue($_self, " # val # ")">>;
 
+def : Pat<(LQ_QuantizeOp
+              (TFL_GreaterEqualOp:$ge_op
+                  $input,
+                  (ConstantOp ConstantValue<"0.0f">))),
+          (LQ_QuantizeOp $input),
+          [(HasOneUse $ge_op)],
+          (addBenefit 150)>;
+
+def : Pat<(LQ_QuantizeOp
+              (TFL_GreaterEqualOp:$ge_op
+                  $input,
+                  $threshold)),
+          (LQ_QuantizeOp
+              (TFL_SubOp $input, $threshold, TFL_AF_None)),
+          [(HasOneUse $ge_op)],
+          (addBenefit 100)>;
+
+def : Pat<(LQ_QuantizeOp
+              (TFL_LessEqualOp:$ge_op $lhs, $rhs)),
+          (LQ_QuantizeOp
+              (TFL_GreaterEqualOp $rhs, $lhs)),
+          [(HasOneUse $ge_op)],
+          (addBenefit 100)>;
+
 // TODO: Check shapes before fusing
 multiclass FuseAddOrSubWithBConv2D<Op binaryOp> {
   def : Pat<(binaryOp
--- a/larq_compute_engine/mlir/transforms/prepare_patterns_common.td
+++ b/larq_compute_engine/mlir/transforms/prepare_patterns_common.td
@@ -4,9 +4,56 @@ include "tensorflow/compiler/mlir/tensorflow/ir/tf_ops.td"
 include "larq_compute_engine/mlir/ir/lce_ops.td"
 include "larq_compute_engine/mlir/transforms/op_removal_patterns.td"
 
+class ConstantValue<string val> : AttrConstraint<CPred<"IsConstantValue($_self, " # val # ")">>;
 
-// This relies on implementation details of larq.math.sign. We should make
-// this more general in the future
+def CreateTFBroadcastToOp : NativeCodeCall<
+    "$_builder.create<TF::BroadcastToOp>("
+        "$0.getLoc(),"
+        "RankedTensorType::get("
+            "$0.getType().cast<RankedTensorType>().getShape(),"
+            "getElementTypeOrSelf($1.getType())),"
+        "$1,"
+        "$2)">;
+
+def CreateTFShapeOp : NativeCodeCall<
+    "$_builder.create<TF::ShapeOp>($0.getLoc(), $1, $2)">;
+
+// Base quantiser patterns that match the `tf.where` implementation of `ste_sign`.
+multiclass QuantDequantPatterns<Op SelectOp> {
+  def : Pat<(SelectOp:$select_op
+                $cond,
+                (ConstantOp ConstantValue<"1.0f">),
+                (ConstantOp ConstantValue<"-1.0f">)),
+            (LQ_DequantizeOp
+                (LQ_QuantizeOp
+                    (CreateTFBroadcastToOp
+                        $select_op,
+                        $cond,
+                        (CreateTFShapeOp
+                            $select_op,
+                            $select_op,
+                            /*use 32bit*/ConstBoolAttrFalse)))),
+            [], (addBenefit 100)>;
+  def : Pat<(SelectOp:$select_op
+                $cond,
+                (ConstantOp ConstantValue<"-1.0f">),
+                (ConstantOp ConstantValue<"1.0f">)),
+            (LQ_DequantizeOp
+                (LQ_QuantizeOp
+                    (CreateTFBroadcastToOp
+                        $select_op,
+                        (TF_LogicalNotOp $cond),
+                        (CreateTFShapeOp
+                            $select_op,
+                            $select_op,
+                            /*use 32bit*/ConstBoolAttrFalse)))),
+            [], (addBenefit 100)>;
+}
+foreach SelectOp = [TF_SelectOp, TF_SelectV2Op]<Op> in
+  defm : QuantDequantPatterns<SelectOp>;
+
+// A fallback for the old version of `ste_sign` that uses a specific `tf.sign`
+// based implementation of `larq.math.sign`.
 def : Pat<(TF_SignOp (TF_AddV2Op (TF_SignOp $arg), $c)),
           (LQ_DequantizeOp (LQ_QuantizeOp $arg)), [], (addBenefit 100)>;
 def : Pat<(TF_SignOp (TF_AddV2Op $c, (TF_SignOp $arg))),
--- a/larq_compute_engine/tflite/kernels/quantization.cc
+++ b/larq_compute_engine/tflite/kernels/quantization.cc
@@ -1,9 +1,12 @@
+#include <type_traits>
+
 #include "larq_compute_engine/core/bitpacking/utils.h"
 #include "ruy/profiler/instrumentation.h"
 #include "tensorflow/lite/c/common.h"
 #include "tensorflow/lite/kernels/internal/cppmath.h"
 #include "tensorflow/lite/kernels/internal/tensor.h"
 #include "tensorflow/lite/kernels/kernel_util.h"
+#include "tensorflow/lite/portable_type_to_tflitetype.h"
 
 using namespace tflite;
 
@@ -20,8 +23,9 @@ TfLiteStatus QuantizePrepare(TfLiteContext* context, TfLiteNode* node) {
   const TfLiteTensor* input = GetInput(context, node, 0);
   TfLiteTensor* output = GetOutput(context, node, 0);
 
-  TF_LITE_ENSURE(context,
-                 input->type == kTfLiteFloat32 || input->type == kTfLiteInt8);
+  TF_LITE_ENSURE(context, input->type == kTfLiteFloat32 ||
+                              input->type == kTfLiteInt8 ||
+                              input->type == kTfLiteBool);
   TF_LITE_ENSURE_EQ(context, output->type, kTfLiteInt32);
 
   int num_dims = NumDimensions(input);
@@ -44,8 +48,9 @@ TfLiteStatus DequantizePrepare(TfLiteContext* context, TfLiteNode* node) {
   TfLiteTensor* output = GetOutput(context, node, 0);
 
   TF_LITE_ENSURE_EQ(context, input->type, kTfLiteInt32);
-  TF_LITE_ENSURE(context,
-                 output->type == kTfLiteFloat32 || output->type == kTfLiteInt8);
+  TF_LITE_ENSURE(context, output->type == kTfLiteFloat32 ||
+                              output->type == kTfLiteInt8 ||
+                              output->type == kTfLiteBool);
 
   int num_dims = NumDimensions(input);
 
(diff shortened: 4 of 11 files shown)


larq_compute_engine/tflite/kernels/quantization.cc
```diff
@@ -80,6 +85,27 @@ TfLiteStatus QuantizeEval(TfLiteContext* context, TfLiteNode* node) {
...
     bitpack_tensor(GetTensorShape(input), GetTensorData<std::int8_t>(input),
                    input->params.zero_point, GetTensorData<TBitpacked>(output));
+  } else if (input->type == kTfLiteBool) {
+    // The strategy here is to interpret the input data as an unsigned integer
+    // (of the same width as the bool type for the target). We then call
+    // bitpacking, with a 'zero point' of 1. This means that the value with all
+    // zero bits will be bitpacked as bit 1, and all other values will be
+    // bitpacked as bit 0. Assuming that `false` is represented by a value with
+    // all zero bits, this gives the correct result of bitpacking `false` as bit
+    // 1 and `true` as bit 0.
```
Tombana, line 95: How do you feel about changing the MLIR pattern to make `true` correspond to float `-1` and `false` to float `+1`, i.e. the boolean is the value of the signbit, `true` when the float value is negative.

Then this kernel code might be simpler, in the sense that the boolean values correspond to the output bits directly.
See also the comment in `prepare_patterns_common.td`.

larq_compute_engine/mlir/transforms/prepare_patterns_common.td
```diff
@@ -4,9 +4,18 @@ include "tensorflow/compiler/mlir/tensorflow/ir/tf_ops.td"
 include "larq_compute_engine/mlir/ir/lce_ops.td"
 include "larq_compute_engine/mlir/transforms/op_removal_patterns.td"
 
+class ConstantValue<string val> : AttrConstraint<CPred<"IsConstantValue($_self, " # val # ")">>;
 
-// This relies on implementation details of larq.math.sign. We should make
-// this more general in the future
+// Base quantiser pattern that matches the `tf.where` implementation of `ste_sign`.
+def : Pat<(TF_SelectV2Op:$select_op
```
Tombana, line 10: This matches `tf.where(x >= 0, 1, -1)` but it does not match `tf.where(x < 0, -1, 1)` which is the same.
I think the converter should accept both forms: it can insert a `not` operation to go into a canonical form.

As far as which of the two would be the better canonical form, see also my comment in the Quantize kernel code: I think the `tf.where(x < 0, -1, 1)` makes more sense from the kernel point of view, because then a `true` is bitpacked to `1` and a `false` is bitpacked to `0`. I don't feel strongly about this, though, but I think both forms should be recognized by the converter.

Tombana (commented):
Looks great!

larq_compute_engine/mlir/transforms/prepare_patterns_common.td
```diff
@@ -4,9 +4,18 @@ include "tensorflow/compiler/mlir/tensorflow/ir/tf_ops.td"
...
+              $cond,
+              (ConstantOp ConstantValue<"1.0f">),
+              (ConstantOp ConstantValue<"-1.0f">)),
```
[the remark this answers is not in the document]
AdamHillier: Yes it's possible to do non-symmetric `(a, b)` and I have a plan to implement that, but for now I thought it would be easier to do that in a following PR. For example, you start to have issues relating to padding values, and stuff like that, that need to be thought about carefully.

larq_compute_engine/tflite/kernels/quantization.cc
AdamHillier: I think that's slightly a question for the Larq API. We could for sure do that, and then require people to write quantisers as e.g. `tf.where(x > 5, minus_ones, ones)` but I feel like instinctively people having the true bits correspond to 1 feels more natural. The fact that we bitpack 1 as 0 and -1 as 1 to me feels like an implementation detail that we shouldn't expose like that, but that's not a belief I hold super strongly.

I'm also not 100% sure what the bit representation of `true` is — I can't find any standard that specifies it. I assume it's the byte `0x01`, but I guess could also be the byte `0xFF`? And if we do know what the bit representation is, can we assume it's the same for all targets? To be fair, at this point in the code I am assuming that `false` is represented by just zero bits, which also isn't specified anywhere, so perhaps the further assumption isn't an issue.

larq_compute_engine/mlir/transforms/prepare_patterns_common.td
AdamHillier: Ah that's a fair point, I hadn't realised that despite having the 'sign canonicalisers' in the optimise pass, there is still a possible issue here, since the `LceQuantise` won't even be inserted. I'll have a think about a solution.

larq_compute_engine/tflite/kernels/quantization.cc
Tombana: I agree that its an implementation detail that shouldn't be exposed in that sense, and I agree that most people will find `true = +1, false = -1` more intuitive. Both forms should still be accepted by the converter (as discussed in the other comment).

I think that indeed the representation of `true` isn't standardized, it can be `1` or `-1=0xffffffff` etc, but `false` is probably always `0` even though that is also not specified.

AdamHillier: Ah okay I think I misunderstood your original comment. I'm completely on board with accepting both forms, and pushing through a TFL `logical_not` op if necessary :)

Then the question is which ought to be the 'canonical' form — since we can push through a `logical_not`, this boils down to whether we want to map `true` -> `0 bit` (which is what I do now) or map `true` -> `1 bit`, and which is easier in terms of kernel implementation.

I *think* that in any case we have to collect sign bits from an unsigned datatype, because as you say we don't know for sure if `true` is represented by `0x01` or `0xFF`. In the former case we have `signed_repr(false) < signed_repr(true)`, but in the latter case the opposite. This means that we would get different results from any operation that relies on collecting sign bits. However, as long as `false` is always represented by `0x00`, `unsigned_repr(false) < unsigned_repr(true)` always holds, and so collecting sign bits (with a zero-point of 1) is guaranteed to work and produce the correct result, no matter what the bit representation of `true` is. The natural format, using this approach, is to map `true` -> `0 bit`, in fact I don't see an easy way to do it for the other way round.

Tombana: You're right, this is the only method that's independent of the `bool` representation :+1:. Lets keep using this form as canonical form then.

larq_compute_engine/mlir/transforms/prepare_patterns_common.td
AdamHillier: Should be covered by c8f25d2.

larq_compute_engine/mlir/transforms/prepare_patterns_common.td
```diff
@@ -4,9 +4,24 @@ include "tensorflow/compiler/mlir/tensorflow/ir/tf_ops.td"
...
+          (LQ_DequantizeOp (LQ_QuantizeOp $cond)),
+          [], (addBenefit 100)>;
+def : Pat<(TF_SelectV2Op:$select_op
+              $cond,
+              (ConstantOp ConstantValue<"-1.0f">),
+              (ConstantOp ConstantValue<"1.0f">)),
+          (LQ_DequantizeOp (LQ_QuantizeOp (TF_LogicalNotOp $cond))),
```
Tombana, line 20: Do you know if this logical not is folded with the `TFL_LessEqualOp` and `TFL_GreaterEqualOp` ? That is, will this also further trigger the optimization patterns?

AdamHillier: It seems to do so, yes :)

This is what I get when converting a model using the quantiser `lambda x: tf.where(x < 0.5, -1.0, 1.0)`:

image

Tombana (approved):
LGTM.
Just a note to ourselves: when merging this into our private fork it will probably not cause any automatic merge conflicts, but we'll need to update the micro quantization kernels (should be fairly trivial, just by copying the changes from the public quantization kernel).

lgeiger (commented):
Nicely done!

It took me a while to wrap my head around boolean input for `lq.quantize`, but I think it makes a lot of sense and is much easier to maintain than passing an additional threshold attribute to the op or doing something similar.

I just have some additional comments regarding broadcasting and the matching of `tf.where`.

larq_compute_engine/mlir/transforms/prepare_patterns_common.td
```diff
@@ -4,9 +4,24 @@ include "tensorflow/compiler/mlir/tensorflow/ir/tf_ops.td"
 include "larq_compute_engine/mlir/ir/lce_ops.td"
 include "larq_compute_engine/mlir/transforms/op_removal_patterns.td"
 
+class ConstantValue<string val> : AttrConstraint<CPred<"IsConstantValue($_self, " # val # ")">>;
 
-// This relies on implementation details of larq.math.sign. We should make
-// this more general in the future
+// Base quantiser patterns that match the `tf.where` implementation of `ste_sign`.
+def : Pat<(TF_SelectV2Op:$select_op
```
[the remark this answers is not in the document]
AdamHillier: I agree we should probably also match the v1 op :+1:

With respect to shapes, the pattern correctly works when the lhs or rhs is lower rank than the input (there's an MLIR test for this). I see what you mean though, that if the condition is lower rank than the lhs or rhs then we need to broadcast it. Do you think it would be easiest to insert a broadcast op then?

larq_compute_engine/mlir/ir/lce_ops.td
```diff
@@ -70,11 +70,11 @@ def LQ_QuantizeOp : LQ_Op<"Quantize", [NoSideEffect]> {
   let summary = "Binary quantize operator";
 
   let description = [{
-Converts floating point or integer tensors to binarized bitpacked tensors.
+Converts floating point, integer, or boolean tensors to binarized bitpacked tensors.
```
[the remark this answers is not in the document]
AdamHillier: I kind of think it's an implementation detail in the same way that collecting the sign bits of numeric input is an implementation detail. If we do document it, I think we should do so [in our API docs](https://docs.larq.dev/compute-engine/api/operators/), which I don't think are auto-generated?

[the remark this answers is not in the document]
lgeiger: Fair point, I missed that we fully control the insertion of this op in the pattern.

Correct, the API docs are just written manually.

larq_compute_engine/mlir/transforms/prepare_patterns_common.td
```diff
@@ -4,9 +4,56 @@ include "tensorflow/compiler/mlir/tensorflow/ir/tf_ops.td"
...
-// This relies on implementation details of larq.math.sign. We should make
-// this more general in the future
+def CreateTFBroadcastToOp : NativeCodeCall<
+    "$_builder.create<TF::BroadcastToOp>("
+        "$0.getLoc(),"
+        "RankedTensorType::get("
+            "$0.getType().cast<RankedTensorType>().getShape(),"
+            "getElementTypeOrSelf($1.getType())),"
+        "$1,"
+        "$2)">;
```
AdamHillier, line 16: @lgeiger after running into the hell of trying to insert nested ops with tablegen, I eventually resorted to this rather contrived native code call. This took ages to get right, but it does now work correctly with broadcasting any of the three select op arguments (verified with new MLIR test cases) :tada:

Tombana: What does this broadcasting allow us to match against that we would not get without it?

AdamHillier: Basically the `tf.where` op supports broadcasting all three of it's inputs (the condition, lhs, and rhs). So this would allow, for example, `tf.where(<tensor shape=(5, 32)>, <ones shape=(5, 5, 32)>, <-ones shape=(5, 5, 32)>)`. That of course would not really make much sense from a ML perspective, but since `tf.where` supports such constructions I don't think it hurts to support it.

larq_compute_engine/mlir/transforms/prepare_patterns_common.td
```diff
@@ -4,9 +4,56 @@ include "tensorflow/compiler/mlir/tensorflow/ir/tf_ops.td"
...
+                        $select_op,
+                        (TF_LogicalNotOp $cond),
+                        (CreateTFShapeOp
+                            $select_op,
+                            $cond,
+                            /*use 32bit*/ConstBoolAttrFalse)))),
+            [], (addBenefit 100)>;
+}
+foreach SelectOp = [TF_SelectOp, TF_SelectV2Op] in
+  defm : QuantDequantPatterns<!cast<Op>(SelectOp)>;
```
lgeiger, line 53: I think we can just do this to simplify the casting (at least that's how TFL handles it).
```suggestion
foreach selectOp = [TF_SelectOp, TF_SelectV2Op]<Op> in
  defm : QuantDequantPatterns<selectOp>;
```

larq_compute_engine/mlir/transforms/prepare_patterns_common.td
```diff
@@ -4,9 +4,56 @@ include "tensorflow/compiler/mlir/tensorflow/ir/tf_ops.td"
...
+    "$_builder.create<TF::BroadcastToOp>("
+        "$0.getLoc(),"
+        "RankedTensorType::get("
+            "$0.getType().cast<RankedTensorType>().getShape(),"
+            "getElementTypeOrSelf($1.getType())),"
+        "$1,"
+        "$2)">;
+
+def CreateTFShapeOp : NativeCodeCall<
+    "$_builder.create<TF::ShapeOp>($0.getLoc(), $1, $2)">;
```
lgeiger, line 19: To make sure I understand this correctly: Why does the `tf.shape` op take two input arguments instead of just one? Or am I missing something here?

larq_compute_engine/mlir/transforms/prepare_patterns_common.td
```diff
@@ -4,9 +4,56 @@ include "tensorflow/compiler/mlir/tensorflow/ir/tf_ops.td"
...
+                (ConstantOp ConstantValue<"1.0f">),
+                (ConstantOp ConstantValue<"-1.0f">)),
+            (LQ_DequantizeOp
+                (LQ_QuantizeOp
+                    (CreateTFBroadcastToOp
+                        $select_op,
+                        $cond,
+                        (CreateTFShapeOp
+                            $select_op,
+                            $cond,
```
lgeiger, line 34: Does this extract the shape from `$select_op` or from `$cond`? I think the right behaviour would be to use the shape from `$select_op`.

larq_compute_engine/mlir/transforms/optimize_patterns_common.td
```diff
@@ -11,6 +11,30 @@ def HasOneUse : Constraint<CPred<"$0.hasOneUse()">>;
 
 class ConstantValue<string val> : AttrConstraint<CPred<"IsConstantValue($_self, " # val # ")">>;
 
+def : Pat<(LQ_QuantizeOp
+              (TFL_GreaterEqualOp:$ge_op
+                  $input,
+                  (ConstantOp:$threshold ConstantValue<"0.0f">))),
+          (LQ_QuantizeOp $input),
+          [(HasOneUse $ge_op), (HasOneUse $threshold)],
```
lgeiger, line 19: I think since this is a constant value we don't care about the usage of it, or do we?
```suggestion
                  (ConstantOp ConstantValue<"0.0f">))),
          (LQ_QuantizeOp $input),
          [(HasOneUse $ge_op)],
```

lgeiger (commented):
This is great, thank you so much for figuring this out!

I just have a few questions to understand the code, other than that this looks great to me 🚀

larq_compute_engine/mlir/transforms/prepare_patterns_common.td
AdamHillier: This is the definition as [used in TF here](https://github.com/tensorflow/tensorflow/blob/25a9edf48c4a26ee3b78eaec72cec6281c44ead7/tensorflow/compiler/mlir/tensorflow/transforms/lower_tf.td#L43-L44).

It takes three args — `$0` is used to get the correct location, `$1` is the input (thing we're getting the shape of), and `$2` is an attribute for the shape op that is annotated "use 32bit" (see e.g. [TF usage here](https://github.com/tensorflow/tensorflow/blob/25a9edf48c4a26ee3b78eaec72cec6281c44ead7/tensorflow/compiler/mlir/tensorflow/transforms/lower_tf.td#L419-L420)).

lgeiger: Ahh I see, thanks that makes sense.

larq_compute_engine/mlir/transforms/prepare_patterns_common.td
AdamHillier: Ah I think you're right there, that's a mistake. For some reason the explicit broadcasting tests still pass, so it must still be working, but yes, will fix.

lgeiger: Not sure why it's working though, but I think it's good to fix. We can probably also reduce the number of arguments from `CreateTFShapeOp` then.

AdamHillier: I haven't reduced the number of args, cause I think it's slightly nicer to use the identical name/construction as in TF, but don't feel too strongly about that.

larq_compute_engine/mlir/transforms/prepare_patterns_common.td
lgeiger, line 19: How do you feel about sharing the first argument?
```suggestion
    "$_builder.create<TF::ShapeOp>($0.getLoc(), $0, $1)">;
```

Do you think it would make sense to also fix the `use 32bit` argument here? Or would this be more complicated than just passing it as a input in the pattern?

larq_compute_engine/mlir/transforms/prepare_patterns_common.td
lgeiger: OK, then let's keep it like this. I just thought if we would also fix the the use 32-bit argument inside the utility function we the signature would match the OP as defined in tablegen and wasn't aware that this utility is borrowed from TF.

lgeiger: OK, then let's keep it like this. I just thought if we would also fix the the use 32-bit argument inside the utility function we the signature would match the OP as defined in tablegen and wasn't aware that this utility is borrowed from TF.

lgeiger (approved):

PR #670, opened 2021-06-24
Remove manylinux container from CI

Tombana:
## What do these changes do?
In our private fork we had problems where the manylinux container is too outdated for our unittests. Combined with the fact that it doesn't actually test anything (the tested code does not really overlap with the converter, which is the only thing that we use the manylinux container for), it makes more sense to remove the container from CI.

## How Has This Been Tested?
CI


.github/workflows/unittests.yml
```diff
@@ -10,26 +10,15 @@ jobs:
   TFLite:
     runs-on: ubuntu-latest
     if: "!contains(github.event.head_commit.message, 'ci-skip')"
-    container:
-      image: tensorflow/tensorflow:2.2.0-custom-op-ubuntu16
-
     steps:
       - uses: actions/checkout@v1
         with:
           submodules: true
```
lgeiger, line 16: I think this should now also allow us to use the newest version of the checkout action:
```suggestion
      - uses: actions/checkout@v2
        with:
          submodules: true
          fetch-depth: 0
```

lgeiger (approved):

.github/workflows/unittests.yml
```diff
@@ -10,26 +10,20 @@ jobs:
...
         run: ./configure.py
         shell: bash
-        env:
-          MANYLINUX2010: 1
-        # CC unit tests needs to run after the configure is finished
-        # since the CC builds rely on compiler options generated by configure.
-      - name: Run TFLite Ops and Core CC Unit Tests
-        run: ./bazelisk test larq_compute_engine/tests:cc_tests --copt=-O2 --distinct_host_configuration=false --test_output=all
+      - name: Install pip dependencies
+        run: pip install numpy --no-cache-dir
```
lgeiger, line 24: I think we also used to also need `six` for TF to build properly. Not sure if this is still the case.
```suggestion
        run: pip install numpy six --no-cache-dir
```

lgeiger (approved):

.github/workflows/unittests.yml
Tombana: Thanks for the tip, lets see what CI does.

AdamHillier (approved):

lgeiger (approved):
This also decreased the build time from ~18min to around ~13min 🚀

PR #675, opened 2021-07-23
Split Python interpreter in two classes

Tombana:
## What do these changes do?
This splits the `Interpreter` python class in `InterpreterBase` (python-only) and `Interpreter` (including the C++ module). The base class is also used by our private interpreters, and this split avoids their dependency on the full LCE interpreter.

## How Has This Been Tested?
CI covers this.

### DIFF
--- a/larq_compute_engine/tflite/python/interpreter.py
+++ b/larq_compute_engine/tflite/python/interpreter.py
@@ -1,4 +1,3 @@
-from larq_compute_engine.tflite.python import interpreter_wrapper_lite
 from larq_compute_engine.tflite.python.interpreter_base import InterpreterBase
 
 __all__ = ["Interpreter"]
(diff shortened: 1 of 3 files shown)


CNugteren (approved):

larq_compute_engine/tflite/python/interpreter.py
```diff
@@ -46,16 +19,12 @@ class Interpreter:
...
     # Arguments
         flatbuffer_model: A serialized Larq Compute Engine model in the flatbuffer format.
         num_threads: The number of threads used by the interpreter.
         use_reference_bconv: When True, uses the reference implementation of LceBconv2d.
-
-    # Attributes
-        input_types: Returns a list of input types.
-        input_shapes: Returns a list of input shapes.
-        output_types: Returns a list of output types.
-        output_shapes: Returns a list of output shapes.
```
lgeiger, line 58: Can we add the Attributes, here again since they appear in [our docs](https://docs.larq.dev/compute-engine/api/python/#interpreter)?

Tombana: Won't the docs pick it up from the base class?

lgeiger: Unfortunately I don't think so, it just parses the markdown.

Tombana: I've moved it: https://github.com/larq/compute-engine/pull/677

PR #694, opened 2021-12-13
Add td_library target to be able to include lce_ops.td

panickal-xmos:
There are two changes :
- Adds a new td_library target to be able to include lce_ops.td externally. In current MLIR and Tensorflow upstream, all td files are defined as part of td_library targets. This will probably need to be done for all td files when compute-engine is upgraded to Tensorflow 2.7
- Adds a missing dependency to @llvm-project//mlir:QuantOps. Missing header dependencies are enforced when building with later bazel versions.

### DIFF
--- a/larq_compute_engine/mlir/BUILD
+++ b/larq_compute_engine/mlir/BUILD
@@ -12,8 +12,8 @@ td_library(
         "ir/lce_ops.td",
     ],
     deps = [
-        "@org_tensorflow//tensorflow/compiler/mlir/lite/quantization:quantization_td_files",
         "@llvm-project//mlir:SideEffectTdFiles",
+        "@org_tensorflow//tensorflow/compiler/mlir/lite/quantization:quantization_td_files",
     ],
 )
 
@@ -29,7 +29,7 @@ gentbl(
     td_file = "ir/lce_ops.td",
     td_includes = ["external/org_tensorflow"],
     td_srcs = [
-        ":lce_ops_td_file"
+        ":lce_ops_td_file",
     ],
 )
 
--- a/larq_compute_engine/mlir/BUILD
+++ b/larq_compute_engine/mlir/BUILD
@@ -11,8 +11,10 @@ td_library(
     srcs = [
         "ir/lce_ops.td",
     ],
-    includes = ["include"],
-    deps = [],
+    deps = [
+        "@org_tensorflow//tensorflow/compiler/mlir/lite/quantization:quantization_td_files",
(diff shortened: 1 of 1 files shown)


panickal-xmos:
@lgeiger The MLIR test fail seems to be due to credentials. Can I do something to fix this?

lgeiger:
> The MLIR test fail seems to be due to credentials. Can I do something to fix this?

Sorry about that, it looks like there was a bug in the latest `gcloud` action release that broke our CI for PRs from forks. This is now fixed on `main`. @panickal-xmos can you rebase your PR onto `main` to fix the CI failure here?

larq_compute_engine/mlir/BUILD
```diff
@@ -1,11 +1,20 @@
...
     default_visibility = ["//visibility:public"],
     licenses = ["notice"],  # Apache 2.0
 )
 
+td_library(
+    name = "lce_ops_td_file",
+    srcs = [
+        "ir/lce_ops.td",
+    ],
+    includes = ["include"],
```
lgeiger, line 14: What is this option for?

larq_compute_engine/mlir/BUILD
```diff
@@ -1,11 +1,20 @@
...
+    deps = [],
```
lgeiger, line 15: I think this should include the following dependencies to be correct:
```
"@org_tensorflow//tensorflow/compiler/mlir/lite/quantization:quantization_td_files",
"@llvm-project//mlir:SideEffectTdFiles",
```

lgeiger (changes requested):
Thank you very much for the PR.

When looking at the [upstream build file](https://github.com/tensorflow/tensorflow/blob/1a15d127ae43a072fde36a2fc35f96a24b80aa75/tensorflow/compiler/mlir/lite/BUILD#L37) is seems like build targets depend on the `td_library` instead of the `*.td` files directly. Could you change the dependency definitions in this file from `ir/lce_ops.td` to the new `lce_ops_tf_file` to make it consistent and to make sure that this build target is also tested properly on CI.

larq_compute_engine/mlir/BUILD
panickal-xmos: Removed it

larq_compute_engine/mlir/BUILD
panickal-xmos: Done

lgeiger (approved):
Looks great! Thanks for the PR 🎉

PR #698, opened 2021-12-16
Initial commit for tfl custom op translation

panickal-xmos:
Add a pass to translate from TFL CustomOp to the relevant Larq ops


larq_compute_engine/mlir/tests/lce_ops_options_test.cc
```diff
@@ -60,9 +60,9 @@ TEST(LCEOpsSerializationTest, BConv2dTest) {
   ASSERT_EQ(m["stride_height"].AsInt32(), 1);
   ASSERT_EQ(m["stride_width"].AsInt32(), 2);
   ASSERT_EQ(m["pad_values"].AsInt32(), 1);
-  ASSERT_EQ((ActivationFunctionType)m["fused_activation_function"].AsInt32(),
-            ActivationFunctionType_RELU);
-  ASSERT_EQ((Padding)m["padding"].AsInt32(), Padding_SAME);
+  ASSERT_EQ((::ActivationFunctionType)m["fused_activation_function"].AsInt32(),
+            ::ActivationFunctionType::RELU);
+  ASSERT_EQ((::Padding)m["padding"].AsInt32(), ::Padding::SAME);
```
lgeiger, line 65: Would this, require changes to the kernel implementation, or are these enums identical?
@Tombana Do you see any problems with this? I guess since the test suite passes, it is fine?

larq_compute_engine/mlir/transforms/translate_tflite.cc
```diff
@@ -0,0 +1,84 @@
...
+
+namespace mlir {
+namespace TFL {
+
+namespace {
+
+struct TranslateToLCE : public PassWrapper<TranslateToLCE, FunctionPass> {
+  void getDependentDialects(DialectRegistry& registry) const override {
+    registry.insert<mlir::TFL::TensorFlowLiteDialect>();
+    registry.insert<mlir::lq::LarqDialect>();
```
lgeiger, line 16:
```suggestion
    registry.insert<mlir::TFL::TensorFlowLiteDialect, mlir::lq::LarqDialect>();
```

larq_compute_engine/mlir/transforms/translate_tflite.cc
```diff
@@ -0,0 +1,84 @@
...
+          custom_op, custom_op->getResultTypes(), custom_op->getOperands());
+    } else if (custom_op.custom_code() == "LceDequantize") {
+      rewriter.replaceOpWithNewOp<lq::DequantizeOp>(
+          custom_op, custom_op->getResultTypes(), custom_op->getOperands());
+    } else if (custom_op.custom_code() == "LceBMaxPool2d") {
+      auto map =
+          flexbuffers::GetRoot((uint8_t*)stringData.data(), stringData.size())
+              .AsMap();
+      rewriter.replaceOpWithNewOp<lq::BMaxPool2dOp>(
+          custom_op, custom_op->getResultTypes(), custom_op->getOperands()[0],
```
lgeiger, line 40: I think this can be simplified a tiny bit:
```suggestion
          custom_op, custom_op->getResultTypes(), custom_op->getOperand(0),
```

larq_compute_engine/mlir/transforms/translate_tflite.cc
```diff
@@ -0,0 +1,84 @@
...
+          map["stride_width"].AsInt32(), map["stride_height"].AsInt32(),
+          map["filter_width"].AsInt32(), map["filter_height"].AsInt32());
+    } else if (custom_op.custom_code() == "LceBconv2d") {
+      auto map =
+          flexbuffers::GetRoot((uint8_t*)stringData.data(), stringData.size())
+              .AsMap();
+      rewriter.replaceOpWithNewOp<lq::Bconv2dOp>(
+          custom_op, custom_op->getResultTypes(), custom_op->getOperands()[0],
+          custom_op->getOperands()[1], custom_op->getOperands()[2],
+          custom_op->getOperands()[3], custom_op->getOperands()[4],
```
lgeiger, line 51:
```suggestion
          custom_op, custom_op->getResultTypes(), custom_op->getOperand(0),
          custom_op->getOperand(1), custom_op->getOperand(2),
          custom_op->getOperand(3), custom_op->getOperand(4),
```

larq_compute_engine/mlir/tf_tfl_passes.cc
```diff
@@ -211,6 +211,8 @@ void AddTFToLCETFLConversionPasses(
       mlir::TFL::CreateInsertCallOnceOpFromSessionInitializerPass());
 
   pass_manager->addPass(mlir::TFL::CreateLegalizeLCEPass());
+  pass_manager->addPass(mlir::TFL::CreateTranslateToLCEPass());
+  pass_manager->addPass(mlir::TFL::CreateLegalizeLCEPass());
```
lgeiger, line 215: Was this leftover from testing? I don't think we should do the roundtrip by default on every conversion, or am I missing something?

lgeiger (changes requested):
I think this is a very clean solution, thank you very much for the PR!

I just have a few very minor comments below, other than that the PR looks great from my side.

larq_compute_engine/mlir/tests/lce_ops_options_test.cc
Tombana: I guess the enums must be identical, but I don't fully see how that is guaranteed. But if this would ever be broken then this would definitely be caught by tests, so I think this is fine :+1:

larq_compute_engine/mlir/tf_tfl_passes.cc
panickal-xmos: I had added it to ensure that any changes in creating the custom ops will be caught since the tests will still pass even if the custom op creation pass is modified. Now thinking about it, it's better to add an MLIR test that runs both passes together. I'll remove this.

larq_compute_engine/mlir/tests/lce_ops_options_test.cc
panickal-xmos: The enums haven't been modified, but had to add this since they were clashing with similarly named ones defined by TensorFlow.

Tombana (approved):

larq_compute_engine/mlir/tests/lce_ops_options_test.cc
lgeiger: Sounds good, thanks for the explanation.

larq_compute_engine/mlir/tests/translate-tflite.mlir
```diff
@@ -0,0 +1,37 @@
+// RUN: lce-tf-opt %s -tfl-legalize-lce -lce-translate-tfl -verify-diagnostics | FileCheck %s
```
lgeiger, line 1: Since we now test the full round-trip in this test already, how do you feel about consolidating this with the [`legalize-lce.mlir`](https://github.com/larq/compute-engine/blob/main/larq_compute_engine/mlir/tests/legalize-lce.mlir) tests?

This can be done via the `--check-prefix` option in the `FileCheck` which allows to run different passes on the same IR. Similar to how it is done [here](https://github.com/tensorflow/tensorflow/blob/f42ecd1bdcee0aea8a2afabbe699679dc07b8b9f/tensorflow/compiler/mlir/lite/tests/legalize-tf-while.mlir#L1-L3).

panickal-xmos: That's a good idea! Let me make the change.

Tombana (approved):

lgeiger (approved):
Great! 🚀

PR #699, opened 2022-01-07
Converter internals unification

CNugteren:
## What do these changes do?
This factors out some common code from `saved_model_tfl_flatbuffer.cc` and `graphdef_tfl_flatbuffer.cc` into a new file. This already leads to some code de-duplication, but is mainly motivated to make future additions of similar functionality easier (e.g. a tflite-mlir-tflite converter).

## How Has This Been Tested?
CI should cover this I hope.

## Related issue number
N/A

### DIFF
--- a/larq_compute_engine/mlir/BUILD
+++ b/larq_compute_engine/mlir/BUILD
@@ -388,6 +388,21 @@ cc_library(
     ],
 )
 
+cc_library(
+    name = "python_common",
+    srcs = ["python/common.cc"],
+    hdrs = ["python/common.h"],
+    deps = [
+        ":lce_tfl_passes",
+        ":tf_to_tfl_flatbuffer",
+        "@org_tensorflow//tensorflow/compiler/mlir/lite:tensorflow_lite",
+        "@org_tensorflow//tensorflow/compiler/mlir/lite:tf_to_tfl_flatbuffer",
+        "@org_tensorflow//tensorflow/compiler/mlir/lite/python:tf_tfl_flatbuffer_helpers",
+        "@org_tensorflow//tensorflow/core:ops",
+        "@pybind11",
+    ],
+)
+
 tf_cc_binary(
     name = "lce-tf-opt",
     srcs = ["lce_mlir_opt.cc"],
@@ -407,6 +422,7 @@ pybind_extension(
     module_name = "tf_tfl_flatbuffer",
     deps = [
         ":lce_tfl_passes",
+        ":python_common",
         ":tf_to_tfl_flatbuffer",
         "@org_tensorflow//tensorflow/cc/saved_model:loader",
         "@org_tensorflow//tensorflow/compiler/mlir/lite:tensorflow_lite",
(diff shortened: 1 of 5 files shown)


larq_compute_engine/mlir/python/common.cc
```diff
@@ -0,0 +1,141 @@
...
+  if (target_str == "arm") {
+    return LCETarget::ARM;
+  } else if (target_str == "xcore") {
+    return LCETarget::XCORE;
+  } else {
+    throw std::runtime_error("Invalid target.");
+  }
+}
+
+Status GetNumInputs(mlir::OwningModuleRef* module, int* num_inputs) {
```
CNugteren, line 44: This function was copied 1-to-1 from `saved_model_tfl_flatbuffer.cc`.

larq_compute_engine/mlir/python/common.cc
```diff
@@ -0,0 +1,141 @@
...
+  if (!input_attr) {
+    return errors::InvalidArgument("no inputs attribute found");
+  }
+  auto input_names = input_attr.cast<mlir::StringAttr>().getValue();
+  input_names.split(function_input_names, ",");
+  *num_inputs = function_input_names.size();
+  return Status::OK();
+}
+
+pybind11::bytes ConvertMLIRModuleToTFLiteFlatBuffer(
```
CNugteren, line 77: This function body was copied from `saved_model_tfl_flatbuffer.cc` and then adjusted in the following ways:
1) The `should_quantize` option was added to make it work with `graphdef_tfl_flatbuffer.cc` as well.
2) The `mark_as_post_training_quant` option was added to make it work with `graphdef_tfl_flatbuffer.cc` as well.

larq_compute_engine/mlir/BUILD
```diff
@@ -407,6 +422,7 @@ pybind_extension(
     module_name = "tf_tfl_flatbuffer",
     deps = [
         ":lce_tfl_passes",
+        ":python_common",
         ":tf_to_tfl_flatbuffer",
         "@org_tensorflow//tensorflow/cc/saved_model:loader",
         "@org_tensorflow//tensorflow/compiler/mlir/lite:tensorflow_lite",
```
Tombana, line 428: I saw you removed a lot of includes from `graphdef_tfl_flatbuffer.cc` and `saved_model_tfl_flatbuffer.cc`, maybe they could also be removed from the `deps` here? They'll still be used through the `python_common` dependency, but it might be a bit cleaner.

Tombana (approved):

larq_compute_engine/mlir/BUILD
CNugteren: Hmm, it is a bit tricky. Technically it works already when I do this:
```bazel
    deps = [
        ":python_common",
    ],
```
But I think despite that I removed a lot of includes, all dependencies listed here are still direct dependencies of `graphdef_tfl_flatbuffer.cc` and `saved_model_tfl_flatbuffer.cc`.

Tombana: Allright. Then it's good to go from my side :+1:

PR #700, opened 2022-01-12
Simplify Python Linux wheel release scripts

CNugteren:
## What do these changes do?
This simplifies the release scripts for the larq-compute-engine Python wheel converter for Linux. This is based on [the latest simplified method from tf-addons release system](https://github.com/tensorflow/addons/blob/master/tools/docker/build_wheel.Dockerfile) to build the larq-compute-engine wheel. The previous scripts resulted in issues, most likely due to running out of memory.

The implications are that this drops support for Python 3.6, since there is no corresponding Docker image available. It also relies on the 'latest' tag, which could result in things suddenly not working anymore as 'latest' changes due to a TensorFlow update. However, the advantage is that we always use the latest images.

This PR requires updating [the docs](https://github.com/larq/docs/blob/master/docs/compute-engine/build/docker.md) separately.

## How Has This Been Tested?
We'll need to do a test release on this branch before merging in. A test run is currently in progress here:
https://github.com/larq/compute-engine/actions/runs/1686835008

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -133,7 +133,7 @@ jobs:
     runs-on: macos-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9]
       fail-fast: false
     steps:
       - uses: actions/checkout@v2
@@ -232,7 +232,7 @@ jobs:
     runs-on: windows-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9]
       fail-fast: false
     steps:
       - name: Configure Pagefile
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -199,7 +199,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


Tombana (approved):

.github/workflows/release.yml
```diff
@@ -199,7 +199,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9]
```
lgeiger, line 202: I'd be in favour of also dropping the Python 3.6 wheels for macOS and Windows and updating the the required minimum Python version:
https://github.com/larq/compute-engine/blob/3826efcc984136e3f7814f1ab65fd6d4b4636389/setup.py#L33

Python 3.6 has reached its end of life in the end of Dec 2021, so I think it makes sense to completely drop it. @larq/compute-engine What do you think?

lgeiger (approved):

.github/workflows/release.yml
CNugteren: Done, 3.6 is now no longer supported.

CNugteren:
I'm merging this one in. The relevant builds have passed in a test release (https://github.com/larq/compute-engine/actions/runs/1686835008), although some others timed out due to a GitHub Actions time limit of 6 hours.

PR #712, opened 2022-02-15
Fuse padding layers into (Depthwise) Convolutions

Tombana:
## What do these changes do?
The MobileNet models in `tf.keras.applications` have unneeded explicit padding layers. This PR introduces an MLIR pass that fuses such layers into (Depthwise) Convolutions.

## How Has This Been Tested?
An MLIR FileCheck test has been added.

### DIFF
--- a/larq_compute_engine/mlir/tests/fuse_padding.mlir
+++ b/larq_compute_engine/mlir/tests/fuse_padding.mlir
@@ -27,6 +27,19 @@ func @fuse_padv2_into_conv_valid(%arg0: tensor<1x64x64x8xf32>) -> tensor<1x64x64
   // CHECK-NEXT: return %0 : tensor<1x64x64x16xf32>
 }
 
+// CHECK-LABEL: @fuse_pad_into_dwconv_valid
+func @fuse_pad_into_dwconv_valid(%arg0: tensor<1x64x64x16xf32>) -> tensor<1x64x64x16xf32> {
+  %cst0 = constant dense<[[0, 0], [1, 1], [1, 1], [0, 0]]> : tensor<4x2xi32>
+  %cst1 = constant dense<1.0> : tensor<1x3x3x16xf32>
+  %cst2 = constant dense<1.0> : tensor<16xf32>
+  %0 = "tfl.pad"(%arg0, %cst0) : (tensor<1x64x64x16xf32>, tensor<4x2xi32>) -> tensor<1x66x66x16xf32>
+  %1 = "tfl.depthwise_conv_2d"(%0, %cst1, %cst2) {depth_multiplier = 1 : i32, dilation_h_factor = 1 : i32, dilation_w_factor = 1 : i32, fused_activation_function = "NONE", padding = "VALID", stride_h = 1 : i32, stride_w = 1 : i32} : (tensor<1x66x66x16xf32>, tensor<1x3x3x16xf32>, tensor<16xf32>) -> tensor<1x64x64x16xf32>
+  return %1 : tensor<1x64x64x16xf32>
+
+  // CHECK: %0 = "tfl.depthwise_conv_2d"(%arg0, %cst, %cst_0) {depth_multiplier = 1 : i32, dilation_h_factor = 1 : i32, dilation_w_factor = 1 : i32, fused_activation_function = "NONE", padding = "SAME", stride_h = 1 : i32, stride_w = 1 : i32} : (tensor<1x64x64x16xf32>, tensor<1x3x3x16xf32>, tensor<16xf32>) -> tensor<1x64x64x16xf32>
+  // CHECK-NEXT: return %0 : tensor<1x64x64x16xf32>
+}
+
 // CHECK-LABEL: @do_not_fuse_padv2_into_conv_wrong_pad_value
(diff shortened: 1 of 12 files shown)


larq_compute_engine/mlir/transforms/prepare_tf.cc
```diff
@@ -103,27 +103,17 @@ bool IsSamePadding(Attribute paddings_attr, Value input, Value output,
...
-         paddings.getValue<int>({3, 0}) == 0 &&
-         paddings.getValue<int>({3, 1}) == 0;
+  // Check that there is no padding in the batch and channel dimensions
+  // Functions defined in `padding.h`
+  return IsNoPadding(paddings, 0) &&
+         IsSamePadding1D(paddings, 1, input_shape[1], output_shape[1],
+                         stride_height) &&
+         IsSamePadding1D(paddings, 2, input_shape[2], output_shape[2],
+                         stride_width) &&
+         IsNoPadding(paddings, 3);
```
CNugteren, line 116: Maybe this was already like this, but does this work with non 2D data? Here you assume the channel dimension is dimension 3, but what if there is only one spatial dimension? I guess then there is no Conv2D and this function doesn't trigger, is that it?

larq_compute_engine/mlir/transforms/fuse_padding.cc
```diff
@@ -0,0 +1,70 @@
...
+  auto output_shape = output.getType().cast<RankedTensorType>().getShape();
+
+  if (!strides_attr.isa<IntegerAttr>()) return false;
+  const int stride = strides_attr.cast<IntegerAttr>().getInt();
+
+  // Check that there is no padding in the batch and channel dimensions
+  return IsNoPadding(paddings, 0) &&
+         IsSamePadding1D(paddings, dimension, input_shape[dimension],
+                         output_shape[dimension], stride) &&
+         IsNoPadding(paddings, 3);
```
CNugteren, line 37: A bit weird perhaps of a function `IsSamePaddingPartial` with a `dimension` argument to also check things in dimension 0 and 3. Is it possible to remove these `IsNoPadding` checks here and call them from the place where this function is called, or is that tricky because of MLIR?

larq_compute_engine/mlir/tests/fuse_padding.mlir
```diff
@@ -0,0 +1,54 @@
+// RUN: lce-tf-opt %s -tfl-fuse-padding -verify-diagnostics | FileCheck %s
+
+// CHECK-LABEL: @fuse_pad_into_conv_valid
+func @fuse_pad_into_conv_valid(%arg0: tensor<1x64x64x8xf32>) -> tensor<1x64x64x16xf32> {
+  %cst0 = constant dense<[[0, 0], [1, 1], [1, 1], [0, 0]]> : tensor<4x2xi32>
+  %cst1 = constant dense<1.0> : tensor<16x3x3x8xf32>
+  %cst2 = constant dense<1.0> : tensor<16xf32>
+  %0 = "tfl.pad"(%arg0, %cst0) : (tensor<1x64x64x8xf32>, tensor<4x2xi32>) -> tensor<1x66x66x8xf32>
+  %1 = "tfl.conv_2d"(%0, %cst1, %cst2) {dilation_h_factor = 1 : i32, dilation_w_factor = 1 : i32, fused_activation_function = "NONE", padding = "VALID", stride_h = 1 : i32, stride_w = 1 : i32} : (tensor<1x66x66x8xf32>, tensor<16x3x3x8xf32>, tensor<16xf32>) -> tensor<1x64x64x16xf32>
```
CNugteren, line 9: Perhaps good to also add a similar test but then for depthwise-conv?

larq_compute_engine/mlir/transforms/prepare_tf.cc
Tombana: This was indeed already like this, and it should indeed work because the function is only ever called if its followed by a (Depthwise)Conv2D. However I think it's a good idea to add some extra checks for this, I will do that.

Tombana: I've added some extra checks in [f28fe61](https://github.com/larq/compute-engine/pull/712/commits/f28fe619410e4b8efdc8045e58b059bf87d0028f)

larq_compute_engine/mlir/transforms/fuse_padding.cc
Tombana: I agree that this was weird. I've changed it in [1497455](https://github.com/larq/compute-engine/pull/712/commits/14974551bc1cee02631a9564584781335b00ddda)

larq_compute_engine/mlir/tests/fuse_padding.mlir
Tombana: Done in [e14ddb2](https://github.com/larq/compute-engine/pull/712/commits/e14ddb254e6f5f0ff44acca5c2013556c033f84f)

CNugteren (approved):

larq_compute_engine/mlir/transforms/fuse_padding.cc
CNugteren: Thanks, looks good now!

Issue #713, opened 2022-02-17
LCEInterpreter and converter design


simonmaurer:
The LCE interpreter from `lce.testing.Interpreter` is a standalone class and exposes different properties of the quantized model (scale and zero-point for example). the converter on the other hand is built upon the two methods `convert_keras_model`/`convert_saved_model`.
can you elaborate on your design decision ?

boiling down to follow-up questions:
1. how could one use different delegates in the **LCE Interpreter**  (e.g. `use_xnnpack=True` as in the cmd line parameters of `lce_benchmark_model`)
2. how to add code and support **LCE converter** to use options such as `tf.lite.OpsSet.SELECT_TF_OPS` (as in regular [TFLite](https://www.tensorflow.org/lite/guide/ops_select))

could you give me a hint on this ?

Issue #711, opened 2022-02-09
Select indirect BGEMM kernels - Benchmarking grouped binary convolutions


simonmaurer:
Given the commits #549, #550, #551 **LCE** supportes grouped binary convolutions. this is great work as the standard TFLite still does not support the `groups` argument for inference: https://github.com/tensorflow/tensorflow/issues/40044
I've successfully created models with appropriate channel dimensions, in which the grouped binary convolutions are correctly identified by the **LCE Converter**.

How can I benchmark this with the **lce_benchmark_model** binary ? In other words how can we select the `indirect_bgemm` kernels, as the regular `bgemm` kernels don't support grouped convolution ?

Additionally there is a flag `use_reference_bconv` in the LCE Interpreter, but I do not know what this actually means.
Assuming if it is set to `True` the binary `bgemm` kernels from https://github.com/larq/compute-engine/tree/main/larq_compute_engine/core/bgemm are selected, otherwise the `indirect_bgemm` from https://github.com/larq/compute-engine/tree/main/larq_compute_engine/core/indirect_bgemm.

**Update**: the assumption is not correct, as `use_reference_bconv` is False by default. so `use_reference_bconv` is explained differently.

Tombana:
We currently don't have a CLI flag in `lce_benchmark_model` to choose between these. For internal benchmarks we simply replaced the registration on the following line:

https://github.com/larq/compute-engine/blob/a2611f8e33f5cb9b4d09b7c9aff7053620a24305/larq_compute_engine/tflite/kernels/lce_ops_register.h#L31-L32

with `Register_BCONV_2D_OPT_INDIRECT_BGEMM`.

I'd welcome a PR to make this into a commandline flag, my suggestion would be:

- Add a `bool use_indirect_bgemm` (default false) argument to `RegisterLCECustomOps` with another `if`-branch next to `use_reference_bconv` in [`lce_ops_register.h`](
https://github.com/larq/compute-engine/blob/a2611f8e33f5cb9b4d09b7c9aff7053620a24305/larq_compute_engine/tflite/kernels/lce_ops_register.h).

[...]

PR #717, opened 2022-03-07
Lce benchmark and interpreter flags

simonmaurer:
In general, this PR allows registering of the different kernel implementations of LceBconv2D (`lce_register_ops.h`) by additional parameters in **LceInterpreter** and flags in **lce_benchmark_model** binary

### lce_benchmark_model
- added two cmdline flags `use_reference_bconv`/`use_indirect_bgemm` as global variables to register respective kernel implementations
- instead of manually parsing the flags in lce_benchmark_main.cc the builtin Flags of TF's **BenchmarkModel** are used. Thus by adding **LceBenchmarkTfLiteModel** as a child class of BenchmarkTfLiteModel the global variables are passed in as reference upon object instantiation and set after calling overidden Run() method
### LceInterpreter
- added parameter `use_indirect_bgemm` to register optimized indirect BGEMM kernel using `Register_BCONV_2D_OPT_INDIRECT_BGEMM()` in *lce_register_ops.h*
-  added parameter `use_xnnpack` in *interpreter_wrapper_lite.cc* to apply `BuiltinOpResolver` when true, otherwise `BuiltinOpResolverWithoutDefaultDelegates`

## Related issue number

#711, #713

### DIFF
--- a/larq_compute_engine/tflite/kernels/lce_ops_register.h
+++ b/larq_compute_engine/tflite/kernels/lce_ops_register.h
@@ -2,13 +2,17 @@
 #define COMPUTE_ENGINE_TFLITE_KERNELS_LCE_OPS_REGISTER_H_
 
 #include "tensorflow/lite/context.h"
+//#include "tensorflow/lite/stderr_reporter.h"
 #include "tensorflow/lite/op_resolver.h"
+#include "tensorflow/lite/tools/logging.h"
 
 // This file contains forward declaration of all custom ops
 // implemented in LCE which can be used to link against LCE library.
 
 namespace compute_engine {
 namespace tflite {
+
+using namespace ::tflite;
 
 TfLiteRegistration* Register_QUANTIZE();
 TfLiteRegistration* Register_DEQUANTIZE();
@@ -22,6 +26,11 @@ TfLiteRegistration* Register_BMAXPOOL_2D();
 inline void RegisterLCECustomOps(::tflite::MutableOpResolver* resolver,
                                  const bool use_reference_bconv = false,
                                  const bool use_indirect_bgemm = false) {
+  if (use_reference_bconv && use_indirect_bgemm) {
+    TFLITE_LOG(WARN) << "WARNING: 'use_reference_bconv' and `use_indirect_bgemm` "
+                         "are both set to true. use_indirect_bgemm==true "
+                         "will have no effect.";
+  }
   resolver->AddCustom("LceQuantize",
                       compute_engine::tflite::Register_QUANTIZE());
   resolver->AddCustom("LceDequantize",
--- a/larq_compute_engine/tflite/kernels/lce_ops_register.h
+++ b/larq_compute_engine/tflite/kernels/lce_ops_register.h
@@ -38,7 +38,6 @@ inline void RegisterLCECustomOps(::tflite::MutableOpResolver* resolver,
       resolver->AddCustom("LceBconv2d",
                           compute_engine::tflite::Register_BCONV_2D());
     }
-
   }
   resolver->AddCustom("LceBMaxPool2d",
                       compute_engine::tflite::Register_BMAXPOOL_2D());
--- a/larq_compute_engine/tflite/kernels/lce_ops_register.h
+++ b/larq_compute_engine/tflite/kernels/lce_ops_register.h
@@ -2,7 +2,6 @@
 #define COMPUTE_ENGINE_TFLITE_KERNELS_LCE_OPS_REGISTER_H_
 
 #include "tensorflow/lite/context.h"
-//#include "tensorflow/lite/stderr_reporter.h"
 #include "tensorflow/lite/op_resolver.h"
 #include "tensorflow/lite/tools/logging.h"
 
--- a/larq_compute_engine/tflite/kernels/lce_ops_register.h
+++ b/larq_compute_engine/tflite/kernels/lce_ops_register.h
@@ -27,9 +27,10 @@ inline void RegisterLCECustomOps(::tflite::MutableOpResolver* resolver,
                                  const bool use_reference_bconv = false,
                                  const bool use_indirect_bgemm = false) {
   if (use_reference_bconv && use_indirect_bgemm) {
-    TFLITE_LOG(WARN) << "WARNING: 'use_reference_bconv' and `use_indirect_bgemm` "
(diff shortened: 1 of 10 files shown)


Tombana (approved):
Looks great, thanks a lot!

larq_compute_engine/tflite/python/interpreter.py
```diff
@@ -40,11 +42,17 @@ def __init__(
         flatbuffer_model: bytes,
         num_threads: int = 1,
         use_reference_bconv: bool = False,
+        use_indirect_bgemm: bool = False,
+        use_xnnpack: bool = False,
```
lgeiger, line 46: I think this changes the default or am I missing something? I think that's fine, but can you explain the reasoning behind it?

simonmaurer:
@Tombana my pleasure. was also thinking about overriding *ValidParams()* to check when both `use_reference_bconv` and `use_indirect_bgemm` are set to true to throw an error/warning (same as BenchmarkTfLiteModel/BenchmarkModel do. similarly in the *LCEInterpreter*, what do you think ?
either we can do it in *BenchmarkTfLiteModel*/*LCEInterpreter* or when registering in *lce_register_ops.h*.

on the other hand there's a platform [check](https://github.com/larq/compute-engine/blob/main/larq_compute_engine/tflite/kernels/bconv2d.cc#L593-L599) to discretely register the regular bconv kernel if the optimized BGEMM kernel is not available. this might be also valuable to throw a warning for the caller

lgeiger (commented):
This looks good to me. But can you explain your use case a bit, I'm curious why you need to turn off XNNPack in the interpreter.

larq_compute_engine/tflite/python/interpreter.py
simonmaurer: hi @lgeiger. following the discussion on #713 I've implemented it like this.
A) one reason to include this flag in the *LceInterpreter* is to make it reflect more the behavior of the benchmark binary.
B) the second reason is dynamic allocation, thus having models with `allocation_type==kTfLiteDynamic`.
we had models like this and when we set `use_xnnpack=true` in the *benchmark_model* there will be an error thrown. this would be true for *LCEInterpreter* as well, as under the hood XNNPACK is "silently" applied.

lgeiger (approved):

Tombana:
> @Tombana my pleasure. was also thinking about overriding _ValidParams()_ to check when both `use_reference_bconv` and `use_indirect_bgemm` are set to true to throw an error/warning (same as BenchmarkTfLiteModel/BenchmarkModel do. similarly in the _LCEInterpreter_, what do you think ? either we can do it in _BenchmarkTfLiteModel_/_LCEInterpreter_ or when registering in _lce_register_ops.h_.

That sounds good yes. It's probably easiest to do it in `lce_ops_register.h`, then all uses are covered. As for the output mechanism, it seems that by default the `Interpreter` uses `DefaultErrorReporter()` (from `#include "tensorflow/lite/stderr_reporter.h"`) so its probably the cleanest to use that instead of calling `printf` directly.

> on the other hand there's a [platform check](https://github.com/larq/compute-engine/blob/main/larq_compute_engine/tflite/kernels/bconv2d.cc#L593-L599) to discretely register the regular bconv kernel if the optimized BGEMM kernel is not available. this might be also valuable to throw a warning for the caller

It seems that the [TFLite conv kernels](https://github.com/tensorflow/tensorflow/blob/919f693420e35d00c8d0a42100837ae3718f7927/tensorflow/lite/kernels/conv.cc#L1187) don't have any warnings there either, so I think we can leave those like it is.

simonmaurer:
> > @Tombana my pleasure. was also thinking about overriding _ValidParams()_ to check when both `use_reference_bconv` and `use_indirect_bgemm` are set to true to throw an error/warning (same as BenchmarkTfLiteModel/BenchmarkModel do. similarly in the _LCEInterpreter_, what do you think ? either we can do it in _BenchmarkTfLiteModel_/_LCEInterpreter_ or when registering in _lce_register_ops.h_.
> 
> That sounds good yes. It's probably easiest to do it in `lce_ops_register.h`, then all uses are covered. As for the output mechanism, it seems that by default the `Interpreter` uses `DefaultErrorReporter()` (from `#include "tensorflow/lite/stderr_reporter.h"`) so its probably the cleanest to use that instead of calling `printf` directly.
> 
> > on the other hand there's a [platform check](https://github.com/larq/compute-engine/blob/main/larq_compute_engine/tflite/kernels/bconv2d.cc#L593-L599) to discretely register the regular bconv kernel if the optimized BGEMM kernel is not available. this might be also valuable to throw a warning for the caller
> 
> It seems that the [TFLite conv kernels](https://github.com/tensorflow/tensorflow/blob/919f693420e35d00c8d0a42100837ae3718f7927/tensorflow/lite/kernels/conv.cc#L1187) don't have any warnings there either, so I think we can leave those like it is.

perfect. I've added this as you suggested with the `DefaultErrorReporter()` which returns a static pointer to report an error for this use case. as it should be a warning it uses `TFLITE_LOG(WARN)` if y'all are okay with that.

Tombana:
> @Tombana @lgeiger any hints why the inclusion of "@org_tensorflow//tensorflow/lite/tools/logging" is only working for tf_cc_binary() in BUILD of `lce_benchmark_model` but not for `interpreter_wrapper_lite` in py_bind_extension() as a dependency?

I'm not sure, I think you might need `..../tools:logging` instead of `..../tools/logging`.

larq_compute_engine/tflite/kernels/lce_ops_register.h
```diff
@@ -2,24 +2,36 @@
 #define COMPUTE_ENGINE_TFLITE_KERNELS_LCE_OPS_REGISTER_H_
 
 #include "tensorflow/lite/context.h"
+//#include "tensorflow/lite/stderr_reporter.h"
```
Tombana, line 5: In case you missed it, this one can be removed. I think the `TFLITE_LOG(WARN)` solution is good :+1:

simonmaurer: thanks a lot 🤙. It worked

Tombana:
I think the PR is finished, right @simonmaurer ? Then it can be merged.

simonmaurer:
yessir 💪

lgeiger (approved):

PR #720, opened 2022-04-01
Make assembly compatible with clang

Tombana:
## What do these changes do?
The syntax used in the inline assembly of the 32-bit ARM kernels was not compatible with the LLVM Integrated Assembler. A workaround was to use `clang -fno-integrated-as` but that makes things complex when crosscompiling and `PATH` is not setup correctly.

This PR uses the syntax that's also used in the ARM documentation; [example](https://developer.arm.com/documentation/dui0802/a/Advanced-SIMD-and-Floating-point-Programming--32-bit-/VDUP).

## How Has This Been Tested?
CI tests the correctness of this kernel.


CNugteren (approved):

Tombana:
This can be merged after https://github.com/larq/compute-engine/pull/721 is merged.

PR #722, opened 2022-04-01
:arrow_up: tensorflow@2.8.0

CNugteren:
## What do these changes do?
This PR updates LCE to TensorFlow 2.8.0. Quite a few changes have been made, and most likely more are needed. I'll highlight some of the more controversial changes below in the form of a comment.

## How Has This Been Tested?
All unittests have been ran locally. Fixes have been made as compilation, linking, or testing errors showed-up. However, some of them do not succeed yet, most notably the MLIR tests. This PR is therefore still a draft.

## Benchmark Results
Latency has not been verified yet.

## Related issue number
N/A.


.tensorflow.bazelrc
```diff
@@ -136,6 +136,7 @@ build:elinux_aarch64 --config=elinux
 build:elinux_aarch64 --cpu=aarch64
 build:elinux_armhf --config=elinux
 build:elinux_armhf --cpu=armhf
+build:elinux_armhf --copt -mfp16-format=ieee
```
CNugteren, line 139: Taken from https://github.com/tensorflow/tensorflow/commit/4ec84aa99372ca7899fb12715edd2bfe3c947c88

larq_compute_engine/mlir/tests/bitpack-weights.mlir
```diff
@@ -2,11 +2,11 @@
 
 // CHECK-LABEL: @bitpack_bconv2d_filters
 func @bitpack_bconv2d_filters(%arg0: tensor<256x32x32x1xi32>, %arg1: tensor<16xf32>, %arg2: tensor<16xf32>, %arg3: none) -> tensor<256x30x30x16xf32> {
-  %cst = constant dense<1.0> : tensor<16x3x3x3xf32>
+  %cst = arith.constant dense<1.0> : tensor<16x3x3x3xf32>
```
CNugteren, line 5: I renamed all `constant dense` into `arith.constant dense` inside the MLIR files and all `ConstantOp` into `Arith_ConstantOp` in the `*.td` files. This made most of the MLIR tests pass again, but it might have been too much: we should understand what the difference between the two is, since there are still a few places inside TensorFlow where `ConstantOp` is used in favor of `Arith_ConstantOp`.

larq_compute_engine/mlir/tf_tfl_passes.cc
```diff
@@ -131,6 +127,14 @@ void AddTFToLCETFLConversionPasses(
   // constant ops.
   pass_manager->addPass(mlir::tf_saved_model::CreateFreezeGlobalTensorsPass());
 
+  // Reduce operands of TFL::While without changing the outcome.
```
CNugteren, line 130: I tried to synchronize this file here as much as possible with the latest changes. That means some new passes have been added. I did not check what they do and therefore are not sure that we need/want them in. This should perhaps be checked?

larq_compute_engine/mlir/tf_to_tfl_flatbuffer.cc
```diff
@@ -82,10 +82,13 @@ Status ConvertTFExecutorToFlatbuffer(mlir::ModuleOp module, bool export_to_mlir,
   }
 
   // This is the only modification compared to the upstream tensorflow file
+  // TODO: This is no longer the case, these files have diverged since TF2.6
```
CNugteren, line 85: >This is the only modification compared to the upstream tensorflow file

This is no longer the case, there is quite some diversion. Should we try to synchronize again?

larq_compute_engine/tflite/kernels/bconv2d.cc
```diff
@@ -273,11 +271,9 @@ TfLiteStatus Prepare(TfLiteContext* context, TfLiteNode* node) {
 
   // Now allocate the buffers
   if (need_im2col) {
-    // Allocate the im2col tensor if necessary
-    if (op_data->im2col_id == kTensorNotAllocated) {
```
CNugteren, line 277: With this if statement things did no longer work, so I had to remove it.

larq_compute_engine/tflite/tests/interpreter_test.py
```diff
@@ -49,15 +49,18 @@ def test_interpreter_multi_input(use_iterator):
     interpreter = Interpreter(converter.convert(), num_threads=2)
     assert interpreter.input_types == [np.float32, np.float32]
     assert interpreter.output_types == [np.float32, np.float32]
-    assert interpreter.input_shapes == [(1, 24, 24, 2), (1, 24, 24, 1)]
-    assert interpreter.output_shapes == [(1, 24 * 24 * 2), (1, 24 * 24 * 1)]
+    assert interpreter.input_shapes == [(1, 24, 24, 1), (1, 24, 24, 2)]
+    assert sorted(interpreter.output_shapes) == [(1, 24 * 24 * 1), (1, 24 * 24 * 2)]
```
CNugteren, line 53: The input order seems to be reversed and the output order was random in my case. Not sure if that is an issue, but I made the test such that it works on my machine.

third_party/tensorflow_patches/fix_armhf_xnnpack.patch
```diff
@@ -0,0 +1,24 @@
+diff --git a/.bazelrc b/.bazelrc
+index b1d4eee905d..ada1395298f 100644
+--- a/.bazelrc
++++ b/.bazelrc
+@@ -572,6 +572,7 @@ build:elinux_aarch64 --distinct_host_configuration=true
+ build:elinux_armhf --config=elinux
+ build:elinux_armhf --cpu=armhf
+ build:elinux_armhf --distinct_host_configuration=true
++build:elinux_armhf --copt -mfp16-format=ieee
```
CNugteren, line 9: This is from https://github.com/tensorflow/tensorflow/commit/4ec84aa99372ca7899fb12715edd2bfe3c947c88

larq_compute_engine/mlir/transforms/quantize_patterns.td
```diff
@@ -69,3 +69,7 @@ def : Pat<(TFL_QuantizeOp
 def : Pat<(TFL_QuantizeOp (LQ_DequantizeOp:$output $input), $qtype),
           (LQ_DequantizeOp $input),
           [(HasOneUse $output)]>;
+
+def : Pat<(LQ_QuantizeOp (TFL_DequantizeOp:$output $input)),
+          (LQ_QuantizeOp $input),
+          [(HasOneUse $output)]>;
```
lgeiger, line 75: This restores `v0.6.2` behaviour when converting `toy_int8_model` from the end2end unittests:

before | after (like v0.6.2)
---|---
Screenshot 2022-04-06 at 13 00 42 | Screenshot 2022-04-06 at 13 00 51

I think this is due to the fact that fake quantizers might be handled slightly differently now. In practice this shouldn't matter much since it's not necessary/recommended to use a int8 quantizer in a feed forward network directly before a binarization.

larq_compute_engine/mlir/tf_tfl_passes.cc
lgeiger: I double checked and could only find a minor difference. Fixed in 114a945b7670d0ffe96678f81d4b72bfbb4a0e09

larq_compute_engine/tests/end2end_test.py
```diff
@@ -38,7 +38,9 @@ def dummy(x):
                 use_bias=False,
                 activation=activation,
             )(x)
-            x = tf.keras.layers.BatchNormalization(momentum=0.7)(x)
+            x = tf.keras.layers.BatchNormalization(
+                momentum=0.7, beta_initializer="random_normal"
```
lgeiger, line 42: Using a random bias works around a known bug in the experimental_default_int8_range pass. This also allows us to re-enable a skipped test below.

larq_compute_engine/mlir/tf_to_tfl_flatbuffer.cc
lgeiger: I think for now this seems to work OK. I'm happy to synchronize this in a follow up PR since it seems they added quite a few new features, but I am a bit unsure if we already need to pull in all the new experimental flags.

larq_compute_engine/tests/end2end_test.py
Tombana: Is it known what the bug is? Can we add a comment to this file to explain why we have the random bias?

larq_compute_engine/mlir/tf_tfl_passes.cc
```diff
@@ -43,6 +42,9 @@ void AddQuantizationPasses(const mlir::TFL::QuantizationSpecs& quant_specs,
       quant_specs.inference_type != quant_specs.inference_input_type;
   pass_manager->addNestedPass<mlir::FuncOp>(
       mlir::TFL::CreatePostQuantizePass(emit_quant_adaptor_ops));
+  pass_manager->addPass(mlir::TFL::CreateLCEQuantizePass());
+  pass_manager->addNestedPass<mlir::FuncOp>(
+      mlir::TFL::CreatePostQuantizePass(emit_quant_adaptor_ops));
```
lgeiger, line 47: Stray `tfl.dequantize` ops might prevent the `lq.quantize(tfl.dequantize(x))` -> `lq.quantize(x)` pattern from properly working.

This change runs the pattern again, after unnecessary `tfl.quantize` and `tfl.dequantize` ops have been removed in the `CreatePostQuantizePass`.

before | after
---|---
Screenshot 2022-04-08 at 10 54 16 | Screenshot 2022-04-08 at 11 00 22

lgeiger (approved):

larq_compute_engine/tflite/kernels/bconv2d.cc
Tombana: I'm still a bit confused by this, they still seem to have this check in the latest conv kernel code: https://github.com/tensorflow/tensorflow/blob/bdb6f62c32ee16016de8ec83e42a96b59ac4f9fa/tensorflow/lite/kernels/conv.cc#L272

My worry is that if `Prepare` is now called several times, after resizing the input for example, that it allocates a new tensor without de-allocating the old one.

CNugteren: I can have another look, I'll revert the fix, re-run the failing test, and try to get more debug info.

larq_compute_engine/mlir/ir/lce_ops.cc
```diff
@@ -73,5 +77,12 @@ void LarqDialect::initialize() {
 #include "larq_compute_engine/mlir/ir/lce_ops.cc.inc"
       >();
 }
+
+Operation* LarqDialect::materializeConstant(OpBuilder& builder, Attribute value,
+                                            Type type, Location loc) {
+  if (arith::ConstantOp::isBuildableWith(value, type))
+    return builder.create<mlir::arith::ConstantOp>(loc, type, value);
+  return nullptr;
+}
```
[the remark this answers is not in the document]
Tombana: In the tensorflow example they first try to make it into a `TFL::ConstOp` before they try the `arith::ConstOp`. I guess we don't need that here because the larq dialect is used for TF ops and not TFL ops?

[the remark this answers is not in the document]
Tombana: In the tensorflow example they first try to make it into a `TFL::ConstOp` before they try the `arith::ConstOp`. I guess we don't need that here because the larq dialect is used for TF ops and not TFL ops?

larq_compute_engine/tflite/kernels/bconv2d.cc
CNugteren: This should be fixed now, in the end all we needed was the temporaries assignment to move outside, not the add-tensors.

larq_compute_engine/mlir/ir/lce_ops.cc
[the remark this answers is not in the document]
lgeiger: If I understand their code correctly it will create a `tfl.pseudo_const` if the [value is opaque](https://mlir.llvm.org/docs/Dialects/Builtin/#opaqueelementsattr) which isn't supported by `arith::ConstantOp`. I don't think this is a case that we will encounter. In general `TFL::ConstOp` will always be [folded to `arith::ConstantOp` if possible](https://github.com/tensorflow/tensorflow/blob/3f878cff5b698b82eea85db2b60d65a2e320850e/tensorflow/compiler/mlir/lite/ir/tfl_ops.cc#L2770-L2794) so I think this should be fine to keep it like this for now.

Tombana (approved):

lgeiger:
I triggered a [test release](https://github.com/larq/compute-engine/actions/runs/2160768163), let's see how this goes.

Tombana:
> I triggered a [test release](https://github.com/larq/compute-engine/actions/runs/2160768163), let's see how this goes.

The windows looks like it succeeded within 12 minutes but actually the build failed with
```
The target you are compiling requires Visual C++ build tools. 
Bazel couldn't find a valid Visual C++ build tools installation on your machine.
```

The linux builds fail because the manylinux toolchain bazel files can't be found; they've probably been updated to newer versions.

lgeiger:
> The windows looks like it succeeded within 12 minutes but actually the build failed with

I pinned the Windows VM version to 2019 in 5bf5748f3d3b4258a3efce19f0d0d7ab9837d85f, let's see if this fixes the issue. Alternatively we could also change the `BAZEL_VC` path to make it work with Windows 2022, but not sure if we would run into other issues then.

> The linux builds fail because the manylinux toolchain bazel files can't be found; they've probably been updated to newer versions.

I pinned the docker image to 2.8 in 3923692ac56690a87439536f82d960cbcfa02a29 which didn't seem to fix things.

It seems like the TensorFlow repo removed their toolchains from the repo. I copied the [toolchain from the TF Addons repo](https://github.com/tensorflow/addons/tree/master/build_deps/toolchains/gcc7_manylinux2010-nvcc-cuda11) in bdb6f2009d72cee0fa17a550e507782ccd8e0718. They use the same docker container to build the wheels as we do, so I expect this to work.

CNugteren:
Two builds failed:
* The Linux Python 3.9 builds. I couldn't see any logs there so I assume there was some GitHub Actions glitch. I've re-triggered it.
* All Windows builds timed out after 360 minutes (6 hours). Earlier build also reached around that point. I've added a time-out of 10 hours now, let's see if that works: https://github.com/larq/compute-engine/runs/6036809952?check_suite_focus=true

CNugteren:
The timeout increase did not work, it still times out after 6 hours: https://github.com/larq/compute-engine/runs/6036809795?check_suite_focus=true. Perhaps that is some global setting or a Github Actions maximum? In any case, this time two out of the 3 Windows builds passed since they finished in five and a half hours, only one timed out. This behaviour was also seen earlier without TF 2.8, it was always tight with this 6 hours maximum. So I do not consider this related to this branch.

My proposal is to merge this in and then investigate making the releases more stable in a separate PR (Linux builds random failures and Windows builds time outs)? I can open an issue for that.

Tombana:
I agree, let's merge it.

I think I've seen this before and in that particular case the system had run out of memory. That could be fixed by adding swap space or changing the number of jobs that bazel runs in parallel. But maybe this problem is different.
```
gcc: internal compiler error: Killed (program cc1plus)
```

.github/workflows/release.yml
```diff
@@ -228,7 +228,8 @@ jobs:
 
   windows-release-wheel:
     name: Build release wheels for Windows
-    runs-on: windows-latest
+    runs-on: windows-2019
+    timeout-minutes: 600
```
lgeiger, line 232: I think we can revert this change, since the maximum runtime on the default GitHub actions runners is [6h by default anyway](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits).
```suggestion
```

PR #723, opened 2022-04-08
TF 2.8: Update converter to sync with upstream tensorflow

lgeiger:
This PR synchronises `tf_to_tfl_flatbuffer.cc` with upstream TensorFlow to fix https://github.com/larq/compute-engine/pull/722#discussion_r840649004


larq_compute_engine/mlir/tf_tfl_passes.cc
```diff
@@ -27,30 +27,33 @@ void AddQuantizationPasses(const mlir::TFL::QuantizationSpecs& quant_specs,
                            mlir::OpPassManager* pass_manager) {
   pass_manager->addNestedPass<mlir::FuncOp>(
       mlir::TFL::CreatePrepareQuantizePass(quant_specs));
-  pass_manager->addPass(mlir::TFL::CreateLCEQuantizePass());
+  pass_manager->addNestedPass<mlir::FuncOp>(mlir::TFL::CreateLCEQuantizePass());
```
lgeiger, line 30: MLIR threw a nice error message that recommended changing our passes to be nested passes. Looks like the overall module structure changed, which required this update.

larq_compute_engine/mlir/python/common.cc
```diff
@@ -118,18 +115,10 @@ pybind11::bytes ConvertMLIRModuleToTFLiteFlatBuffer(
...
 
-  mlir::PassManager pm(&context, mlir::OpPassManager::Nesting::Implicit);
-  tensorflow::SetCrashReproducer(pm);
-
-  tensorflow::AddTFToLCETFLConversionPasses(quant_specs, &pm, target);
-
-  // Convert back to outlined while format for export back to flatbuffer.
-  pm.addPass(mlir::TFL::CreateWhileOutlinePass());
-  pm.addPass(mlir::TFL::CreateRuntimeVerifyPass());
-
```
Tombana, line 129: I see that this block has moved into `ConvertTFExecutorToFlatbuffer`, and `AddTFToLCETFLConversionPasses` is now split in two.
However `tensorflow::SetCrashReproducer(pm);`, `CreateWhileOutlinePass` and `CreateRuntimeVerifyPass` were not moved there, is that intentional? I see that `CreateRuntimeVerifyPass` is one of the passes inside `TFToLCETFL` so that's probably fine (although it's no longer the final pass).

larq_compute_engine/mlir/tf_to_tfl_flatbuffer.cc
```diff
@@ -70,25 +78,64 @@ Status ConvertTFExecutorToFlatbuffer(mlir::ModuleOp module, bool export_to_mlir,
...
+  // Freeze variables if a session is provided.
+  if (session.hasValue()) {
+    mlir::TFL::ErrorCollectorInstrumentation collector(module.getContext());
+    if (failed(mlir::tf_saved_model::FreezeVariables(module,
+                                                     session.getValue()))) {
+      auto status = statusHandler.ConsumeStatus();
+      mlir::TFL::ErrorCollector* collector =
+          mlir::TFL::ErrorCollector::GetErrorCollector();
+      if (!collector->CollectedErrors().empty()) {
+        return errors::InvalidArgument("Variable constant folding is failed.");
```
Tombana, line 105:
```suggestion
        return errors::InvalidArgument("Variable constant folding has failed.");
```

larq_compute_engine/mlir/tf_to_tfl_flatbuffer.cc
```diff
@@ -70,25 +78,64 @@ Status ConvertTFExecutorToFlatbuffer(mlir::ModuleOp module, bool export_to_mlir,
...
+    }
+  }
+  pass_manager.clear();
+  tensorflow::AddPostVariableFreezingTFToLCETFLConversionPasses(
+      saved_model_dir, quant_specs, &pass_manager, target);
+  if (failed(pass_manager.run(module))) {
+    auto status = statusHandler.ConsumeStatus();
+    mlir::TFL::ErrorCollector* collector =
+        mlir::TFL::ErrorCollector::GetErrorCollector();
+    for (const auto& error_data : collector->CollectedErrors()) {
```
Tombana, line 117: Is it intentional that this error collection stuff is done if the second phase failed, but not if the first phase failed (i.e. line 93)?

Tombana (approved):
Thanks for taking care of this!

larq_compute_engine/mlir/tf_to_tfl_flatbuffer.cc
lgeiger: To be fair, I don't know. This is primarily [copied from TensorFlow](https://github.com/tensorflow/tensorflow/blob/3f878cff5b698b82eea85db2b60d65a2e320850e/tensorflow/compiler/mlir/lite/tf_to_tfl_flatbuffer.cc#L247-L251) and they seem to do it like this.

larq_compute_engine/mlir/python/common.cc
lgeiger: > However `CreateWhileOutlinePass` and `CreateRuntimeVerifyPass` were not moved there, is that intentional

These were moved to `tf_tfl_passes.cc` following https://github.com/tensorflow/tensorflow/commit/6347e46c2159293e7307a6ef71f7c8bc5ceec32a

> However `tensorflow::SetCrashReproducer(pm);` were not moved there, is that intentional

This was done following https://github.com/tensorflow/tensorflow/commit/a68046e7d2ee3396fee7cac07f915294619a15a7 so I assume that this functionality is now handled somewhere else internally.

Tombana (approved):

PR #728, opened 2022-04-21
Add bazel cache for release jobs

Tombana:
## What do these changes do?
This adds a bazel cache for release jobs:
- if they timeout we can simply restart them and they should finish within the time limit
- if they fail for any other reason we can try fixes without having to wait for 6 hours

## How Has This Been Tested?
It is running here: https://github.com/larq/compute-engine/actions/runs/2200323727

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -140,12 +140,21 @@ jobs:
       - uses: actions/setup-python@v3
         with:
           python-version: ${{ matrix.python-version }}
+      - uses: google-github-actions/auth@v0
+        continue-on-error: true
+        with:
+          credentials_json: ${{ secrets.gcs_bazel_cache }}
       - name: Build macOS wheels
         run: |
           python --version
           python -m pip install delocate wheel setuptools numpy six --no-cache-dir
 
           ./configure.py
+
+          if [[ -n $GOOGLE_APPLICATION_CREDENTIALS ]]; then
+            echo -e 'build --remote_http_cache=https://storage.googleapis.com/plumerai-bazel-cache/lce-release-macos' >> .bazelrc.user
+            echo -e 'build --google_default_credentials' >> .bazelrc.user
+          fi
 
           bazelisk build :build_pip_pkg --copt=-fvisibility=hidden --copt=-mavx --copt=-mmacosx-version-min=10.13 --linkopt=-mmacosx-version-min=10.13 --linkopt=-dead_strip --distinct_host_configuration=false
           bazel-bin/build_pip_pkg artifacts --plat-name macosx_10_13_x86_64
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -140,13 +140,22 @@ jobs:
...
+          credentials_json: ${{ secrets.gcs_bazel_cache }}
       - name: Build macOS wheels
         run: |
           python --version
           python -m pip install delocate wheel setuptools numpy six --no-cache-dir
 
           ./configure.py
 
+          if [[ -n $GOOGLE_APPLICATION_CREDENTIALS ]]; then
+            echo -e 'build --remote_http_cache=https://storage.googleapis.com/plumerai-bazel-cache/lce-release-macos' >> .bazelrc.user
```
CNugteren, line 155: Just a question, would it help for Bazel to do something like this instead:
```suggestion
            echo -e 'build --remote_http_cache=https://storage.googleapis.com/plumerai-bazel-cache/lce-release-macos-${{ matrix.python-version }}' >> .bazelrc.user
```
Then each Python version will have its own cache. Or is Bazel smart enough to not overwrite those caches?

CNugteren (approved):

.github/workflows/release.yml
Tombana: Good idea, I'll add that.

Tombana:
New run with different folders for each python version, and with passing the credentials into the manylinux container: https://github.com/larq/compute-engine/actions/runs/2200491384

.github/workflows/release.yml
```diff
@@ -140,13 +140,22 @@ jobs:
...
+        with:
+          credentials_json: ${{ secrets.gcs_bazel_cache }}
       - name: Build macOS wheels
         run: |
           python --version
           python -m pip install delocate wheel setuptools numpy six --no-cache-dir
 
           ./configure.py
 
+          if [[ -n $GOOGLE_APPLICATION_CREDENTIALS ]]; then
```
lgeiger, line 154: For the release workflow this should always exist, but I agree it's probably good to keep this check just to be safe.

.github/workflows/release.yml
```diff
@@ -206,13 +224,27 @@ jobs:
...
+
+          cp $GOOGLE_APPLICATION_CREDENTIALS gcloud-credentials.json
+
           docker run -e LCE_RELEASE_VERSION=${{ github.event.inputs.version }} \
+            -e GOOGLE_APPLICATION_CREDENTIALS=/compute-engine/gcloud-credentials.json \
             -v ${PWD}:/compute-engine -w /compute-engine \
             tensorflow/build:2.8-python${{ matrix.python-version }} \
             .github/tools/release_linux.sh
 
+          rm gcloud-credentials.json
```
lgeiger, line 246: How about directly mounting the secret into the docker container?
```suggestion
          docker run -e LCE_RELEASE_VERSION=${{ github.event.inputs.version }} \
            -e GOOGLE_APPLICATION_CREDENTIALS=/tmp/gcloud-credentials.json \
            -v $GOOGLE_APPLICATION_CREDENTIALS:/tmp/gcloud-credentials.json:ro \
            -v ${PWD}:/compute-engine -w /compute-engine \
            tensorflow/build:2.8-python${{ matrix.python-version }} \
            .github/tools/release_linux.sh
```

lgeiger (commented):
Awesome! I just have one minor comment, other than that this looks great!

.github/workflows/release.yml
Tombana: That is much nicer indeed, thanks.

Tombana:
On the previous run only the python 3.7 job failed with an internal gcc error.
I've applied Lukas' suggestion of mounting the gcloud credentials directly, and I've set off another run: https://github.com/larq/compute-engine/actions/runs/2202486415
If the cache is working as expected these should now all finish quickly.

lgeiger (approved):

Tombana:
The cache is working but not completely: the builds take about ~25 minutes, so something is working but it's still rebuilding some parts.
(Some of the failing workflows are unrelated, caused by a failed download)

PR #733, opened 2022-04-25
Add Python 3.10 to release builds

CNugteren:
## What do these changes do?
Add Python 3.10 to the release build scripts. This is now possible since TF 2.8 supports Python 3.10.

## How Has This Been Tested?
A test release build is triggered to verify if all works fine (and to fill the cache):
https://github.com/larq/compute-engine/actions/runs/2220640761

## Benchmark Results
N/A

## Related issue number
N/A

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,10 +12,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - name: Set up Python 3.10
+      - name: Set up Python 3.8
         uses: actions/setup-python@v3
         with:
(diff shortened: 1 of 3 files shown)


Tombana:
There is another
```
gcc: internal compiler error
```
We could try `--local_ram_resources=HOST_RAM*.40` to the bazel flags.

CNugteren:
I went with a re-run first, let's see if that resolves it.

CNugteren:
The test release worked fine: https://github.com/larq/compute-engine/actions/runs/2220640761, so I'll merge this in.

PR #749, opened 2022-08-22
⬆️ TensorFlow@2.10

### LINKED ISSUE #748: Get Operator-wise Profiling Results
Our group wants to profile BNNs on some ARM64 processors. We tried running `./lce_benchmark_model_aarch64` with the `--enable_op_profiling` set to `true`. The result looks like this (for QuickNet), with many floating-point operators grouped together as `TfLiteXNNPackDelegate` in the report. Is there a way to get the inference time for more specific operators like ADD, MUL, CONV_2D, MAX_POOL_2D, etc? Thank you! 

Screen Shot 2022-08-18 at 3 11 24 PM

lgeiger:
## What do these changes do?
This PR upgrade our TensorFlow dependency to 2.10

## How Has This Been Tested?
CI

## Benchmark Results
TODO

## Related Issue Number
Closes #748


lgeiger:
Looks like on ARM32 we are still running into https://github.com/tensorflow/tensorflow/issues/57364. Locally I ran into another build issue when trying to run the end2end tests, but maybe that's not a problem on CI.

CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ set(TFLITE_SOURCE_DIR "${TENSORFLOW_SOURCE_DIR}/tensorflow/lite")
 add_subdirectory("${TFLITE_SOURCE_DIR}" "${CMAKE_CURRENT_BINARY_DIR}/tensorflow-lite" EXCLUDE_FROM_ALL)
 
 # Generic compilation options and settings
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
```
CNugteren, line 14: Side question: does this mean anything for our CI builds and releases in the `bboxpi` repository? There we do have to make sure that we can link with old compilers for example.

[Here](https://github.com/plumerai/bboxpi/blob/main/.github/workflows/cpp.yml) we build with GCC 8, which should be fine according to https://gcc.gnu.org/projects/cxx-status.html#cxx17

So I guess it is all fine but just double-checking.

lgeiger:
It looks like we might not even need to restrict the supported version of TensorFlow as long as `flatbuffers==2.0` is installed.  
E.g. TF 2.3 and older didn't even rely on the `flatbuffers` pip package so we might be OK just requiring `flatbuffers==2.0` as a dependency. For me locally the end2end tests pass with `tensorflow==2.9` and `flatbuffers==2.0`.

Tombana:
We could also print a runtime warning if `version(tensorflow) <= 2.10` without aborting completely.

.bazelrc
```diff
@@ -25,6 +26,9 @@ build --define=with_xla_support=false
 # Disable MKL
 build --define=enable_mkl=false --define=build_with_mkl=false --define=tensorflow_mkldnn_contraction_kernel=0
 
+# Disable XNNPACK on elinux_armhf due to https://github.com/tensorflow/tensorflow/issues/57364
+build:elinux_armhf --define tflite_with_xnnpack=false
```
CNugteren, line 30: As far as I understand, this is for 32-bit ARM, right? So should a similar define then also be added to the 32-bit RPi3 configuration below?

lgeiger: The 32-bit RPi3 config inherits from `elinux_armhf` so this will automatically be applied there as well:
https://github.com/larq/compute-engine/blob/84e30bd80e18d2aeba923d6c44d40098b8702184/.bazelrc#L29

CNugteren (approved):
Looks good, I approve the changes. Under the condition that we wait another week or so till TF 2.10 is actually released, right?

Tombana (approved):

PR #754, opened 2022-09-12
Suppress C++ compiler warnings

lgeiger:
## What do these changes do?
This PR suppresses C++ warnings since our build since the CI build logs are getting really long. This is copied directly from the [TensorFlow bazel config](https://github.com/tensorflow/tensorflow/blob/e6177657b966d4a0d7c9dcff0e87789bfc968da8/.bazelrc#L295-L301). Not sure whether this ignores any important warnings, but to be honest the logs are so messy that it is almost impossible to see something in there anyway.

## How Has This Been Tested?
I tested this with a full converter build and it reduces the logged lines in GitHub actions from 19373 to 339.

### DIFF
--- a/.tensorflow.bazelrc
+++ b/.tensorflow.bazelrc
@@ -76,6 +76,15 @@ build:dynamic_kernels --copt=-DAUTOLOAD_DYNAMIC_KERNELS
 # Don't trigger --config=<host platform> when cross-compiling.
 build:android --noenable_platform_specific_config
 
+# Suppress C++ compiler warnings, otherwise build logs become 10s of MBs.
+build:android --copt=-w
+build:ios --copt=-w
+build:linux --copt=-w
+build:linux --host_copt=-w
+build:macos --copt=-w
+build:windows --copt=/W0
+build:windows --host_copt=/W0
+
 # Tensorflow uses M_* math constants that only get defined by MSVC headers if
 # _USE_MATH_DEFINES is defined.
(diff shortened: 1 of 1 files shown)


Tombana (approved):

lgeiger:
Looking at TensorFlow's bazel config again, it seems like they [specifically enable some warnings on Linux](https://github.com/tensorflow/tensorflow/blob/e6177657b966d4a0d7c9dcff0e87789bfc968da8/.bazelrc#L303-L313). Should we do the same?

Tombana:
> Looking at TensorFlow's bazel config again, it seems like they [specifically enable some warnings on Linux](https://github.com/tensorflow/tensorflow/blob/e6177657b966d4a0d7c9dcff0e87789bfc968da8/.bazelrc#L303-L313). Should we do the same?

I think that on CI it is indeed a bit useless to have any warnings printed, since there are so many and most of them are not from our code. If it would be possible to only get warnings for everyhing in the `larq_compute_engine` dir, that might be a nice improvement. It might also be nice to make this "no warnings" option something for CI only, so that local runs still have all/most warnings. To achieve that maybe we have to add these no-warning flags in the github actions bazel setup script, i.e. `echo "build:linux -no_warnings" >> .bazelrc.user` or something. Or maybe make a `build:ci_no_warnings` config where we disable the warnings, and then in the github actions file add `--config=ci_no_warnings`. What do you think?

PR #784, opened 2023-08-02
⬆️ tensorflow@2.13.0

CNugteren:
## What do these changes do?
This updates TensorFlow from version 2.11 to 2.13 (note that 2.12 was skipped, so this is a double update). The code was updated to make it compile and run again.

I checked the diff for `tf_tfl_passes.cc` to see if we need to update anything but I don't think it is significant. You can see it for yourself with (in the `third_party/tensorflow` folder):
```
git diff d5b57ca93e506df258271ea00fc29cf98383a374 1cb1a030a62b169d90d34c747ab9b09f332bf905  -- tensorflow/compiler/mlir/lite/tf_tfl_passes.cc
```

I removed `third_party/tensorflow_patches/embedded_toolchain.patch` since I believe that same patch has been applied to the mainline TensorFlow already. And I removed `third_party/tensorflow_patches/disable_forced_mkl.patch` because that file is now completely different, so I'm hoping it is no longer necessary - or perhaps CI will warn us.

Remains a draft until CI is happy -  best to wait with reviewing.

## How Has This Been Tested?
I ran some of the compilation and tests locally, but CI should be able to cover everything.

## Related issue number
https://github.com/plumerai/lce-micro/issues/1358 (internal issue)


Tombana (approved):

lgeiger (approved):
Very nice! Thanks for going through and updating this 🎉

PR #785, opened 2023-08-07
:arrow_up: 0.13.0

CNugteren:
Bump to 0.13.0 (not 0.12.0 to reflect the fact that it is based on TF 2.13, not 2.12).

Changelog will be in the release notes.


lgeiger:
While we're at it should we also add wheels for Python 3.11?

Tombana (approved):

PR #786, opened 2023-08-07
Build wheels for Python 3.11

lgeiger:
This PR adds Python 3.11 to our CI test and release matrix.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -137,7 +137,7 @@ jobs:
     runs-on: macos-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10"]
+        python-version: [3.7, 3.8, 3.9, "3.10", 3.11]
       fail-fast: false
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


Tombana:
Perhaps this docker image should be updated as well:
https://github.com/larq/compute-engine/blob/9f095bd6fad5829a9de87af6e16dad2c6f7af0c0/.github/workflows/release.yml#L248

Also, I think we should drop support for Python 3.7 and 3.8 (maybe even 3.9?). It seems tensorflow requires at least 3.8 anyway.

lgeiger:
> Also, I think we should drop support for Python 3.7 and 3.8 (maybe even 3.9?). It seems tensorflow requires at least 3.8 anyway.

Yeah I think it's a good idea to drop support for Python 3.7 and 3.8. The latest Numpy release also only support Python 3.9 an newer so we can just align to [their schedule](https://numpy.org/neps/nep-0029-deprecation_policy.html) as well.

Tombana (approved):

CNugteren (approved):

PR #791, opened 2023-08-09
Use command parameter files when compiling on Windows

lgeiger:
Hopefully this fixed our Windows release builds. TensorFlow has the [same setting enabled](https://github.com/tensorflow/tensorflow/commit/2ff9f49e5a00af7d1afe05a3f7915e3d067b6126).

A test release is currently running [here](https://github.com/larq/compute-engine/actions/runs/5812935736).


Tombana (approved):
I think it's working. The job failed due to a timeout, but I've restarted them now.

> relatively short limit

I'm not sure if 32K is short, I think TF just has long commandlines :sweat_smile:

CNugteren (approved):

PR #805, opened 2024-05-24
Update Tensorflow to 2.16.1

CNugteren:
## What do these changes do?
Very much a draft, only updated TF itself, CI, and Bazel so far. Still to be done: update the actual C++ code as needed.


larq_compute_engine/mlir/transforms/fuse_padding.td
```diff
@@ -43,8 +43,7 @@ def : Pat<(TFL_Conv2DOp:$conv_output
           [(HasOneUse $pad_output),
            (NoBatchAndChannelPadding $paddings),
            (SamePaddingHeight $paddings, $input, $conv_output, $stride_h),
-           (SamePaddingWidth $paddings, $input, $conv_output, $stride_w)],
-          (addBenefit 100)>;
```
Tombana, line 47: There is now an extra argument _before_ the benefit-argument. So we could either insert a `[], ` here before `addBenefit`, or just remove it. @lgeiger do you know if we depend on these benefit values? A lot of them seem arbitrary to me, but some seem to be chosen specifically.

build_pip_pkg.sh
```diff
@@ -29,7 +29,8 @@ function is_windows() {
 }
 
 if is_windows; then
-  PIP_FILE_PREFIX="bazel-bin/build_pip_pkg.exe.runfiles/larq_compute_engine/"
+  # On windows, the workspace name is lce to avoid the path length limit of MSVC
```
CNugteren, line 32: :open_mouth: :angry:

CNugteren (commented):
Thanks for going through all this effort @Tombana :pray:  Looks ready to merge to me (I can't approve since I'm the original author of the PR) :+1: .

larq_compute_engine/mlir/transforms/fuse_padding.td
Tombana: I looked through the patterns again, and I added back one of the `addBenefit`s and left all the others out.

The one I added back was in order to give priority to `GreaterEqual(x, 0) -> Quantize` pattern over the more generic `GreaterEqual(x, threshold) -> Quantize(x - threshold)` pattern. This might avoid a 'subtract 0' operation.
For all other patterns, I don't think the benefit value really mattered: you could first convert a `ste.sign(x)` to `Quantize(Dequantize(x))` or first convert a Conv to a Bconv; that order does not matter. (Besides, it only applies to patterns _within_ one pass)

Tombana (approved):

.github/workflows/release.yml
```diff
@@ -154,33 +154,35 @@ jobs:
...
 
           if [[ -n $GOOGLE_APPLICATION_CREDENTIALS ]]; then
-            echo -e 'build --remote_http_cache=https://storage.googleapis.com/plumerai-bazel-cache/lce-release-macos-python${{ matrix.python-version }}' >> .bazelrc.user
+            echo -e 'build --remote_cache=https://storage.googleapis.com/plumerai-bazel-cache/lce-release-macos-python${{ matrix.python-version }}' >> .bazelrc.user
             echo -e 'build --google_default_credentials' >> .bazelrc.user
           fi
 
-          bazelisk build :build_pip_pkg --copt=-fvisibility=hidden --copt=-mavx --linkopt=-dead_strip --distinct_host_configuration=false
-          bazel-bin/build_pip_pkg artifacts --plat-name macosx_10_14_x86_64
+          bazelisk build :build_pip_pkg --config=release_macos_x86 --config=release_cpu_macos --copt=-fvisibility=hidden --linkopt=-dead_strip
```
lgeiger, line 165: Should we still keep `--distinct_host_configuration=false`? Though happy to keep it as is if it works

lgeiger (approved):
Awesome! Thank you so much for upgrading 🙏

.github/workflows/release.yml
Tombana: That flag was removed in this bazel version :)
I think it would still be nice to have the same effect to avoid a few double builds, but in order to achieve that with the new bazel I think you have to do add some complex toolchain files.

PR #807, opened 2024-06-20
Add Python 3.12 release

Tombana:
## What do these changes do?

Add Python 3.12 to the release builds and use Python 3.12 in regular CI.
I made Python 3.10 the minimal version so that we still support the latest 3 Python versions, similar to the previous release.

The testrelease is running here: https://github.com/larq/compute-engine/actions/runs/9596990350


lgeiger (approved):
🚀

Tombana:
It seems that the locked requirements.txt file is quite different for Python 3.12. Perhaps we should do the same as what TF does and have separate `requirements_{python_version}.txt` files.

CNugteren (approved):

